Issue #38, opened 2022-03-01
Support partial base classes


latonz:
Mapperly supports base classes with mapping methods which have a body.
E.g.
```c#
class DateMapper
{
    protected DateOnly DateTimeToDateOnly(DateTime dt) => DateOnly.FromDateTime(dt);
}

[Mapper]
partial class DtoMapper : DateMapper
{
    public partial CarDto ToDto(Car car);
}
```
In this sample, when the `DtoMapper` needs to map a `DateTime` to a `DateOnly` Mapperly will use the `DateMapper.DateTimeToDateOnly` method.

However if the `DateMapper` is itself a `Mapper` with partial methods, these won't be re-used. This should be improved.
E.g.
```c#
[Mapper]
partial class BaseMapper
{
    protected partial BaseDto ToBaseDto(BaseObject source);
}

[Mapper]
partial class DtoMapper : BaseMapper
{
    public partial CarDto ToDto(Car car);
}
```
If the `DtoMapper` needs to map a property of source type `BaseObject` to target type `BaseDto` it should re-use the already built mapping `BaseMapper.ToBaseDto`, but this currently does not work.

PR #42, opened 2022-03-01
feat: support partial base classes

latonz:
Closes https://github.com/riok/mapperly/issues/38

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilder/UserMethodMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilder/UserMethodMappingBuilder.cs
@@ -14,7 +14,7 @@ public static class UserMethodMappingBuilder
         // extract user implemented and user defined mappings from mapper
         foreach (var methodSymbol in ExtractMethods(mapperSymbol))
         {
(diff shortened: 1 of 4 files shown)


src/Riok.Mapperly/Descriptors/MappingBuilder/UserMethodMappingBuilder.cs
```diff
@@ -14,15 +14,19 @@ public static class UserMethodMappingBuilder
...
+            var mapping = BuilderUserDefinedMapping(ctx, methodSymbol) ?? BuildUserImplementedMapping(methodSymbol, false);
             if (mapping != null)
                 yield return mapping;
         }
 
         // extract user implemented mappings from base methods
         foreach (var method in ExtractBaseMethods(ctx.Compilation.ObjectType, mapperSymbol))
         {
-            var mapping = BuildUserImplementedMapping(method);
+            // partial method declarations are allowed vor base classes,
```
CommonGuy, line 25: Change this to
> Partial method declarations are allowed for base classes,
> but still treated as user implemented methods,
> since the user should provide an implementation elsewhere.
> This is the case if a partial mapper class is extended.

latonz: Improved.

CommonGuy (approved):

Issue #58, opened 2022-03-29
Add support for static mappers


zitmen:
Hi, would you consider support of static mappers?
For example this:
```csharp
[Mapper]
public static partial class DtoMapper
{
    public static partial CarDto CarToCarDto(Car car);
}
```
It currently generates an empty class.
Since I did not see this in the examples and don't have experience with MapStruct, I don't dare to say it is a bug :-)
Thanks!

latonz:
Thank you for your feature request, sounds like a good idea to implement this. Feel free if you want to contribute a PR.

TheR00st3r:
Static Classes with ExtensionMethods would be rly great

   ``` csharp
    [Mapper]
    public partial class DtoMapper
    {
        public static partial User ToUser(this ApplicationUser applicationUser);
    }
   ```

latonz:
Implemented by #76

PR #60, opened 2022-04-02
Added support for static mappers

zitmen:
Resolves issue #58

I have tested it with this trivial sample:
```csharp
[Mapper]
public static partial class TestStaticMapper
{
    public static partial IdObjectDto IdObjectToIdObjectDto(IdObject value);
    public static partial IdObjectDto ToDto(this IdObject value);

    public static partial void IdObjectToIdObjectDto2(IdObjectDto input, IdObject output);
    public static partial void ToDto2(this IdObject input, IdObject output);
}


[Mapper]
public partial class TestMapper
{
    public partial IdObjectDto ModelToDto(IdObject value);
    public static partial IdObjectDto ToDto(IdObject value);

    public partial void ModelToDto2(IdObject value, IdObjectDto dto);
    public static partial void ToDto2(IdObject value, IdObjectDto dto);
}
```

I originally thought I would add tests, but to be honest, I have no idea how to make them work as I have trouble even compiling them. I guess I need to work on my "build tools" skills :-)

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilder/UserMethodMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilder/UserMethodMappingBuilder.cs
@@ -93,13 +93,11 @@ public static class UserMethodMappingBuilder
         => m.Parameters.Length == 2
             && m.ReturnsVoid
             && !m.IsAsync
-            && !m.IsStatic
             && !m.IsGenericMethod;
 
     private static bool IsNewInstanceMappingMethod(IMethodSymbol m)
         => m.Parameters.Length == 1
             && !m.ReturnsVoid
             && !m.IsAsync
-            && !m.IsStatic
             && !m.IsGenericMethod;
 }
--- a/src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
@@ -22,6 +22,10 @@ public abstract class MethodMapping : TypeMapping
 
     protected bool Partial { get; set; }
 
+    protected bool IsStatic { get; set; }
+
+    protected bool IsExtensionMethod { get; set; }
+
     protected string MethodName
     {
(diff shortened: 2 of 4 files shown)


latonz:
Thank you for your PR!

You should be able to run the tests via your IDE's integrated features or by running `dotnet test` on a terminal in the project's root directory.

A few hints to add tests:
* Add an integration test
  * You can pretty much copy the `Riok.Mapperly.IntegrationTests.Mapper.TestMapper` for example as `Riok.Mapperly.IntegrationTests.Mapper.StaticTestMapper` and set all methods and the class as static
  * Add methods in the `Riok.Mapperly.IntegrationTests.MapperTest` for the newly created `StaticTestMapper`
* Add unit tests
  * Appropriate test classes include `Riok.Mapperly.Tests.Mapping.UserMethodTest`. Consider als adding a new `Riok.Mapperly.Tests.Mapping.StaticMapperTest` class.

On a first review of the PR a few things are missing here:
* Mapperly generates private mapping methods which get called by the generated mapping method implementations. For example, mapping a class Car with a property of type Maker would generate a method `private MakerDto ToMakerDto(Maker source)`. Currently these methods are always instance methods. These should be static when a static mapper is being used.
* User implemented mapping methods are reused by Mapperly if possible. With the implementation of static mapping methods, new restrictions need to be implemented. A user implemented instance (non static) mapping method (implemented in the mapper partial class, any base class or an interface) can only be called from an instance method, but not from static methods.

Please let me know if you need further assistance :slightly_smiling_face:

zitmen:
Thanks for the hints, but I just cannot make it work. I have the most up to date VS'22.
What is the trick to make it to generate the mapper and then to compile?
image

zbecknell:
I hope this gets added. I'm happy to see this library gaining traction, as I've had my eye on source generated mappers for a while.

latonz:
Thanks for your efforts in this PR. Since there was no recent activity, I implemented this feature in https://github.com/riok/mapperly/pull/76. Therefore I'll close this PR.

PR #100, opened 2022-07-14
fix: InvalidOperationException (Collection was modified) on BuildInitOnlyPropertyMappings method

vicfergar:


latonz:
Thank you for your PR!

Could you add a test case which works on your PR but does not work on the main branch? Thank you 😊

vicfergar:
> Thank you for your PR!
> 
> Could you add a test case which works on your PR but does not work on the main branch? Thank you 😊

Actually, I could not build the IntegrationTests project without this change. Not sure why it was not breaking `test` workflows on previous runs 🤔

latonz:
Ah got it, are your building with .NET Framework? The code works fine with .NET 6 but breaks on .NET Framework. I created https://github.com/riok/mapperly/issues/101 to run the tests on .NET Framework as well.

latonz (approved):

PR #131, opened 2022-08-30
fix: publish preview releases on nuget and adjust readme badges

CommonGuy:


.github/workflows/release-preview.yml
```diff
@@ -14,9 +14,9 @@ jobs:
     uses: ./.github/workflows/release.yml
     with:
       environment: next
-      nuget_source: 'https://nuget.pkg.github.com/riok/index.json'
+      nuget_source: 'https://api.nuget.org/v3/index.json'
```
latonz, line 17: I'd instead set the default of `release.yml`>inputs>nuget_source to this value and remove it from stable and preview.

latonz (changes requested):

.github/workflows/release-preview.yml
CommonGuy: Improved

latonz (approved):

PR #141, opened 2022-09-15
fix: ignore accessible source properties with inaccessible getters and accessible target properties with inaccessible getters

### LINKED ISSUE #140: "The property or indexer cannot be used in this context because the set accessor is inaccessible"
**Describe the bug**
Properties with private/protected setters throw exception

**To Reproduce**
```csharp
public class Foo
{
    public string Bar { get; private set; } //see, the property is public but the setter is private
}
```

**Expected behavior**
I suggest skipping properties with non-public setters

latonz:
Closes #140

### DIFF
--- a/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
@@ -303,4 +303,56 @@ private partial D MapToD(C source);
 
         return TestHelper.VerifyGenerator(source);
     }
+
+    [Fact]
+    public Task WithPrivateTargetSetterShouldIgnoreAndDiagnostic()
+    {
+        var source = TestSourceBuilder.Mapping(
+            "A",
+            "B",
+            "class A { public string StringValue { get; set; } public int IntValue { get; private set; } }",
+            "class B { public string StringValue { get; private set; } public int IntValue { private get; set; } }");
+
+        return TestHelper.VerifyGenerator(source);
+    }
+
+    [Fact]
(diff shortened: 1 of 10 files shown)


test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
```diff
@@ -303,4 +303,56 @@ public Task WithNotFoundIgnoredSourcePropertyShouldDiagnostic()
...
+    [Fact]
+    public Task WithPrivateSourcePathGetterShouldIgnoreAndDiagnostic()
+    {
+        var source = TestSourceBuilder.Mapping(
+            "A",
+            "B",
+            "class A { public C StringValue { private get; set; } public int IntValue { get; private set; } }",
+            "class B { public C StringValue { get; set; } public int IntValue { private get; set; } }",
+            "class C { public string StringValue { get; set; } }",
+            "class D { public string StringValue { get; set; } }");
```
CommonGuy, line 354: Class `D` is unused here. In all the new test cases, I would use a name other than `StringValue` for the properties of type `C` and `D`. For example `Nested` or something else

latonz: Improved.

CommonGuy: Now class `C` is unused in all tests

latonz: 🤦 should be correct now.

CommonGuy (approved):

PR #181, opened 2022-11-18
feat: Required Properties (C# 11)

DerMave:
willing to add some more tests if neccessary

### DIFF
--- a/src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
+++ b/src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
@@ -182,4 +182,12 @@ internal static class DiagnosticDescriptors
         DiagnosticCategories.Mapper,
         DiagnosticSeverity.Error,
         true);
+
+    public static readonly DiagnosticDescriptor RequiredPropertyNotMapped = new DiagnosticDescriptor(
+        "RMG023",
+        "Source property was not found for required target property",
+         "Required property {0} on mapping target type was not found on the mapping source type {1}",
+        DiagnosticCategories.Mapper,
+        DiagnosticSeverity.Error,
+        true);
 }
--- a/src/Riok.Mapperly/AnalyzerReleases.Unshipped.md
+++ b/src/Riok.Mapperly/AnalyzerReleases.Unshipped.md
@@ -1,2 +1,7 @@
 ; Unshipped analyzer release
 ; https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+RMG023 | Mapper | Error | DiagnosticDescriptors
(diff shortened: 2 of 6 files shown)


latonz:
Thanks for your PR!

You may need to adjust your commit message to match the conventional commit message specification.

latonz:
If a required property cannot be mapped, Mapperly should emit an error diagnostic.

latonz (changes requested):
Thanks for your PR!

You may need to adjust your commit message to match the conventional commit message specification.

If a required property cannot be mapped, Mapperly should emit an error diagnostic.

DerMave:
added tests for the diagnostic and fixed the commit messages

test/Riok.Mapperly.Tests/_snapshots/ObjectPropertyInitPropertyTest.RequiredPropertySourceNotFoundShouldDiagnostic.00.verified.txt
```diff
@@ -0,0 +1,28 @@
+﻿{
+  Diagnostics: [
+    {
+      Id: RMG012,
+      Title: Mapping source property not found,
+      Severity: Info,
```
latonz, line 6: I think this should be of level error, as this can never work for required properties. Therefore a new diagnostic descriptor should be introduced (eg. `RequiredPropertyNotMapped`) with a severity of error which is reported for properties with the required flag instead of the `RMG012`.

latonz (changes requested):

CommonGuy:
@DerMave The tests are currently failing because `Microsoft.CodeAnalysis, Version=4.4.0.0` cannot be found. The Microsoft.CodeAnalysis package is provided by the SDK. To get it to work, I think you need to change the SDK version in the `global.json` file to 7.0.100.

latonz:
> @DerMave The tests are currently failing because `Microsoft.CodeAnalysis, Version=4.4.0.0` cannot be found. The Microsoft.CodeAnalysis package is provided by the SDK. To get it to work, I think you need to change the SDK version in the `global.json` file to 7.0.100.

This should work after rebasing onto `main`.

latonz (changes requested):
After rebasing onto `main` the pipeline should work...

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -182,4 +182,11 @@ internal static class DiagnosticDescriptors
         DiagnosticCategories.Mapper,
         DiagnosticSeverity.Error,
         true);
+    public static readonly DiagnosticDescriptor RequiredPropertyNotMapped = new DiagnosticDescriptor(
```
latonz, line 185: To match the code style, could you add an empty line between above this line?

latonz (changes requested):
The tests are failing. The reason for this is that since you generated your snapshot the verify tests source generator dependency was updated to the latest major release which included a breaking change in the naming of the snapshots. That's why you need to regenerate/rename the snapshots of the tests you added.

DerMave:
Hey @latonz ,
thanks for your feedback!
I made the changes, rebased and cleaned it up as much as I can. I hope everything is fine now.

PR #205, opened 2022-12-14
fix: multi target roslyn 4.4 and 4.0

### LINKED ISSUE #204: Mappers using 2.6.0-next-1 do not build against the .NET6 SDK
**Describe the bug**
Project using 2.6.0-next-1 does not build against the .NET6 SDK

**To Reproduce**
Add mapperly 2.6.0-next-1 to a project targering netstandard2.0, on a system where the .NET6 SDK is installed.

Error during build time:
The analyzer assembly ... references version '4.4.0.0' of the compiler, which is newer than the currently running version '4.3.0.0'. 

**Expected behavior**
Projects are able to be built using the .NET6 SDK. Currently it seems like it is only able to build with the .NET 7 SDK installed, which indeed includes the 4.4.0.0 version of the compiler, which is not the LTS branch.

This seems to be the result of #179 which bumped the required version from 4.0.1 to 4.4.0

 - Mapperly Version: 2.6.0-next-1
 - .NET Version: NET 6.0.4xx
 - Target Framework: netstandard2.0
 - OS: ubuntu 20.04 on Azure DevOps

latonz:
Closes #204


src/Riok.Mapperly/Riok.Mapperly.csproj
```diff
@@ -17,8 +17,20 @@
     <PackageIcon>logo.png</PackageIcon>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <ROSLYN_VERSION Condition="'$(ROSLYN_VERSION)' == ''">4.4</ROSLYN_VERSION>
+    <DefineConstants Condition="'$(ROSLYN_VERSION)' == '4.4'">$(DefineConstants);ROSLYN4_0_OR_GREATER</DefineConstants>
```
CommonGuy, line 22: Shouldn't this constant be called `ROSLYN4_4_OR_GREATER`?

latonz: Whooops, fixed.

Issue #226, opened 2023-01-12
Mapping to readonly array


heggi:
**Is your feature request related to a problem? Please describe.**
Protobuf compiler (protoc) generate class with fields which has only getter for arrays.
Example
```proto
syntax = "proto3";
option csharp_namespace = "Test";

message TestMessage {
    repeated int32 repeatedField = 1;
}
```
generate to 
```csharp
public sealed partial class TestMessage : pb::IMessage<TestMessage>
{
// skip many code
    private readonly pbc::RepeatedField<int> repeatedField_ = new pbc::RepeatedField<int>();
    public pbc::RepeatedField<int> RepeatedField {
      get { return repeatedField_; }
    }
// skip many code
}
```
Usualy this used as
```csharp
var test = new TestMessage();
test.RepeatedField.Add(111);
test.RepeatedField.AddRange(new[] { 111, 222, 333 });
```

Now Mapperly ignore properties like that.

**Describe the solution you'd like**
Add to generated code construction like
```csharp
if (target.RepeatedField != null)
{
    target.RepeatedField.AddRange(source.RepeatedField);
}
```
or 
```csharp
target.RepeatedField?.AddRange(source.RepeatedField);
```

**Additional context**
Tested with latest version 2.6.0-next.3
Google.Protobuf version 3.21.12
Grpc.Tools version 2.50.0

latonz:
Thank you for this report. This is definitely something which needs improvement.

jamescarterbellMSFT:
I was trying to find a stop gap solution for this same exact problem (also doing gRPC mappings) and found the following generate code that *seems* right, but sadly the parent generator doesn't seem to find the void mapping methods:
`private partial void MapToFooBars(List<FooBarModel>? source, RepeatedField<FooBarMessage> target)`

Generates:
```
if (source == null)
    return;
target.Capacity = source.Capacity;
target.this[] = MapToFooBar(source.this[]);
```

Blackclaws:
Its similar for dictionaries in generated grpc code. Those are MapFields which also only have a getter. Basically you have to copy the contents from one to the other.

jamescarterbellMSFT:
[...]

PR #234, opened 2023-01-17
Adds support for readonly IList mapping 

jamescarterbellMSFT:
This PR addresses #226 by introducing a new pathway for generating and keeping track of reference based mappings inside of mapperly.  This should allow (for example) and IList to IList map to still be a simple assignment, but special cases and IList to readonly IList mapping to be via adding to an existing IList reference.

### DIFF
--- a/src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
@@ -103,17 +103,18 @@ public class DescriptorBuilder
         var ctx = new SimpleMappingBuilderContext(this);
         ObjectFactories = ObjectFactoryBuilder.ExtractObjectFactories(ctx, _mapperSymbol);
     }
-    public ITypeMapping? FindMapping(ITypeSymbol sourceType, ITypeSymbol targetType)
-        => _mappings.FindMapping(sourceType, targetType);
+    public ITypeMapping? FindMapping(ITypeSymbol sourceType, ITypeSymbol targetType, RefKind targetRefkind = RefKind.None)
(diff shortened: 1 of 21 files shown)


jamescarterbellMSFT:
Note: this special case pathway is currently using RefKind, but since mapperly only deals with properties (which can't be passed using 'ref') this will need to be changed.

I'm also not sure how the generator tests work, would love some pointers on how a test for this feature would look.

latonz:
Thank you for your PR. A brief description of how the tests work can be found [here](https://mapperly.riok.app/docs/contributing/tests). You already extended the integration tests. The integration tests reference the analyzers and run them on the mappers and snapshot the mapped objects as well as the generated code by using [VerifyTests/Verify](https://github.com/VerifyTests/Verify). If you never used VerifyTests you may find it useful to read the documentation of the project. To be more accurate, in the GitHub Actions, the integration tests reference the nupkg instead of the analyzer projects.
The unit tests are located in the `test/Riok.Mapperly.Tests` project. The mapping tests are located in the `Mapping` directory.  Probably the one you need to extend is the `EnumerableTest` and the `DictionaryTest`. In these classes you can also find examples of existing tests. You can use the `TestSourceBuilder` class to build a mapping class source and the `TestHelper` run the source generator and verify it's result.

As soon as the PR is ready for a first review, you can request myself as a reviewer.

latonz:
Thanks for your PR contribution! Due to the great need for this feature, I implemented this in a separate PR, so that Mapperly users are unblocked from using it with readonly lists. Sorry for the inconvenience.
If your PR implements additional features to #226, feel free to open a new Issue/PR.

PR #239, opened 2023-01-19
feat: constructor mapping support for MapProperty attribute

martinothamar:
Hi, thanks for this great library!

I had an issue where I tried to map a class to a record with primary constructor, where one of the property names were not matching, so I had to use the `MapProperty` attribute, this failed and I think I tracked down why. I included a couple of tests to illustrate the issue. The `ClassesCanMapToPrimaryConstructorRecordsWhenUsingMapProperty` tests failed, but it worked without `MapProperty`, so hopefully I've added the correct mapping code

### DIFF
--- a/src/Riok.Mapperly/AnalyzerReleases.Shipped.md
+++ b/src/Riok.Mapperly/AnalyzerReleases.Shipped.md
@@ -65,3 +65,5 @@ RMG025  | Mapper   | Error    | To use reference handling it needs to be enabled
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
 RMG026  | Mapper   | Info     | Cannot map from indexed property
+RMG027  | Mapper   | Warning  | A constructor parameter can have one configuration at max
+RMG028  | Mapper   | Error    | Constructor parameter cannot handle target paths
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilder/NewInstanceObjectPropertyMappingBodyBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilder/NewInstanceObjectPropertyMappingBodyBuilder.cs
@@ -198,12 +198,7 @@ public static class NewInstanceObjectPropertyMappingBodyBuilder
         var skippedOptionalParam = false;
         foreach (var parameter in ctor.Parameters)
         {
-            if (!PropertyPath.TryFind(
-                ctx.Mapping.SourceType,
-                MemberPathCandidateBuilder.BuildMemberPathCandidates(parameter.Name),
-                ctx.IgnoredSourcePropertyNames,
-                StringComparer.OrdinalIgnoreCase,
-                out var sourcePath))
+            if (!TryFindConstructorParameterSourcePath(ctx, parameter, out var sourcePath))
             {
                 if (!parameter.IsOptional)
                     return false;
@@ -229,6 +224,57 @@ public static class NewInstanceObjectPropertyMappingBodyBuilder
             var ctorMapping = new ConstructorParameterMapping(parameter, propertyMapping, skippedOptionalParam);
             constructorParameterMappings.Add(ctorMapping);
             mappedTargetPropertyNames.Add(parameter.Name);
+        }
+
+        return true;
+    }
+
+    private static bool TryFindConstructorParameterSourcePath(
+        NewInstanceMappingBuilderContext ctx,
+        IParameterSymbol parameter,
+        [NotNullWhen(true)] out PropertyPath? sourcePath)
+    {
+        sourcePath = null;
+
+        if (!ctx.PropertyConfigsByRootTargetName.TryGetValue(parameter.Name, out var propertyConfigs))
+        {
+            return PropertyPath.TryFind(
+                ctx.Mapping.SourceType,
+                MemberPathCandidateBuilder.BuildMemberPathCandidates(parameter.Name),
+                ctx.IgnoredSourcePropertyNames,
+                StringComparer.OrdinalIgnoreCase,
+                out sourcePath
+            );
+        }
+
+        if (propertyConfigs.Count > 1)
+        {
+            ctx.BuilderContext.ReportDiagnostic(
+                DiagnosticDescriptors.MultipleConfigurationsForConstructorParameter,
+                parameter.Type,
+                parameter.Name);
+        }
+
+        var propertyConfig = propertyConfigs.First();
+        if (propertyConfig.Target.Count > 1)
+        {
+            ctx.BuilderContext.ReportDiagnostic(
+                DiagnosticDescriptors.ConstructorParameterDoesNotSupportPaths,
+                parameter.Type,
+                string.Join(".", propertyConfig.Target));
+            return false;
+        }
+
+        if (!PropertyPath.TryFind(
+            ctx.Mapping.SourceType,
+            propertyConfig.Source,
+            out sourcePath))
+        {
+            ctx.BuilderContext.ReportDiagnostic(
+                DiagnosticDescriptors.MappingSourcePropertyNotFound,
+                propertyConfig.Source,
+                ctx.Mapping.SourceType);
+            return false;
         }
 
         return true;
--- a/src/Riok.Mapperly/Descriptors/MappingBuilder/NewInstanceMappingBuilderContext.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilder/NewInstanceMappingBuilderContext.cs
@@ -17,6 +17,8 @@ public class NewInstanceMappingBuilderContext : ObjectPropertyMappingBuilderCont
 
     public void AddConstructorParameterMapping(ConstructorParameterMapping mapping)
     {
+        PropertyConfigsByRootTargetName.Remove(mapping.Parameter.Name);
+
         SetSourcePropertyMapped(mapping.DelegateMapping.SourcePath);
         Mapping.AddConstructorParameterMapping(mapping);
     }
--- a/src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
+++ b/src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
@@ -213,4 +213,20 @@ internal static class DiagnosticDescriptors
         DiagnosticCategories.Mapper,
         DiagnosticSeverity.Info,
         true);
+
+    public static readonly DiagnosticDescriptor MultipleConfigurationsForConstructorParameter = new DiagnosticDescriptor(
+        "RMG027",
+        "A constructor parameter can have one configuration at max",
+        "The constructor parameter at {0}.{1} can have one configuration at max",
+        DiagnosticCategories.Mapper,
+        DiagnosticSeverity.Warning,
+        true);
+
+    public static readonly DiagnosticDescriptor ConstructorParameterDoesNotSupportPaths = new DiagnosticDescriptor(
+        "RMG028",
+        "Constructor parameter cannot handle target paths",
+        "Cannot map to constructor parameter target path {0}.{1}",
+        DiagnosticCategories.Mapper,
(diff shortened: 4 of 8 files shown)


src/Riok.Mapperly/Descriptors/MappingBodyBuilder/NewInstanceObjectPropertyMappingBodyBuilder.cs
```diff
@@ -197,40 +197,72 @@ private static void BuildInitOnlyPropertyMappings(NewInstanceMappingBuilderConte
...
-
-            var propertyMapping = new NullPropertyMapping(delegateMapping, sourcePath, ctx.BuilderContext.GetNullFallbackValue(paramType));
-            var ctorMapping = new ConstructorParameterMapping(parameter, propertyMapping, skippedOptionalParam);
-            constructorParameterMappings.Add(ctorMapping);
-            mappedTargetPropertyNames.Add(parameter.Name);
         }
 
         return true;
+
+        static bool TryAddMapping(
```
latonz, line 239: IMO instead of the ref parameterised method you proposed the following approach could simplify the code and improve the overall readability:
Add a method with the following signature:
```csharp
private static bool TryFindConstructorParameterSourcePath(
        NewInstanceMappingBuilderContext ctx,
        IParameterSymbol parameter,
        [NotNullWhen(true)] out PropertyPath? sourcePath)
```
which tries to get the `MapPropertyAttribute` from `PropertyConfigsByRootTargetName`, validates it, creates diagnostics as needed (see `BuildInitPropertyMapping `) and resolves the source property path or if `PropertyConfigsByRootTargetName` doesn't return an entry, try to resolve the source property directly (as currently implemented in the else branch).
In the `foreach (var parameter in ctor.Parameters)` loop then call this new method. If true is returned execute the code you currently have in `TryAddMapping`. If false is returned, check if the parameter is optional, if it is, set `skippedOptionalParam` to true and continue, if it isn't optional return false.

src/Riok.Mapperly/Descriptors/MappingBodyBuilder/NewInstanceObjectPropertyMappingBodyBuilder.cs
```diff
@@ -195,42 +195,88 @@ private static void BuildInitOnlyPropertyMappings(NewInstanceMappingBuilderConte
     {
         constructorParameterMappings = new HashSet<ConstructorParameterMapping>();
         mappedTargetPropertyNames = new HashSet<string>();
-        var skippedOptionalParam = false;
+        bool skippedOptionalParam = false;
```
latonz, line 198: I would stick with `var` as it was used before and it is used all across the codebase.

src/Riok.Mapperly/Descriptors/MappingBodyBuilder/NewInstanceObjectPropertyMappingBodyBuilder.cs
```diff
@@ -195,42 +195,88 @@ private static void BuildInitOnlyPropertyMappings(NewInstanceMappingBuilderConte
...
+            var propertyMapping = new NullPropertyMapping(delegateMapping, sourcePath, ctx.BuilderContext.GetNullFallbackValue(paramType));
+            var ctorMapping = new ConstructorParameterMapping(parameter, propertyMapping, skippedOptionalParam);
+            constructorParameterMappings.Add(ctorMapping);
             mappedTargetPropertyNames.Add(parameter.Name);
         }
 
         return true;
     }
+
+    private static bool TryFindConstructorParameterSourcePath(NewInstanceMappingBuilderContext ctx, IParameterSymbol parameter, [NotNullWhen(true)] out PropertyPath? sourcePath)
```
latonz, line 232: IMO each parameter on a new line could improve readability of this method signature a lot.

src/Riok.Mapperly/Descriptors/MappingBodyBuilder/NewInstanceObjectPropertyMappingBodyBuilder.cs
```diff
@@ -195,42 +195,88 @@ private static void BuildInitOnlyPropertyMappings(NewInstanceMappingBuilderConte
...
+        }
+
+        if (!PropertyPath.TryFind(
+            ctx.Mapping.SourceType,
+            MemberPathCandidateBuilder.BuildMemberPathCandidates(parameter.Name),
+            ctx.IgnoredSourcePropertyNames,
+            StringComparer.OrdinalIgnoreCase,
+            out sourcePath)
+        )
+            return false;
```
latonz, line 278: If the if clause spans over multiple lines, this codebase usually uses braces for the body of the if. However this could be transformed in a direct return:
```csharp
return PropertyPath.TryFind(...
```

src/Riok.Mapperly/Descriptors/MappingBodyBuilder/NewInstanceObjectPropertyMappingBodyBuilder.cs
```diff
@@ -195,42 +195,88 @@ private static void BuildInitOnlyPropertyMappings(NewInstanceMappingBuilderConte
...
         }
 
         return true;
     }
+
+    private static bool TryFindConstructorParameterSourcePath(NewInstanceMappingBuilderContext ctx, IParameterSymbol parameter, [NotNullWhen(true)] out PropertyPath? sourcePath)
+    {
+        sourcePath = null;
+
+        if (ctx.PropertyConfigsByRootTargetName.TryGetValue(parameter.Name, out var propertyConfigs))
```
latonz, line 236: Inverting this condition and keeping the early return could lead to easier to read code since the else part is much less code (early return on the case no config is found).

src/Riok.Mapperly/Descriptors/MappingBuilder/NewInstanceMappingBuilderContext.cs
```diff
@@ -17,6 +17,9 @@ public void AddInitPropertyMapping(PropertyAssignmentMapping mapping)
 
     public void AddConstructorParameterMapping(ConstructorParameterMapping mapping)
     {
+        if (PropertyConfigsByRootTargetName.ContainsKey(mapping.Parameter.Name))
```
latonz, line 20: I think the contains key call is not needed at all, since the remove is a no-op if the key isn't found.

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -213,4 +213,20 @@ internal static class DiagnosticDescriptors
         DiagnosticCategories.Mapper,
         DiagnosticSeverity.Info,
         true);
+
+    public static readonly DiagnosticDescriptor MultipleConfigurationsForConstructorArgument = new DiagnosticDescriptor(
```
latonz, line 217: I think the term parameter would fit better here than argument (same in the message and for RMG028).

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyConstructorResolverTest.cs
```diff
@@ -349,5 +349,37 @@ public void RecordToRecordFlattenedNonDirectAssignmentNullHandling()
             .Should()
             .HaveSingleMethodBody(@"var target = new B(source.Nested == null ? throw new System.ArgumentNullException(nameof(source.Nested.Value)) : (double)source.Nested.Value);
     return target;");
+    }
+
+    [Fact]
+    public void CanResolveToRecordConstructorWithMapPropertyAttribute()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            "[MapProperty(nameof(ThingClass.Id), nameof(ThingRecord.ThingId))] partial ThingRecord ToRecord(ThingClass thing);",
```
latonz, line 358: This codebase almost always uses mapping test classes A and B, therefore I'd rename `ThingClass` to `A` and `ThingRecord` to `B` to be consistent.

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
```diff
@@ -65,3 +65,5 @@ RMG025  | Mapper   | Error    | To use reference handling it needs to be enabled
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
 RMG026  | Mapper   | Info     | Cannot map from indexed property
+RMG027  | Mapper   | Warning  | A constructor argument can have one configuration at max
+RMG028  | Mapper   | Warning  | Constructor argument cannot handle target paths
```
latonz, line 69: This one should be of severity error, shouldn't it?

latonz (changes requested):
No worries at all, thats what reviews are for! I'm glad you are contributing 😊 The implementation looks pretty good already!

Yap, new diagnostics are good. Also adding them to the released file is how we do it as we don't want to create release commits.

I think the current design is ok. To apply the MapProperty attribute to map to ctor parameters, one couldn't use the nameof operator but can still use the string overload. IMO thats also what should be done in the `CanResolveToClassConstructorWithMapPropertyAttribute` test case: `[MapProperty(nameof(ThingClass.Id), "id2)]`.

martinothamar:
Addressed the new comments now, and included the documentation in the document you suggested.

Did I get the diagnostics severity right? In that one should be error, the other warning?

I did find one "issue" when writing the documentation. When a target type has both a constructor and a settable public property, then it is both mapping the property to the constructor parameter as well as the public setter of the target property:

```csharp
public class Car
{
    public string ModelName { get; set; }
}

public class CarDto
{
    // highlight-start
    public CarDto(string model)
    // highlight-end
    {
        ModelName = model;
    }

    public string ModelName { get; set; }
}

[Mapper]
public partial class CarMapper
{
    // highlight-start
    [MapProperty(nameof(Car.ModelName), "model")]
    // highlight-end
    public partial CarDto ToDto(Car car);
}
```
Generates the following

```csharp
#nullable enable
namespace Riok.Mapperly.Testss
{
    public partial class CarMapper
    {
        public partial Riok.Mapperly.Testss.CarDto ToDto(Riok.Mapperly.Testss.Car car)
        {
            var target = new Riok.Mapperly.Testss.CarDto(car.ModelName);
            target.ModelName = car.ModelName;
            return target;
        }
    }
}
```

I'm not sure I define this a bug, in this specific case it's a noop, but there are cases where this can produce some unexpected results for sure

martinothamar:
In terms of code formatting, I've enjoyed my experience with [belav/csharpier](https://github.com/belav/csharpier). I use it across all projects using the MsBuild package in a [Directory.Build.props file in my Mediator](https://github.com/martinothamar/Mediator/blob/main/Directory.Build.props) project and use it for various projects at work as well. Maybe that can be considered here as well

latonz:
I think the behaviour you described is as expected (at least with the current implementation). Mapperly uses the target properties as the list of properties to map, and therefore doesn't know this property is already set via the ctor. I think this case should be rare though (no matching/manual matched ctor parameter but a matching setable property). And there is an easy fix for such cases by using the `MapperIgnoreTargetAttribute`.

Regarding the formatting: this codebase uses the dotnet built-in `dotnet format` command to check the format of the source code based on the editorconfig file. Until now this was sufficient... I didn't know csharpier and will check it out for sure. However, I think there need to be some crucial  advantages over dotnet format to pull in another dependency.

Could you squash your commits together in one feature commit with a commit message according to the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) specification (the commit message will land in the release notes) and rebase your branch onto the latest `riok:main`?

martinothamar:
Updated commit msg and PR title and rebased.

Without going too off-topic here, the drawback with `dotnet format` that I found was that it didn't really deal with line length much ([seems like that is still the case](https://github.com/dotnet/format/issues/246)). Csharpier seems to handle everything, and is more prettier-like/opinionated.

latonz (approved):

Issue #229, opened 2023-01-13
Nullability in array mappings not quite correct


CommonGuy:
Mapping an array in a "nullable disabled" context generates incorrect nullable annotations. The generated mapping method returns `string[]`, but it should be `string?[]`

PR #252, opened 2023-01-25
fix: support mapping to target readonly collections

### LINKED ISSUE #226: Mapping to readonly array

latonz:
Introduces a new concept of ExistingTargetMappings, which are mappings which map to an existing target. Currently only Collection-, Dictionary- and ObjectPropertyMappings support existing target mappings.
For an object property an existing target mapping is tried to be built if it the source and target property is readable, but the target property is not setable.

Closes #226.


test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
```diff
@@ -78,6 +78,71 @@ public void WithExistingInstance()
...
+            "partial void Map(IReadOnlyDictionary<string, A> source, IDictionary<string, B> target)",
+            "record A(string Value);",
+            "record B(string Value);");
+
+        TestHelper.GenerateMapper(source)
+            .Should()
+            .HaveMapMethodBody("""
+                    foreach (var item in source)
+                    {
+                        target.Add(item.Key, MapToB(item.Value));
```
CommonGuy, line 141: I think it would be safer to use the indexer instead of `.Add`, otherwise it could lead to collisions if the target already contains entries. This is an edge case IMO, could be cleaned up separately.

test/Riok.Mapperly.Tests/_snapshots/EnumerableTest.ShouldUpgradeNullabilityInDisabledNullableContextInSelectClause#Mapper.g.verified.cs
```diff
@@ -26,12 +26,12 @@ public partial class Mapper
 
     private D[] MapToDArray(C[] source)
     {
-        var target = new D[source.Length];
+        var target = new D?[source.Length];
```
CommonGuy, line 29: While the target is now correctly created, the nullability of the return type is incorrect.

The nullability of the `C[]` type is also incorrect (should be `C?[]`). This could also be solved in a separate PR, since an issue for this is still open

latonz: Should be resolved in https://github.com/riok/mapperly/issues/229

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
latonz: Improved.

CommonGuy (approved):

PR #282, opened 2023-03-16
chore: Use raw strings in test assertions

latonz:


CommonGuy (approved):

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
```diff
@@ -252,7 +225,7 @@ public Task WithManualMappedNotFoundTargetPropertyShouldDiagnostic()
     public Task WithManualMappedNotFoundSourcePropertyShouldDiagnostic()
     {
         var source = TestSourceBuilder.MapperWithBodyAndTypes(
-            "[MapProperty(nameof(A.StringValue9), nameof(B.StringValue2)] partial B Map(A source);",
+            "[MapProperty(\"A.StringValue9\", nameof(B.StringValue2)] partial B Map(A source);",
```
ni507, line 228: What is the need of this change? To test if it also works with double quotes?

This is a bit confusing for me. Would "B.StringValue2" also throw a property not found error? If this is not the case, IMO the message in the snapshot `Specified property A.StringValue9 on source type A was not found` should no contain the `A.` prefix for the property, since the source type is explicitly named in second part of the phrase.

Issue #214, opened 2023-01-08
Support Select/Projection of an IQueryable


pwelter34:
**Is your feature request related to a problem? Please describe.**
AutoMapper supports ProjectTo<TDestination>(this IQueryable) for creating a select expression for EF or other querable providers.  Honestly, after this project, its the last reason to use AutoMapper.  If this project could create the needed extension method, that would be super useful.

**Describe the solution you'd like**

Original
```c#
[Mapper]
public static partial class CarMapper
{
    public static partial IQueryable<CarDto> ProjectToCarDto(this IQueryable<Car> source);
}
```

Generated

```c#
public static partial class CarMapper
{
    public static IQueryable<CarDto> ProjectToCarDto(this IQueryable<Car> source)
    {
        return source
            .Select<Car, CarDto>(original =>
                new CarDto
                {
                    Make = original.Make, 
                    Model = original.Model
                }
            );
    }
}
```

**Additional context**

https://docs.automapper.org/en/stable/Queryable-Extensions.html

latonz:
This is something we would like to implement. However, this needs some effort.

Rel. https://github.com/riok/mapperly/discussions/99

latonz:
Idea on how this could work:

A partial method for the projection is defined in the mapper:
```csharp
public static partial IQueryable<CarDto> ProjectToDto(this IQueryable<Car> q);
```

To manually configure properties, additional mapping methods are defined in the mapper as needed:
```csharp
[MapProperty("Model.Name", "ModelName")]
public static partial CarDto ToDto(Car car);
```

Mapperly generates the implementation of the projection method based on the configurations on the other defined methods, if available.
The methods with the configurations are also implemented by Mapperly.
This generates more code than neccessairy,
but it is an approach to make queryable projection mappings configurable.

Complete example:
```csharp
// classes
public class Car
{
    public int Id { set; get; }

    public Manufacturer Manufacturer { set; get; } = new();
}

public class Manufacturer
{
    public string Name { set; get; } = string.Empty;
}

public class CarDto
{
    public int Id { set; get; }
    
    public string Producer { set; get; } = string.Empty;
}

// mapper definition
[Mapper]
public static partial class CarMapper
{
    public static partial IQueryable<CarDto> ProjectToDto(this IQueryable<Car> q);

    [MapProperty("Manufacturer.Name", "Producer")]
    public static partial CarDto ToDto(Car car);
}

// generated mapper implementation
public partial class CarMapper
{
[...]

PR #287, opened 2023-03-17
feat: Support IQueryable projection mappings

latonz:
Closes https://github.com/riok/mapperly/issues/214


docs/docs/02-configuration/11-queryable-projections.mdx
```diff
@@ -0,0 +1,73 @@
...
+    // highlight-start
+    .ProjectToDto()
+    // highlight-end
+    .ToListAsync();
+```
+
+</TabItem>
+</Tabs>
+
+This is useful in combination with EntityFramework and other ORM solutions
```
CommonGuy, line 37: Would change this to
> This is useful in combination with Entity Framework and other ORM solutions
which expose `IQueryable<T>`. Only fields present in the target class will be retrieved from the database.

The last sentence (code inlining) is not interesting to Mapperly consumers IMO, which is why I would remove it

docs/docs/02-configuration/11-queryable-projections.mdx
```diff
@@ -0,0 +1,73 @@
...
+```csharp
+[Mapper]
+public static partial class CarMapper
+{
+    // highlight-start
+    public static partial IQueryable<CarDto> ProjectToDto(this IQueryable<Car> q);
+    // highlight-end
+
+    // highlight-start
+    [MapProperty(nameof(Car.Manufacturer), nameof(CarDto.Producer)]
```
CommonGuy, line 69: Wouldn't it make more sense to support this on the projection method?

docs/docs/02-configuration/12-conversions.md
```diff
@@ -47,3 +48,18 @@ public partial class CarMapper
   ...
 }
 ```
+
+## Enable only specific automatic conversions
+
+To enable only specific conversion types, set `EnabledConversions` the conversion type to enable:
+
+```csharp
+// this disables conversions using the ToString() method:
```
CommonGuy, line 57: Would change this to

> // This disables conversions using the ToString() method, which is enabled by default:

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -229,4 +229,36 @@ internal static class DiagnosticDescriptors
...
+        "RMG030",
+        "Reference loop detected while mapping to an init only property",
+        "Reference loop detected while mapping from {0}.{1} to the init only property {2}.{3}, consider ignoring this property",
+        DiagnosticCategories.Mapper,
+        DiagnosticSeverity.Error,
+        true);
+
+    public static readonly DiagnosticDescriptor ReferenceLoopInCtorMapping = new DiagnosticDescriptor(
+        "RMG031",
+        "Reference loop detected while mapping to a ctor parameter",
```
CommonGuy, line 251: Would always spell out "constructor" instead of using the abbreviation

src/Riok.Mapperly/Descriptors/Mappings/QueryableProjectionMapping.cs
```diff
@@ -0,0 +1,45 @@
...
+        ITypeSymbol sourceType,
+        ITypeSymbol targetType,
+        ITypeMapping delegateMapping) : base(sourceType, targetType)
+    {
+        _delegateMapping = delegateMapping;
+    }
+
+    public override IEnumerable<StatementSyntax> BuildBody(TypeMappingBuildContext ctx)
+    {
+        // #nullable disable
```
CommonGuy, line 31: Why is this needed here? Add a comment. May also need to be added to the documentation

docs/docs/02-configuration/11-queryable-projections.mdx
latonz: Improved.

docs/docs/02-configuration/11-queryable-projections.mdx
latonz: https://www.itrockt.ch/mitglieder/riok-gmbh/

src/Riok.Mapperly/Descriptors/Mappings/QueryableProjectionMapping.cs
CommonGuy: I don't think just disabling nullability solves this problem correctly. What happens if values are actually null?

Don't we need to handle all nullable values correctly anyway? I don't quite get why this is needed

Issue #251, opened 2023-01-23
Dictionary and Enumerable Mappings: Call `EnsureCapacity` if source count is known and target has a `EnsureCapacity(int)` method


latonz:
If Mapperly generates a `foreach` / `Add` mapping the capacity of the target is increased as needed.
If the source/target count is known (type does have a `Count`/ `Length` property or `Enumerable.TryGetNonEnumeratedCount()` returns true) and the target type does have a Method `EnsureCapacity(int)`, `EnsureCapacity(source.Count + target.Count)` can be called to ensure the target collection has a big enough capacity for all source objects plus the existing target objects and does not need to resize as mapped objects are added.

Blackclaws:
This is already done via constructor arguments for dictionaries and lists I think. Calling EnsureCapacity there would be redundant

latonz:
You are correct, if the target is created by Mapperly via `new` and the target type is a List or Dictionary. But if the target instance already exists, it is created via an object factory or is any type implementig `IDictionary<,>` or `IList<>` but is not Dictionary/List (eg. a custom List implementation), the capacity is not ensured.

Issue #250, opened 2023-01-23
Support System.Collections.Immutable types


latonz:
Support collection types in `System.Collections.Immutable` as mapping target types.
Eg. Mapperly should be able to implement a mapping from an `IEnumerable<long>` to an `ImmutableList<int>`.

jamesfoster:
While this feature is not implemented is there a generic way to work around this?

Something like:

```csharp
private static ImmutableList<TDest> ToImmutableList<TDest, TSource>(
    IEnumberable<TSource> source,
    Func<TSource, TDest> map
)
{
   return source.Select(map).ToImmutableList();
}
```

We can obviously add a workaround for each type but it doesn't seem to be possible to implement it generically.

latonz:
Currently there is no support for generic user implemented mapping methods... I can't think of a feasible workaround 😢

jamesfoster:
Thanks for the response. Without knowing too much about how source generators work, would it not be possible to support the suggested code? This could be useful for any generic type with a single type argument. if you know how to map from `TSource` to `TDest` you can supply a value for `map`. You can then use this method to map from any `A<TSource>` to a `B<TDest>`.

TimothyMakkison:
> Currently there is no support for generic user implemented mapping methods... I can't think of a feasible workaround 😢

I'll give it a go. Looks like `EnumerableMappingBuilder.ResolveCollectMethodName` can be used to add `ToImmutableList` etc. Not sure which immutable collections should be added.

latonz:
@jamesfoster this could probably be a new feature of Mapperly. However, I think it adds a lot of complexity and I'm not sure if it justifies the added complexity.

jamesfoster:
> @jamesfoster this could probably be a new feature of Mapperly. However, I think it adds a lot of complexity and I'm not sure if it justifies the added complexity.

No worries. My use case was immutable collections anyway, and it looks like you're working on it. I can work around it for the time being by adding the non-generic mapper for each type as needed.

Thanks.

latonz:
https://github.com/riok/mapperly/pull/305 implements support for the following immutable collections:
* `ImmutableList<T>`
* `ImmutableArray<T>`
* `ImmutableHashSet<T>`
* `ImmutableQueue<T>`
* `ImmutableSortedSet<T>`

latonz:
[...]

Issue #286, opened 2023-03-17
Enumerable / Dictionary mappings do not support explicit interface implementations


latonz:
Related to https://github.com/riok/mapperly/issues/283

When mapping from one collection or dictionary to another Mapperly checks whether appropriate interfaces are implemented. However, it does not check whether these interfaces are explicitly implemented which leads to compile errors.

**To Reproduce**

Mapper Source
```csharp
[Mapper]
public partial class MyMapper
{
    public partial A MapToA(IReadOnlyDictionary<string, string> dict);
}

public class A : IDictionary<string, string>
{
       // removed other members for simplicity
       
       private readonly Dictionary<string, string> _dictionaryImplementation = new();

        string IDictionary<string, string>.this[string key]
        {
            get => _dictionaryImplementation[key];
            set => _dictionaryImplementation[key] = value;
        }
}
```

Generated mapping method
```csharp
private partial A ToA(System.Collections.Generic.IReadOnlyDictionary<string, string> x)
{
    var target = new A();
    foreach (var item in x)
    {
        target[item.Key] = item.Value;
    }

    return target;
}
```

`target[item.Key] = item.Value;` produces a compiler error as the indexed setter is only available via the explicit interface implementation. The generated code in this case should be sth. like:
```csharp
private partial A ToA(System.Collections.Generic.IReadOnlyDictionary<string, string> x)
{
    var target = new A();
    var targetDict = (IDictionary<string, string>)target;
    foreach (var item in x)
    {
        targetDict[item.Key] = item.Value;
    }

    return target;
}
```

PR #305, opened 2023-03-29
feat: add mapping to immutable sequences

TimothyMakkison:
Partially resolving #250 and #251.

- Added `ImmutableList` and `ImmutableArray`
- Added `EnsureCapacity` generator to `ForEachAddEnumerableMapping`, it works on `ICollection` and `ICollection<T>` and casts the sequence in case the interface was implemented explicitly (aka arrays).
- Added `AddRange` for existing `List`
- Add `IsArray` guard to `TryBuildExistingTargetMapping`, Array explicitly implements a dummy `IList.Add` method that thows `NotImplementedException`. as a result mapperly will attempt to generate a foreach loop adding items to the array.

As `ImmutableList` and `ImmutableArray` cannot be mutated without a setter method, should `EnumerableMappingBuilder.TryBuildExistingTargetMapping` report an error instead of returning null?

I'm not 100% happy with how I added `EnsureCapacity` it will have to be revisited when being used for dictionaries. A future change could use `Enumerable.TryGetNonEnumeratedCount` instead.


src/Riok.Mapperly/Descriptors/Mappings/LinqAddRangeMapping.cs
```diff
@@ -0,0 +1,50 @@
...
+        // Select / Map if needed
+        if (_selectMethod != null)
+        {
+            var sourceMapExpression = _elementMapping.Build(ctx.WithSource(lambdaParamName));
+            var convertLambda = SimpleLambdaExpression(Parameter(Identifier(lambdaParamName))).WithExpressionBody(sourceMapExpression);
+            mappedSource = StaticInvocation(_selectMethod, ctx.Source, convertLambda);
+        }
+        else
+        {
+            mappedSource = _elementMapping.Build(ctx);
```
latonz, line 43: This seems to be uncovered by tests, could you add a test which covers this?

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -58,14 +65,30 @@ public static class EnumerableMappingBuilder
...
 
+        // TODO Error: RMG008?
+        if (ctx.Target.IsArrayType()
+            || ctx.Target.ImplementsGeneric(ctx.Types.ImmutableArrayT, out _)
+            || ctx.Target.ImplementsGeneric(ctx.Types.ImmutableListT, out _))
+        {
+            return null;
+        }
+
+        if (ctx.Target.ImplementsGeneric(ctx.Types.List, out _))
```
latonz, line 76: Is a `Select` + `AddRange` mapping faster than a `foreach` + `Add` mapping (assumed https://github.com/riok/mapperly/issues/251 is implemented)? It makes sense to me if the `elementMapping` is synthetic, but does it also make sense for non-synthetic mappings?

latonz (changes requested):
Thank you for another great contribution!
Could you split the support for the immutable array/list (https://github.com/riok/mapperly/issues/250) and EnsureCapacity (https://github.com/riok/mapperly/issues/251) into two separate PR's? Separating features into single PR's makes reviews a lot easier and the changes are easier to track.

TimothyMakkison:
Thanks for the code review, I'll get round to creating a separate EnsureCapacity pull request.

- Added some support for some Immutable type.
- Created `HasImplicitInterfaceMethod` to prevent mapperly calling ICollection<T>.Add on Immutable types when they have an explicit implementation - related to #286. I can remove this and manually check if you don't like this.
- Added test to ensure that mapperly does not attempt to call ICollection<T>.Add on an array type.
- Added integraton tests, it looks like I only generated values for `Roslyn_4_5`. How would I run it for other versions to be like #300? I'm not sure why `ImmutableArrayValue` `ImmutableQueueValue` and `ImmutableStackValue` are the only immutable types that appear inside `RecursiveObject` [here](https://github.com/riok/mapperly/pull/305/files#diff-3d68b0b1676cb425b7402a21499bce6b1f410c02965625099f06f2822f105179).
```C#
// RecursiveObject...
ImmutableArrayValue: null,
ImmutableQueueValue: [],
ImmutableStackValue: [],
```
- Update names of generic types in `WellKnowTypes`

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -128,30 +148,77 @@ private static (bool CanMapWithLinq, string? CollectMethod) ResolveCollectMethod
...
+        var selectMethod = elementMapping.IsSynthetic
+            ? null
+            : ResolveLinqMethod(ctx, SelectMethodName);
+
+        return new LinqEnumerableMapping(ctx.Source, ctx.Target, elementMapping, selectMethod, collectMethod);
+    }
+
+    private static IMethodSymbol? ResolveImmutableCollectMethod(MappingBuilderContext ctx)
+    {
+        if (ctx.Target.ImplementsGeneric(ctx.Types.ImmutableArrayT, out _))
```
latonz, line 192: First I was wondering why this works as `ImplementsGeneric` expects the `genericInterfaceSymbol` parameter to be an interface. It does actually work, since `ImplementsGeneric` does have a `SymbolEqualityComparer.Default.Equals(t.OriginalDefinition, genericInterfaceSymbol)` check at the beginning of the implementation.
I would consider this an implementation detail of the  `ImplementsGeneric` method and instead of calling this method replace these calls with `SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableArrayT)`.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -78,6 +78,32 @@ internal static IEnumerable<IPropertySymbol> GetAllAccessibleProperties(this ITy
         return genericIntf != null;
     }
 
+
+    // TODO: refactor and rename.
```
latonz, line 82: Resolve the todo,
nit: two empty lines instead of one before the method...

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -78,6 +78,32 @@ internal static IEnumerable<IPropertySymbol> GetAllAccessibleProperties(this ITy
...
+        if (!symbol.ImplementsGeneric(inter, out var typedInter))
+            return false;
+
+        var interfaceMethodSymbol = typedInter.GetMembers(methodName).OfType<IMethodSymbol>().Single();
+
+        var methodInterImplementaton = symbol.FindImplementationForInterfaceMember(interfaceMethodSymbol) as IMethodSymbol;
+
+        // if null then the method is unimplemented
+        if (methodInterImplementaton is null)
+            return true;
```
latonz, line 101: Should this method really return true in case the method is not implemented at all? `HasImplicitInterfaceMethod` resolving to true for unimplemented methods seems unexpected to me. But this can never happen anyway (due to the `ImplementsGeneric` check on L92, can it? So probably just an exception could be thrown?

test/Riok.Mapperly.IntegrationTests/BaseMapperTest.cs
```diff
@@ -79,6 +81,12 @@ protected TestObject NewTestObj()
                 SourceTargetSameObjectType = new TestObject(8) { IntValue = 99, RequiredValue = 98, },
                 StackValue = new Stack<string>(new[] { "1", "2", "3" }),
                 QueueValue = new Queue<string>(new[] { "1", "2", "3" }),
+                ImmutableArrayValue = ImmutableArray.Create(new[] { "1", "2", "3" }),
```
latonz, line 84: These array creations are redundant as the target parameter is a `params`...

latonz (changes requested):
Thank you for the updates 😊 
* I'm fine with the `HasImplicitInterfaceMethod` check 👍 
* Integration tests: Once the github action integration tests for net6.0 and net framework have run, the received snapshot files can be downloaded from the [artifacts](https://github.com/riok/mapperly/actions/runs/4579113048?pr=305). If you wan't to run the tests locally you can either try to use [act](https://github.com/nektos/act) (I personally never tried that) or run the commands of the test github action manually (see [test.yml github action](https://github.com/riok/mapperly/blob/main/.github/workflows/test.yml#L46)).
* I don't know why the other values don't show up in the integration test snapshots... Probably it is some serialization logic of VerifyTests. If you assign a value, it will show up in the snapshot.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
TimothyMakkison: Spot on, ordinarily you'd expect this method to be implemented as part of the underlying interface. But in cases like:
```C#
public class RepeatedField : ICollection<int> { }
```
The class `RepeatedField` "implements" the interface but hasn't implemented any methods. This normally fails to compile but except for the unit tests where the above example is common.

My solution is hacky but its better than updating all the tests. I'll add a comment explaining why I've done this.

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -58,13 +72,16 @@ public static class EnumerableMappingBuilder
...
 
-        if (ctx.Target.ImplementsGeneric(ctx.Types.Queue, out _))
+        if (ctx.Target.ImplementsGeneric(ctx.Types.QueueT, out _))
             return new ForEachAddEnumerableExistingTargetMapping(ctx.Source, ctx.Target, elementMapping, nameof(Queue<object>.Enqueue));
 
-        if (ctx.Target.ImplementsGeneric(ctx.Types.ICollection, out _))
+        // create a foreach loop with add calls if source is not an array
+        // and  ICollection.Add(T): void is implemented and not explicit
+        // ensures add is not called and immutable types
+        if (!ctx.Target.IsArrayType() && ctx.Target.HasImplicitInterfaceMethod(ctx.Types.ICollectionT, nameof(ICollection<object>.Add)))
```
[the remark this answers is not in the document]
TimothyMakkison: `HasImplicitInterfaceMethod` uses `FindImplementationForInterfaceMember` to get the matching interface method. For some reason this returns null if the symbol is an array type. 

This could be due to how interfaces work with arrays generics, I dont think `GetMembers` works on array types.
```C#
// if the symbol is an array.
symbol.GetMembers().Count() == 0;
```

TimothyMakkison:
Thanks for the suggestions, I'll try to add the integration files for 4_4 and under.

I've noticed a slight oversight, the [tests for .Net framework 4.8](https://github.com/riok/mapperly/actions/runs/4579113048/jobs/8093723358) always fails because [Immutable types](https://github.com/riok/mapperly/actions/runs/4579113048/jobs/8093723358) are .Net core only. I'm a little suprised I could use them in a .Net Standard 2.0 project.

Not sure how to resolve this, my first instinct would be to multi target .Net Standard 2.0 and .Net Core and use preprocessor directives. Thoughts?

test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
```diff
@@ -633,6 +759,78 @@ public Task MapToExistingQueueShouldWork()
         return TestHelper.VerifyGenerator(source);
     }
 
+    [Fact]
+    public Task EnumerableToReadonlyImmutableList()
```
latonz, line 763: Not sure if this and the following tests make sense.... They are trying to map from an `int` property to an `Immutable*`, shouldn't the source property type be of type `IEnumerable<int>`?

test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
```diff
@@ -692,4 +902,16 @@ public void EnumerableMappingDisabledShouldDiagnostic()
...
+    [Fact]
+    public Task EnumerableToReadOnlyArrayPropertyShouldntForeachAdd()
+    {
+        var source = TestSourceBuilder.Mapping(
+            "A",
+            "B",
+            "class A { public ICollection<int> Value { get; } }",
+            "class B { public int[] Value { get; } }");
+
+        return TestHelper.VerifyGenerator(source);
```
latonz, line 915: I would adjust the name of the test method to `EnumerableToReadOnlyArrayPropertyShouldDiagnostic` and add a comment (eg. `cannot create a mapping which maps to a read only array`).
Probably also change the type of assertion from snapshot to the fluent assertion style instead of the snapshot (`TestHelper.GenerateMapper(source, TestHelperOptions.AllowInfoDiagnostics) .Should()...`), since the snapshot content isn't the thing being tested here.

latonz (changes requested):
Thanks for the updates 😊

To resolve the issue of the immutable types not being available in net framework, adding the [System.Collections.Immutable nuget package](https://www.nuget.org/packages/System.Collections.Immutable/) to `Riok.Mapperly.IntegrationTests.csproj` as a conditional dependency (in the same item group as `Portable.System.DateTimeOnly`) should help.

TimothyMakkison:
Thanks for the review. I've added and updated the tests, changed `HasImplicitMethod` and added the conditional dependency.

I'm not familiar with XUnit verify does it automatically delete snapshot data when the corresponding method is removed or renamed?

latonz (changes requested):
VerifyTests does not automatically delete unused snapshot files, they need to be cleaned up manually.

TimothyMakkison:
Got the integration tests to work 😅 removed the unused snapshot data, refactored the test and did some code cleanup.
Should be ready to merge 🤞

latonz (changes requested):
The integration tests, especially the ones for the older Roslyn versions are indeed a hassle. If you have an idea how the situation could be improved, I would be glad to hear it 😊 

Looks good now, could you rebase onto main?

latonz (approved):

Issue #68, opened 2022-04-25
Strict mode


WeUseBc:
Hi and thank you for making this :)

One of the reasons why I'm not keen on using runtime mappers, is that by refactoring and various changes in the codebase, I might rename properties in my domain model, which silently breaks mapping to e.g. dto's.

I would really like a "strict mode" where all the properties of the target is required which should fail if something is missing. Then I would have to explicitly have to add either [Ignore] or [MapProperty] to handle those cases. A bit like we get an compiler error if mapping between types isn't possible.

Is this something that you would consider adding?

latonz:
Currently there is a diagnostic `RMG012: Mapping source property not found` which is emitted when a source property is not found with the severity of info. However, you can set its severity to  `error` by adding `dotnet_diagnostic.RMG012.severity = error` to your editorconfig. Currently the message of `RMG012` is not optimal (it doesn't note what the target class is. Also there is no diagnostic if a property does only exist on source but not on target.

A diagnostic for unmapped source properties sounds like a good idea to implement. Feel free if you want to contribute a PR.

WeUseBc:
Thanks for your answer :)

I haven't worked with source generators at all, so could you give some pointers on how you would implement such a thing?

latonz:
To add the target type to the message of `RMG012`: Look up usages of `MappingSourcePropertyNotFound` in `Riok.Mapperly.Diagnostics.DiagnosticDescriptors`, extend its message format and add the target type message argument on the callers side.

To add a new diagnostic when there is an unmapped source property add a new entry in `Riok.Mapperly.Diagnostics.DiagnosticDescriptors` and `AnalyzerReleases.Shipped.md`.
[...]

PR #310, opened 2023-04-01
fix: add target type to the RMG012 diagnostic

toddlucas:
RMG023 was also updated to include the target type, due to the way the message is invoked alongside RMG012. The message format of RMG012 was updated to follow the format of RMG020.

This relates to #68

### DIFF
--- a/src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
+++ b/src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
@@ -96,8 +96,8 @@ internal static class DiagnosticDescriptors
 
     public static readonly DiagnosticDescriptor SourcePropertyNotFound = new DiagnosticDescriptor(
         "RMG012",
-        "Mapping source property not found",
-        "Property {0} on source type {1} was not found",
+        "Source property was not found for target property",
+        "The property {0} on the mapping target type {1} was not found on the mapping source type {2}",
         DiagnosticCategories.Mapper,
         DiagnosticSeverity.Info,
         true);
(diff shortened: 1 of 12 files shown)


src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -96,8 +96,8 @@ internal static class DiagnosticDescriptors
 
     public static readonly DiagnosticDescriptor SourcePropertyNotFound = new DiagnosticDescriptor(
         "RMG012",
-        "Mapping source property not found",
-        "Property {0} on source type {1} was not found",
+        "Source property was not found for target property",
+        "The property {0} on the mapping target type {1} was not found on the mapping source type {2}",
```
latonz, line 100: My english is not too good, but I'm not sure about this sentence. It says the property was not found on the target type and not on the source type 🤔 But it is actually only missing on the source type, isn't it?
What about:
`While mapping to {0} on {1} no property named {0} could be found on the source type {1}`

and 
`While mapping to the required property {0} on {1} no property named {0} could be found on the source type {1}`
for `RMG023`?

latonz (changes requested):
Thank you for your contribution 😊 Displaying the target type in the message of these diagnostics definitely makes sense!

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
toddlucas: It seems like your English is great. I'm a native speaker and it's confusing. There are two existing messages which I was attempting to parallel.

* RMG020: "The property {0} on the mapping source type {1} is not mapped to any property on the mapping target type {2}"
* RMG023: "Required property {0} on mapping target type was not found on the mapping source type {1}"

These are the updated versions (23 now includes the target type):
* RMG020: "The property {0} on the mapping source type {1} is not mapped to any property on the mapping target type {2}"
* RMG023: "Required property {0} on mapping target type {1} was not found on the mapping source type {2}"

This change updates 12 so that it's a combination of 20 and 23:
* RMG012: "The property {0} on the mapping target type {1} was not found on the mapping source type {2}"

I'd be happy to switch it around if you prefer. In that case, we might want to update 20 to keep it consistent.

latonz: Ah nvm, now I understand it.

latonz (approved):

latonz (approved):

PR #312, opened 2023-04-04
Add ensure capacity for enumerable types.

TimothyMakkison:
Complete #251 for enumerable types. I'm not sure how Explicit `Count` values should be treated, right now it gets the runtime type using the `is` operator. I think it should either try not to get the value or just cast the value so the explicit property can be accessed.

- Added `EnsureCapacityBuilder`, `CanEnsureCapacity` checks if ensure capacity can be used, storing the relevant information in `EnsureCapacityInfo` to be used by `TryBuildEnsureCapacityStatement` when the syntax is built.
  - If both source and target have compile time sizes `TryBuildEnsureCapacityStatement` generates a simple call to `EnsureCapacity`.
  - If source does not have a length/count property, `is` statements are created to check the runtime type, invoking `EnsureCapacity` with the resulting type.
  - Not sure where `EnsureCapacityBuilder` should be placed.
- Added tests, waiting for integration tests from github actions.

### DIFF
--- a/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacity.cs
+++ b/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacity.cs
@@ -1,0 +1,20 @@
+﻿using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Riok.Mapperly.Descriptors.Mappings;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity;
+
+public abstract class EnsureCapacity
+{
+    private const string EnsureCapacityName = "EnsureCapacity";
+
+    public abstract StatementSyntax Build(TypeMappingBuildContext ctx, ExpressionSyntax target);
+
+    protected static ExpressionStatementSyntax EnsureCapacityStatement(ExpressionSyntax target, ExpressionSyntax sourceCount, ExpressionSyntax targetCount)
+    {
+        var sumMethod = BinaryExpression(SyntaxKind.AddExpression, sourceCount, targetCount);
+        return ExpressionStatement(Invocation(MemberAccess(target, EnsureCapacityName), sumMethod));
+    }
+}
--- a/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
@@ -1,0 +1,71 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Helpers;
+
+namespace Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity;
+
+/// <summary>
+/// Generates an <see cref="EnsureCapacity?"/> of types <see cref="EnsureCapacityNonEnumerated"/> or <see cref="EnsureCapacityMember"/> depending on type information.
+/// </summary>
+public static class EnsureCapacityBuilder
+{
+    private const string EnsureCapacityName = "EnsureCapacity";
+    private const string CountPropertyName = nameof(ICollection<object>.Count);
+    private const string LengthPropertyName = nameof(Array.Length);
+    private const string TryGetNonEnumeratedCountMethodName = "TryGetNonEnumeratedCount";
+
+    public static EnsureCapacity? TryBuildEnsureCapacity(ITypeSymbol sourceType, ITypeSymbol targetType, WellKnownTypes types)
+    {
+        var capacityMethod = targetType.GetAllMethods(EnsureCapacityName)
+                        .OfType<IMethodSymbol>()
+                        .FirstOrDefault(x => x.Parameters.Length == 1
+                        && x.Parameters[0].Type.SpecialType == SpecialType.System_Int32
+                        && !x.IsStatic);
+
+        // if EnsureCapacity is not available then return null
+        if (capacityMethod == null)
+            return null;
+
+        // if target does not have a count then return null
+        if (!TryGetNonEnumeratedCount(targetType, types, out var targetSizeProperty))
+            return null;
+
+        // if target and source count are known then create a simple EnsureCapacity statement
+        if (TryGetNonEnumeratedCount(sourceType, types, out var sourceSizeProperty))
+            return new EnsureCapacityMember(targetSizeProperty, sourceSizeProperty);
+
+        sourceType.ImplementsGeneric(types.IEnumerableT, out var iEnumerable);
+
+        var nonEnumeratedCountMethod = types.Enumerable.GetMembers(TryGetNonEnumeratedCountMethodName)
+                                                        .OfType<IMethodSymbol>()
+                                                        .FirstOrDefault(x => x.ReturnType.SpecialType == SpecialType.System_Boolean && x.IsStatic && x.Parameters.Length == 2 && x.IsGenericMethod);
+
+        // if source does not have a count use GetNonEnumeratedCount, calling EnusureCapacity if count is available
+        var typedNonEnumeratedCount = nonEnumeratedCountMethod.Construct(iEnumerable!.TypeArguments.ToArray());
+        return new EnsureCapacityNonEnumerated(targetSizeProperty, typedNonEnumeratedCount);
+    }
+
+    private static bool TryGetNonEnumeratedCount(ITypeSymbol value, WellKnownTypes types, [NotNullWhen(true)] out string? expression)
+    {
+        if (value.IsArrayType())
+        {
+            expression = LengthPropertyName;
+            return true;
+        }
+
+        if (value.ImplementsGeneric(types.ICollectionT, CountPropertyName, out _, out var hasCollectionCount) && !hasCollectionCount)
+        {
+            expression = CountPropertyName;
+            return true;
+        }
+
+        if (value.ImplementsGeneric(types.IReadOnlyCollectionT, CountPropertyName, out _, out var hasReadOnlyCount) && !hasReadOnlyCount)
+        {
+            expression = CountPropertyName;
+            return true;
+        }
+
+        expression = null;
+        return false;
+    }
+}
--- a/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
@@ -1,6 +1,7 @@
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity;
 using Riok.Mapperly.Descriptors.Mappings;
 using Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
 using Riok.Mapperly.Diagnostics;
@@ -78,16 +79,16 @@ public static class EnumerableMappingBuilder
             return null;
 
         if (ctx.Target.ImplementsGeneric(ctx.Types.StackT, out _))
-            return new ForEachAddEnumerableExistingTargetMapping(ctx.Source, ctx.Target, elementMapping, nameof(Stack<object>.Push));
+            return CreateForEach(nameof(Stack<object>.Push));
 
         if (ctx.Target.ImplementsGeneric(ctx.Types.QueueT, out _))
-            return new ForEachAddEnumerableExistingTargetMapping(ctx.Source, ctx.Target, elementMapping, nameof(Queue<object>.Enqueue));
+            return CreateForEach(nameof(Queue<object>.Enqueue));
 
         // create a foreach loop with add calls if source is not an array
         // and  ICollection.Add(T): void is implemented and not explicit
         // ensures add is not called and immutable types
         if (!ctx.Target.IsArrayType() && ctx.Target.HasImplicitInterfaceMethod(ctx.Types.ICollectionT, AddMethodName))
-            return new ForEachAddEnumerableExistingTargetMapping(ctx.Source, ctx.Target, elementMapping, AddMethodName);
+            return CreateForEach(AddMethodName);
 
         // if a mapping could be created for an immutable collection
         // we diagnostic when it is an existing target mapping
@@ -97,6 +98,12 @@ public static class EnumerableMappingBuilder
         }
 
         return null;
+
+        ForEachAddEnumerableExistingTargetMapping CreateForEach(string methodName)
+        {
+            var ensureCapacityStatement = EnsureCapacityBuilder.TryBuildEnsureCapacity(ctx.Source, ctx.Target, ctx.Types);
+            return new ForEachAddEnumerableExistingTargetMapping(ctx.Source, ctx.Target, elementMapping, methodName, ensureCapacityStatement);
+        }
     }
 
     private static ITypeMapping? BuildElementMapping(MappingBuilderContext ctx)
@@ -164,7 +171,10 @@ public static class EnumerableMappingBuilder
         // and  ICollection.Add(T): void is implemented and not explicit
         // ensures add is not called and immutable types
         if (!ctx.Target.IsArrayType() && ctx.Target.HasImplicitInterfaceMethod(ctx.Types.ICollectionT, AddMethodName))
-            return new ForEachAddEnumerableMapping(ctx.Source, ctx.Target, elementMapping, objectFactory, AddMethodName);
+        {
+            var ensureCapacityStatement = EnsureCapacityBuilder.TryBuildEnsureCapacity(ctx.Source, ctx.Target, ctx.Types);
+            return new ForEachAddEnumerableMapping(ctx.Source, ctx.Target, elementMapping, objectFactory, AddMethodName, ensureCapacityStatement);
+        }
 
         return null;
     }
--- a/test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
@@ -605,7 +605,7 @@ public class EnumerableTest
     [Fact]
     public void EnumerableToExplicitAddCustomCollection()
     {
-        // should not create a mapping using a looped add inside a foreach loop when th add method is explicit
+        // should not create a mapping using a looping add method inside a foreach loop when the add method is explicit
         var source = TestSourceBuilder.Mapping(
             "IEnumerable<int>",
             "B",
@@ -632,6 +632,51 @@ public class EnumerableTest
     }
 
     [Fact]
+    public void EnumerableToExistingCustomCollection()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            "partial void Map(IEnumerable<int> source, A target);",
+            "class A : List<int> { }");
+        TestHelper.GenerateMapper(source)
+            .Should()
+            .HaveSingleMethodBody(
+                """
+                if (global::System.Linq.Enumerable.TryGetNonEnumeratedCount(source, out var sourceCount))
+                {
+                    target.EnsureCapacity(sourceCount + target.Count);
+                }
+
+                foreach (var item in source)
+                {
+                    target.Add(item);
+                }
+                """);
+    }
+
+    [Fact]
+    public void ReadOnlyCollectionToList()
+    {
+        var source = TestSourceBuilder.Mapping(
+            "A",
+            "B",
+            "class A { public IReadOnlyCollection<int> Value { get; } }",
+            "class B { public List<int> Value { get; } }");
+        TestHelper.GenerateMapper(source)
+            .Should()
+            .HaveMapMethodBody(
+                """
+                var target = new global::B();
+                target.Value.EnsureCapacity(source.Value.Count + target.Value.Count);
+                foreach (var item in source.Value)
+                {
+                    target.Value.Add(item);
+                }
+
+                return target;
+                """);
+    }
+
+    [Fact]
     public void EnumerableToStack()
     {
         var source = TestSourceBuilder.Mapping(
@@ -644,6 +689,11 @@ public class EnumerableTest
             .HaveMapMethodBody(
                 """
                 var target = new global::B();
+                if (global::System.Linq.Enumerable.TryGetNonEnumeratedCount(source.Value, out var sourceCount))
+                {
+                    target.Value.EnsureCapacity(sourceCount + target.Value.Count);
+                }
+
                 foreach (var item in source.Value)
                 {
                     target.Value.Push((long)item);
@@ -666,6 +716,11 @@ public class EnumerableTest
             .HaveMapMethodBody(
                 """
                 var target = new global::B();
+                if (global::System.Linq.Enumerable.TryGetNonEnumeratedCount(source.Value, out var sourceCount))
+                {
+                    target.Value.EnsureCapacity(sourceCount + target.Value.Count);
+                }
+
                 foreach (var item in source.Value)
                 {
                     target.Value.Enqueue((long)item);
--- a/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityMember.cs
+++ b/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityMember.cs
@@ -1,0 +1,30 @@
+﻿using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Riok.Mapperly.Descriptors.Mappings;
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity;
+
+/// <summary>
+/// Represents a call to EnsureCapacity on a collection where both the source and targets sizes are accessible.
+/// </summary>
+/// <remarks>
+/// <code>
+/// target.EnsureCapacity(source.Length + target.Count);
+/// </code>
+/// </remarks>
+public class EnsureCapacityMember : EnsureCapacity
+{
+    private readonly string _targetAccessor;
+    private readonly string _sourceAccessor;
+
+    public EnsureCapacityMember(string targetAccessor, string sourceAccessor)
+    {
+        _targetAccessor = targetAccessor;
+        _sourceAccessor = sourceAccessor;
+    }
+
+    public override StatementSyntax Build(TypeMappingBuildContext ctx, ExpressionSyntax target)
+    {
+        return EnsureCapacityStatement(target, MemberAccess(ctx.Source, _sourceAccessor), MemberAccess(target, _targetAccessor));
+    }
+}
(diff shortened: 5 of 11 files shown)


src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Riok.Mapperly.Helpers;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
+
+public class EnsureCapacityInfo
```
latonz, line 11: This code base usually follows one type per file approach.

To match existing architecture convert these properties to private fields and create an exposed method `BuildEnsureCapacity()` which creates the syntax (content of `TryBuildEnsureCapacityStatement`) (similar to how the object factories and mappings work).

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,131 @@
...
+    private static bool TryGetNonEnumeratedCount(ITypeSymbol value, WellKnownTypes types, [NotNullWhen(true)] out string? expression)
+    {
+        if (value.IsArrayType())
+        {
+            expression = "Length";
+            return true;
+        }
+        if (value.HasImplicitInterfaceProperty(types.ICollectionT, "Count") || value.HasImplicitInterfaceProperty(types.IReadOnlyCollectionT, "Count"))
+        {
+            expression = "Count";
```
latonz, line 124: Introduce const's and probably nameofs for these strings.

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            expression = EnsureCapacityStatement(target, MemberAccess(ctx.Source, state.SourceAccessor), MemberAccess(target, state.TargetAccessor));
+            return true;
+        }
+
+        // generate check for runtime type, calling EnsureCapacity if it is a collection/readonlyCollection
+        // this is used if source is IEnumerable
+        var targetCount = MemberAccess(target, state.TargetAccessor);
+
+        var ifCollection = IfIsTypeEnsureCapacityStatement("collection", state.CollectionType!, ctx, target, targetCount);
```
latonz, line 89: If you use named identifiers in the generated code always make sure to use `ctx.NameBuilder.New` to ensure that the name is unique.

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        // if source accessor is avaailable then create simple EnsureCapacity call
+        if (state.SourceAccessor is not null)
+        {
+            expression = EnsureCapacityStatement(target, MemberAccess(ctx.Source, state.SourceAccessor), MemberAccess(target, state.TargetAccessor));
+            return true;
+        }
+
+        // generate check for runtime type, calling EnsureCapacity if it is a collection/readonlyCollection
+        // this is used if source is IEnumerable
```
latonz, line 86: If the target framework supports `Enumerable.TryGetNonEnumeratedCount`, I'd use that instead. This can be found out if the `TryGetNonEnumeratedCount` method symbol is available on the enumerable type.
Probably use a type switch / pattern matching as fallback.

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,131 @@
...
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
+
+public class EnsureCapacityInfo
+{
+    public string? SourceAccessor { get; set; }
+    public string TargetAccessor { get; set; } = null!;
+    public TypeSyntax? CollectionType { get; set; }
```
latonz, line 15: I'd stick to the symbols here instead of using the syntax and just build the typesyntax in the build method, if it is required there.

test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
```diff
@@ -565,6 +601,52 @@ public void EnumerableToCreatedQueue()
...
+            "A",
+            "B",
+            "class A { public int[] Value { get; } }",
+            "class B { public Stack<long> Value { get; set; } }");
+        TestHelper.GenerateMapper(source)
+            .Should()
+            .HaveMethodBody("MapToStack",
+                """
+                var target = new System.Collections.Generic.Stack<long>();
+                target.EnsureCapacity(source.Length + target.Count);
```
latonz, line 616: This is not yet optimal, the built in collections all have a ctor accepting a capacity parameter which would reduce the array allocations and could be used by Mapperly. However, this could also be adressed in a separate PR.

latonz (changes requested):
Thank you for another great contribution 🎉 

I'd create a new namespace `Riok.Mapperly.Descriptors.Enumerables` for these two classes (`EnsureCapacityInfo` and `EnsureCapacityBuilder`)

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/EnsureCapacityBuilder.cs
TimothyMakkison: I'd thought of doing that but wasn't sure how to check which framework is being used. I could check if Enumerable has the corresponding method.

test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
TimothyMakkison: > However, this could also be adressed in a separate PR.

I'd planned on updating the issue after completing this PR. I was going to look at how you did this with dictionaries.

What concrete types is this needed for? Types like `Stack<T>`, `Queue<T>` have constructors that take `IEnumerable<T>`  so I was going to use that where possible.

latonz: Sounds good to me. `Stack<T>`, `Queue<T>` are all, as I think `List<T>` and co are always mapped to via linq.

TimothyMakkison: Should I make those changes a separate PR?

latonz: I'd do it separately if possible.

TimothyMakkison:
- Added version specific `TryGetNonEnumeratedCount` - I've updated the integration test to test this.
- Only use symbols until ensure capacity is built.
- Refactored `EnsureCapacityBuilder` into 3 different versions - wasnt a fan of creating 5 files but it felt better than using ad-hoc discriminant union. I can convert it back to the previous version if you prefer.
- All generated code uses the global type.

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilderIsType.cs
```diff
@@ -0,0 +1,51 @@
...
+    public EnsureCapacityBuilderIsType(string targetAccessor, INamedTypeSymbol collectionType, INamedTypeSymbol readonlyCollectionType)
+    {
+        _targetAccessor = targetAccessor;
+        _collectionType = collectionType;
+        _readonlyCollectionType = readonlyCollectionType;
+    }
+
+    public StatementSyntax BuildEnsureCapacityStatement(TypeMappingBuildContext ctx, ExpressionSyntax target)
+    {
+        var targetCount = MemberAccess(target, _targetAccessor);
```
latonz, line 28: To improve the readability add comments with how the generated code looks like.

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            expression = CountMethodName;
+            return true;
+        }
+
+        expression = null;
+        return false;
+    }
+
+    public static ExpressionStatementSyntax EnsureCapacityStatement(ExpressionSyntax target, ExpressionSyntax sourceCount, ExpressionSyntax targetCount)
```
latonz, line 78: I'd move this syntax conversion into an abstract base class EnsureCapacity to follow the concept of MappingBuilders and Mappings.

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -74,18 +75,24 @@ public static class EnumerableMappingBuilder
...
         // create a foreach loop with add calls if source is not an array
         // and  ICollection.Add(T): void is implemented and not explicit
         // ensures add is not called and immutable types
         if (!ctx.Target.IsArrayType() && ctx.Target.HasImplicitInterfaceMethod(ctx.Types.ICollectionT, AddValueMethodName))
-            return new ForEachAddEnumerableExistingTargetMapping(ctx.Source, ctx.Target, elementMapping, AddValueMethodName);
+            return CreateForEach(AddValueMethodName);
 
         return null;
+
+        ForEachAddEnumerableExistingTargetMapping CreateForEach(string propertyName)
```
latonz, line 91: Is the parameter name `propertyName` correct here?

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -103,6 +103,29 @@ internal static bool HasImplicitInterfaceMethod(this ITypeSymbol symbol, INamedT
         return !methodInterImplementaton.ExplicitInterfaceImplementations.Any();
     }
 
+    internal static bool HasImplicitInterfaceProperty(this ITypeSymbol symbol, INamedTypeSymbol inter, string propertyName)
```
latonz, line 106: Do we need the limitation of properties here?
I'd probably prefer the approach of https://github.com/riok/mapperly/pull/341#discussion_r1167471503.

test/Riok.Mapperly.IntegrationTests/Dto/TestObjectDto.cs
```diff
@@ -22,7 +22,7 @@ public TestObjectDto(int ctorValue, int unknownValue = 10, int ctorValue2 = 100)
         public int IntInitOnlyValue { get; init; }
 
 #if NET7_0_OR_GREATER
-        public required int RequiredValue { get; init; }
+        public int RequiredValue { get; init; }
```
latonz, line 25: Mhmmm, is this change expected?

test/Riok.Mapperly.IntegrationTests/Models/TestObject.cs
```diff
@@ -21,7 +21,9 @@ public TestObject(int ctorValue, int unknownValue = 10, int ctorValue2 = 100)
         public int IntInitOnlyValue { get; init; }
 
 #if NET7_0_OR_GREATER
-        public required int RequiredValue { get; init; }
+
+        public int RequiredValue { get; init; }
```
latonz, line 25: Mhmmm, is this change expected?

test/Riok.Mapperly.IntegrationTests/Dto/TestObjectDto.cs
```diff
@@ -53,6 +53,8 @@ public TestObjectDto(int ctorValue, int unknownValue = 10, int ctorValue2 = 100)
 
         public TestObjectNestedDto[]? NullableReadOnlyObjectCollection { get; set; }
 
+        public Stack<int> EnsureCapacityValue { get; set; } = new();
```
latonz, line 56: Is this separate test property really required? What additional test case does it test?

latonz (changes requested):
I like the new approach. 
I'd adjust the names to the general concept of Mapperly with MappingBuilders and Mappings:
> The idea is to use a MappingBuilder to construct a Mapping object. The Mapping then should only transform the given information into a syntax. The MappingBuilder should provide all the required information for the Mapping to do only the transformation.

I'd rename the `EnsureCapacityBuilder*` to `EnsureCapacity*` (as they are not builder in the sense of Mapperly). However, the `EnsureCapacityHelper` would be a builder.
`IEnusreCapacityBuilder` => `EnsureCapacity` (abstract base class)
`EnsureCapacityHelper` => `EnsureCapacityBuilder`
`EnsureCapacityBuilderNonEnumerated` => `EnsureCapacityNonEnumerated`
`EnsureCapacityBuilderSimple` => `EnsureCapacityMember`
`EnsureCapacityBuilderIsType` => `EnsureCapacityWellKnownTypes`

test/Riok.Mapperly.IntegrationTests/Dto/TestObjectDto.cs
TimothyMakkison: `Enumerable.TryGetNonEnumeratedCount` was added in .Net 6+. If a lower version is used a statement using pattern matching is created (`EnsureCapacityBuilderIsType`). I took advantage of the .Net Framework 4.8 integration test to ensure that this works.

I could probably remove `EnsureCapacityBuilderIsType` as its never used in .Net 6+

test/Riok.Mapperly.IntegrationTests/Models/TestObject.cs
TimothyMakkison: mb I used Codemaid for linting and it still doesn't support modern C# 😞

test/Riok.Mapperly.IntegrationTests/Dto/TestObjectDto.cs
latonz: Ah now I got it, this tests a mapping `IEnumerable<string>` => `Stack<int>` (which uses the type pattern matching / TryGetNonEnumeratedCount respectively), `Stack<string>` => `Stack<int>` tests the case for which the source count is available directly.
I'd keep `EnsureCapacityIsType` for now, as Mapperly tries to support .NET Framework as good as possible.

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacity.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity;
+
+public abstract class EnsureCapacity
+{
+    private const string EnsureCapacityName = "EnsureCapacity";
+
+    public abstract StatementSyntax BuildEnsureCapacityStatement(TypeMappingBuildContext ctx, ExpressionSyntax target);
+
+    public static ExpressionStatementSyntax EnsureCapacityStatement(ExpressionSyntax target, ExpressionSyntax sourceCount, ExpressionSyntax targetCount)
```
latonz, line 16: This method should be protected, shouldn't it?

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        value.ImplementsGeneric(types.ICollectionT, CountMethodName, out var _, out var hasCollectionCount);
+        if (hasCollectionCount)
+        {
+            expression = CountMethodName;
+            return true;
+        }
+
+        value.ImplementsGeneric(types.IReadOnlyCollectionT, CountMethodName, out var _, out var hasReadOnlyCount);
```
latonz, line 71: Nit: the `var` keyword in `out var _` is not needed, is it? (same on L64).

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Helpers;
+
+namespace Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity;
+
+public static class EnsureCapacityBuilder
+{
+    private const string EnsureCapacityName = "EnsureCapacity";
+    private const string CountMethodName = nameof(ICollection<object>.Count);
+    private const string LengthMethodName = nameof(Array.Length);
```
latonz, line 11: These are not really method but property names...

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+    private const string TryGetNonEnumeratedCountMethodName = "TryGetNonEnumeratedCount";
+
+    public static EnsureCapacity? TryCreateEnsureCapacityBuilder(ITypeSymbol sourceType, ITypeSymbol targetType, WellKnownTypes types)
```
latonz, line 14: I'd simplify this method's name (it doesn't return a builder). Eg. `TryBuildEnsureCapacity` (would also align with `TryBuildMapping`).

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacity.cs
```diff
@@ -0,0 +1,21 @@
...
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+
+namespace Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity;
+
+public abstract class EnsureCapacity
+{
+    private const string EnsureCapacityName = "EnsureCapacity";
+
+    public abstract StatementSyntax BuildEnsureCapacityStatement(TypeMappingBuildContext ctx, ExpressionSyntax target);
```
latonz, line 14: I'd simplify this method name. There is only one build method in this class which is exposed and since the class is named `EnsureCapacity` it should be clear that it is an ensure capacity statement. IMO `Build` should be simple and understandable.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -79,6 +79,49 @@ internal static IEnumerable<IPropertySymbol> GetAllAccessibleProperties(this ITy
...
+
+        genericIntf = t.AllInterfaces.FirstOrDefault(x => x.IsGenericType && SymbolEqualityComparer.Default.Equals(x.OriginalDefinition, genericInterfaceSymbol));
+
+        if (genericIntf == null)
+        {
+            isExplicit = false;
+            return false;
+        }
+
+        var interfaceSymbol = genericIntf.GetMembers(symbolName).Single();
```
latonz, line 104: TBD: maybe switching to `First` instead of `Single` will be more resilient if a future release adds a method overload for a given method.

test/Riok.Mapperly.IntegrationTests/Models/TestObject.cs
```diff
@@ -22,6 +22,7 @@ public TestObject(int ctorValue, int unknownValue = 10, int ctorValue2 = 100)
 
 #if NET7_0_OR_GREATER
         public required int RequiredValue { get; init; }
+
```
latonz, line 25: Nit: Probably revert this change?

test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
```diff
@@ -631,6 +631,47 @@ public void MapToExistingExplicitAddCustomCollection()
             .HaveSingleMethodBody("");
     }
 
+    [Fact]
+    public void MapToExistingExplicitCountCustomCollection()
+    {
+        // should not create an EnsureCapacity statement when the target has an explicit count
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            "partial void Map(IEnumerable<int> source, B target);",
+            "class B : List<int> { int List<int>.Count { get; } }");
```
latonz, line 640: I don't get this test case, class B extends the List class, but implements the base class property explicit. This does not really work, does it?

test/Riok.Mapperly.IntegrationTests/_snapshots/NET_48/StaticMapperTest.SnapshotGeneratedSource.verified.cs
```diff
@@ -319,7 +323,18 @@ public static partial class StaticTestMapper
             return target;
         }
 
-        private static global::System.Collections.Generic.Stack<int> MapToStack(global::System.Collections.Generic.Stack<string> source)
+        private static global::System.Collections.Generic.Stack<int> MapToStack(global::System.Collections.Generic.IEnumerable<string> source)
+        {
+            var target = new global::System.Collections.Generic.Stack<int>();
+            foreach (var item in source)
```
latonz, line 329: Shouldn't this be the point where the type tests should be emitted? As mentioned in another comment, I'm also ok with removing the `EnsureCapacityIsType` but then this integration test property could also be removed.

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        // if target does not have a count then return null
+        if (!TryGetNonEnumeratedCount(targetType, types, out var targetSizeProperty))
+            return null;
+
+        // if target and source count are known then create a simple EnsureCapacity statement
+        if (TryGetNonEnumeratedCount(sourceType, types, out var sourceSizeProperty))
+            return new EnsureCapacityMember(targetSizeProperty, sourceSizeProperty);
+
+        sourceType.ImplementsGeneric(types.IEnumerableT, out var iEnumerable);
```
latonz, line 34: Instead of asserting `iEnumerable` is non-null later, I'd if + return null here:
```c#
if (!sourceType.ImplementsGeneric(types.IEnumerableT, out var iEnumerable))
    return null;
```

latonz (changes requested):
Looks pretty good 😊 Thank you a lot for your efforts!

A few nitpicks left and then this should be ready to merge 🚀

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -79,6 +79,49 @@ internal static IEnumerable<IPropertySymbol> GetAllAccessibleProperties(this ITy
...
+        {
+            IMethodSymbol methodSymbol => !methodSymbol.ExplicitInterfaceImplementations.Any(),
+            IPropertySymbol propertySymbol => !propertySymbol.ExplicitInterfaceImplementations.Any(),
+            _ => throw new NotImplementedException(),
+        };
+
+        return true;
+    }
+
     internal static bool HasImplicitInterfaceMethod(this ITypeSymbol symbol, INamedTypeSymbol inter, string methodName)
```
latonz, line 125: This could now be simplified a lot, couldn't it? (eg. `symbol.ImplementsGeneric(inter, methodName, out var isExplicit) && !isExplicit`

test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
TimothyMakkison: Yeah this test is terrible for many reasons. I decided not to generate an EnsureCapacity statement when the target type has an `EnsureCapacity` method and an explicit length. This is because I deemed it to be an edge case and a complete pain to handle with `EnsureCapacityIsType`. 

For some reasons I made a (broken) test to ensure it won't generate an `EnsureCapacity` statement under these conditions.

I could either delete this test, fix it, or update the logic to insert a cast to cast the target when its `Count` is explicit.

latonz: I'm okay with deleting this test.

TimothyMakkison:
Turns out `EnsureCapacity` wasn't added until [.Net 6+](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.ensurecapacity?view=net-8.0#system-collections-generic-list-1-ensurecapacity(system-int32)), so `EnsureCapacityIsType` and the integration test are not needed.

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,69 @@
...
+        // if target does not have a count then return null
+        if (!TryGetNonEnumeratedCount(targetType, types, out var targetSizeProperty))
+            return null;
+
+        // if target and source count are known then create a simple EnsureCapacity statement
+        if (TryGetNonEnumeratedCount(sourceType, types, out var sourceSizeProperty))
+            return new EnsureCapacityMember(targetSizeProperty, sourceSizeProperty);
+
+        if (!sourceType.ImplementsGeneric(types.IEnumerableT, out var iEnumerable))
+            return null;
```
latonz, line 35: Is this ever reached? `TryBuildEnsureCapacity` is only called for enumerables, isn't it?

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        var nonEnumeratedCountMethod = types.Enumerable.GetMembers(TryGetNonEnumeratedCountMethodName)
+                                                        .OfType<IMethodSymbol>()
+                                                        .FirstOrDefault(x => x.ReturnType.SpecialType == SpecialType.System_Boolean && x.IsStatic && x.Parameters.Length == 2 && x.IsGenericMethod);
+
+        // if source does not have a count use GetNonEnumeratedCount, calling EnusureCapacity if count is available
+        var typedNonEnumeratedCount = nonEnumeratedCountMethod.Construct(iEnumerable.TypeArguments.ToArray());
```
latonz, line 42: This leads to a null ref exception on .NET Framework... add a null guard:
```c#
if (nonEnumeratedCountMethod == null)
    return null;

...
```

latonz (changes requested):

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
TimothyMakkison: .NET Framework can't get to this section, [line 23](https://github.com/riok/mapperly/pull/312/files/f19efb7c50ed837ab5986cf93b4ac58f64072d96#diff-3f8932b41774a45b65c962c4ae183d71993fe8112d68981d3ab20f2a1d53172fR23) prevents non .NET 6+ code from getting here because they don't have `EnsureCapacity` available yet.

src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
TimothyMakkison: I used ImplementsGeneric to get the typed version of `IEnumerableT`. I originally didn't use this redundant null, but changed it when you [suggested I remove the null suppression](https://github.com/riok/mapperly/pull/312#discussion_r1167799422)

Issue #72, opened 2022-05-11
Improve incremental source generator support


latonz:
With the current approach, almost everything is regenerated all the time. This should be better abstracted to make better use of the incremental source generator. Approaches to be discussed.

Sergio0694:
You'll want to use `ForAttributeWithMetadataName` (requires Roslyn 4.3.1 or greater) to target uses of `[Mapper]` uses. Then you'll need to build a fully comparable model to pass down in the pipeline. Watch out if you have collection types in your model, as eg. `ImmutableArray<T>` is not equatable. If you need references, an example you can look at is our generators in the MVVM Toolkit, you can find them here: https://aka.ms/mvvmtoolkit/dotnet 🙂

latonz:
@Sergio0694 
Thank you for the inputs / help! I started to work on this [here](https://github.com/latonz/mapperly/tree/72-improve-incremental-source-generator-support). However, it needs some more effort to complete.
In the meantime, since Mapperly only generates the implementation of user defined partial methods, couldn't `RegisterImplementationSourceOutput` be used instead of `RegisterSourceOutput` (which should lead to similar performance improvements)?

nieldy:
I'm implementing a couple of source generators myself and came across the same problem, that I was re-generating the source on any change within the same project. One way I found to drastically improve this is to use a comparer that will only evaluate Equals to false if the syntax nodes aren't equal (and ignore the compilation). Sample code is below...

``` cs
    public void Initialize(IncrementalGeneratorInitializationContext context)
    {
        var provider = GetIncrementalValuesProvider(context);
        
        // Combine the CompilationProvider with the transformed data, but add a custom comparer so that
        // differences can be intelligently determined in order to avoid unnecessarily invalidation of
        // the cached generated code. This is vital to avoid lots of unneeded generation in the IDE, as
        // without the comparer any change to any file in the same project will cause a re-generation.
        // See performance tips at https://www.thinktecture.com/en/net/roslyn-source-generators-performance/
        var combined = provider
            .Combine(context.CompilationProvider)
            .WithComparer(GetComparer())
            .Collect()
            .SelectMany((array, _) => array.Distinct());
            
        context.RegisterSourceOutput(combined, (ctx, data) => Execute(ctx, data.Item2, data.Item1));
    }
```

``` cs
/// <summary>
/// Compares a syntax node and compilation combination by ignoring the compilation and only comparing the
[...]

PR #327, opened 2023-04-12
Use `ForAttributeWithMetadataName` 

TimothyMakkison:
`ForAttributeWithMetadataName` was added to target types with attached attributes. Should have a performance boost - (not familiar with source generators).

### DIFF
--- a/src/Riok.Mapperly/MapperGenerator.cs
+++ b/src/Riok.Mapperly/MapperGenerator.cs
@@ -14,41 +14,15 @@ namespace Riok.Mapperly;
 public class MapperGenerator : IIncrementalGenerator
 {
     private const string GeneratedFileSuffix = ".g.cs";
-    private static readonly string _mapperAttributeName = typeof(MapperAttribute).FullName;
+
+    public static readonly string MapperAttributeName = typeof(MapperAttribute).FullName;
 
     public void Initialize(IncrementalGeneratorInitializationContext context)
     {
-        var mapperClassDeclarations = context.SyntaxProvider
-            .CreateSyntaxProvider(
-                static (s, _) => IsSyntaxTargetForGeneration(s),
-                static (ctx, _) => GetSemanticTargetForGeneration(ctx))
-            .WhereNotNull();
+        var mapperClassDeclarations = SyntaxProvider.GetClassDeclarations(context);
 
         var compilationAndMappers = context.CompilationProvider.Combine(mapperClassDeclarations.Collect());
         context.RegisterImplementationSourceOutput(compilationAndMappers, static (spc, source) => Execute(source.Left, source.Right, spc));
-    }
-
-    private static bool IsSyntaxTargetForGeneration(SyntaxNode node)
-        => node is ClassDeclarationSyntax { AttributeLists.Count: > 0 };
-
-    private static ClassDeclarationSyntax? GetSemanticTargetForGeneration(GeneratorSyntaxContext ctx)
-    {
-        var classDeclaration = (ClassDeclarationSyntax)ctx.Node;
-        foreach (var attributeListSyntax in classDeclaration.AttributeLists)
-        {
-            foreach (var attributeSyntax in attributeListSyntax.Attributes)
-            {
-                if (ctx.SemanticModel.GetSymbolInfo(attributeSyntax).Symbol is not IMethodSymbol attributeSymbol)
-                    continue;
-
-                var attributeContainingTypeSymbol = attributeSymbol.ContainingType;
-                var fullName = attributeContainingTypeSymbol.ToDisplayString();
-                if (fullName == _mapperAttributeName)
-                    return classDeclaration;
-            }
-        }
-
-        return null;
     }
 
     private static void Execute(Compilation compilation, ImmutableArray<ClassDeclarationSyntax> mappers, SourceProductionContext ctx)
@@ -59,8 +33,7 @@ public class MapperGenerator : IIncrementalGenerator
 #if DEBUG_SOURCE_GENERATOR
         DebuggerUtil.AttachDebugger();
 #endif
-
-        var mapperAttributeSymbol = compilation.GetTypeByMetadataName(_mapperAttributeName);
+        var mapperAttributeSymbol = compilation.GetTypeByMetadataName(MapperAttributeName);
         if (mapperAttributeSymbol == null)
             return;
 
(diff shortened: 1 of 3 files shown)


TimothyMakkison:
Seems like the pipeline can't find 'GeneratorAttributeSyntaxContext'. I successfully ran the .Net Framework 48 integration test so I'm not sure what's failing

latonz:
> Seems like the pipeline can't find `GeneratorAttributeSyntaxContext`. I successfully ran the .Net Framework 48 integration test so I'm not sure what's failing

This depends on the Roslyn version which is used to run the source generator. AFAIK this depends whether the source generator runs in Visual Sutdio (which uses the Roslyn version which is shipped with VS), via cli (which uses the Roslyn version bundled by de sdk in use), or via other tools. Mapperly already multi targets Roslyn 4.0, 4.4 and 4.5 and provides a constant `ROSLYN4_4_OR_GREATER`. However, I'm not sure if the added complexity is worth it here.

TimothyMakkison:
> This depends on the Roslyn version which is used to run the source generator.

Makes sense forgot about that 😆 

>  I'm not sure if the added complexity is worth it here, wdyt?

I'd hoped this was a quick small change so hadn't given this much thought (haven't benchmarked a SG before). [https://github.com/dotnet/runtime/issues/70754](This PR) claims that roslyn may do up to x40 times less work with this change. The PR did raise some concerns, although I'm not sure if they apply here.

I'm not familiar with source generators so I can't say how beneficial this change would be, I've updated this PR to use `ROSLYN4_4_OR_GREATER`, I'll leave the final decision to you. You could revisit this if performance becomes a problem.

latonz (changes requested):
I'm happy to accept this change 👍 

The biggest performance improvement could be achieved by implementing a "real" incremental source generator, as the current implementation is not really incremental (tracked in https://github.com/riok/mapperly/issues/72). I started working in that quite some time ago, but didn't find the time to complete it.

Do you use Mapperly in a project where Mapperly becomes a performance bottleneck?

TimothyMakkison:
> The biggest performance improvement could be achieved by implementing a "real" incremental source generator, as the current implementation is not really incremental (tracked in #72). I started working in that quite some time ago, but didn't find the time to complete it.

My knowledge of roslyn and source generators are limited to partially rewriting codemaid and reading [these articles](https://www.thinktecture.com/en/net/roslyn-source-generators-performance/). I'm not sure if its possible to do memoisation when passing `Compilation` in a pipeline. I wonder how the dotnet team solve this for their source generators.

You could move the distinct call from `Execute` to the pipeline, assuming it caches it would probably only yield a minor improvement. 🤷 
```C#
var distincted = mapperClassDeclarations.Collect().Select((myObjects, _) => myObjects.Distinct().ToImmutableArray());
var compilationAndMappers = context.CompilationProvider.Combine(distincted);
```

> Do you use Mapperly in a project where Mapperly becomes a performance bottleneck?

Only in a small personal project

TimothyMakkison:
For what it's wortht the `System.Text.Json` does the exact same thing. It looks incredibly difficult to remove compilation while remaining correct. Perhaps a full rewrite could use compilation in the pipeline, get the type symbols, largely building most of the final syntax in type symbol form and then using a custom converter. Good luck 😁

I did think that most of the logic could be removed from `Execute` and turned into a pipeline step. If the resulting `CompilationUnitSyntax` is the same execute wouldn't run for that class node. I'm not sure how much of an improvement this would be and there would still be the problem of `SourceProductionContext`. It looks like it's only used for reporting diagnostics so perhaps it could be removed from `DescriptorBuilder` and some some alternative way of passing diagnostics from the pipeline into execute could be used.

src/Riok.Mapperly/SyntaxProvider/Roslyn4_4_OrGreater.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Riok.Mapperly.Helpers;
+
+namespace Riok.Mapperly.SyntaxProvider;
+
+internal class Roslyn4_4_OrGreater
+{
+#if ROSLYN4_4_OR_GREATER
```
latonz, line 9: Is there a reason why the precompiler directive here is only around the method but not around the entire file as in `Roslyn4_3_OrLower`?

src/Riok.Mapperly/SyntaxProvider/Roslyn4_3_OrLower.cs
```diff
@@ -0,0 +1,42 @@
+﻿#if !ROSLYN4_4_OR_GREATER
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Riok.Mapperly.Helpers;
+
+namespace Riok.Mapperly.SyntaxProvider;
+
+internal class Roslyn4_3_OrLower
```
latonz, line 8: These classes could also be static, couldn't they?

src/Riok.Mapperly/SyntaxProvider/VersionedSyntaxProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Riok.Mapperly.SyntaxProvider;
+
+internal static class VersionedSyntaxProvider
```
latonz, line 6: I'd remove this class and simply name both roslyn version depended classes the same (`SyntaxProvider`). Since only one should be included at a time during compilation this should work.

TimothyMakkison: What should I name the files, they can't both be `SyntaxProvider`. Does conditional compilation in the csproj prevent this issue?

latonz: I'd move them to `Riok.Mapperly` and name the files `SyntaxProvider.Roslyn4.0.cs` and `SyntaxProvider.Roslyn4.4.cs` and the class `SyntaxProvider`.

TimothyMakkison:
Thanks for the help, I'll rebase when the tests pass.

latonz (changes requested):

src/Riok.Mapperly/SyntaxProvider.Roslyn4.0..cs
```diff
@@ -0,0 +1,43 @@
+﻿#if !ROSLYN4_4_OR_GREATER
```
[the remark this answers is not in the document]
TimothyMakkison: How on earth did you spot that 👀

latonz (approved):

PR #340, opened 2023-04-14
Fix crash when object to object mapping

### LINKED ISSUE #313: Source generator hangs (and Visual Studio crashes) when mapping property type Object
**Describe the bug**

Using this example code, will make Visual Studio hang and crash as soon as you add the `Object? Item` property, or `[Mapper]` attribute (I assume it is the source generator in some infinite loop or something):

```csharp
    [Mapper(UseDeepCloning = true)]
    public partial class FiltersViewModelMapper : Mapper
    {
        public partial FiltersViewModel Clone(FiltersViewModel f);

        //uncomment this and it fixes it
        // private object? ObjectToObject(object? t) => t;
    }

    public class FiltersViewModel
    {
        public bool IsActive { get; set; }
        public List<FilterViewModel>? Filters { get; set; }
    }

    public class FilterViewModel
    {
        public string? Title { get; set; }
        public List<FilterItemViewModel>? Items { get; set; }
    }

    public class FilterItemViewModel
    {
        public bool IsSelected { get; set; }
        public object? Item { get; set; } // causing hang and crash
    }
```

The mapper will probably not know what to do with a reference to type Object -- but it should not crash visual studio.
Workaround is adding `private object? ObjectToObject(object? t) => t;` to the mapper so the generator knows what to do.

**Expected behavior**
Probably skipping the `Object? Item` property? Or just setting the reference as I do in my usecase.

- Mapperly Version:  2.7.0
- .NET Version: .NET 7.0.x
- Target Framework: .net7.0
- OS: Win11
- Visual Studio: 17.5.2

TimothyMakkison:
Resolves #313

Resolve ide hanging/crash when object to object mapping when deep clone is enabled. I mentioned my concerns in the issue, is it correct for a simple copy to be used when deep cloning is enabled? Should I instead make this a diagnostic errror or would updating the docs and adding a note to the `UseDeepCloning` attribute be enough?

- Added another branch to `SpecialTypeMappingBuilder` for object->object edge case
- Added object->object and object?->object? tests

### DIFF
--- a/src/Riok.Mapperly/AnalyzerReleases.Shipped.md
+++ b/src/Riok.Mapperly/AnalyzerReleases.Shipped.md
@@ -77,4 +77,5 @@ Rule ID | Category | Severity | Notes
 RMG029  | Mapper   | Error    | Queryable projection mappings do not support reference handling
 RMG030  | Mapper   | Error    | Reference loop detected while mapping to an init only property
 RMG031  | Mapper   | Warning  | Reference loop detected while mapping to a constructor property
-RMG032 | Mapper | Warning | The enum mapping strategy ByName cannot be used in projection mappings
+RMG032  | Mapper   | Warning  | The enum mapping strategy ByName cannot be used in projection mappings
+RMG033  | Mapper   | Info     | Object mapped to another object without deep clone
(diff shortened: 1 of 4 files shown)


latonz (changes requested):
Looks good to me, as you already considered, I'd introduce a diagnostic for this case `object mapped to another object without deep clone` with a default severity of Info.

The commit message indicates that this is a feature. I'd say this is more of a bug fix and therefore the commit message should start with `fix: `

TimothyMakkison:
Keep getting RS2007, where should I put the new rule?

latonz:
> Keep getting RS2007, where should I put the new rule?

I think the format of the unshipped file is missing the release header. However, you can add it to the shipped file directly anyway.

TimothyMakkison:
I was eventually able to add it to the unshipped file 😅
I'll move the rule to shipped under the new release 2.9

latonz:
The upcoming release is 2.8.0 not 2.9.0. You can add it to 2.8.0

PR #341, opened 2023-04-15
fix: Insert a cast to `IDictionary` when an explicit setter is present

### LINKED ISSUE #286: Enumerable / Dictionary mappings do not support explicit interface implementations

TimothyMakkison:
Resolves #286
- Looks for an explicit setter, casting the target dictionary if one is found. - not sure if `GetExplicitIndexer` should be in `DictionaryMappingBuilder` or `ForEachSetDictionaryExistingTargetMapping`
- All casts use the global type
- Added tests

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
@@ -11,6 +11,7 @@ namespace Riok.Mapperly.Descriptors.MappingBuilders;
 public static class DictionaryMappingBuilder
 {
     private const string CountPropertyName = nameof(IDictionary<object, object>.Count);
+    private const string SetterIndexerPropertyName = "set_Item";
 
     private const string ToImmutableDictionaryMethodName = nameof(ImmutableDictionary.ToImmutableDictionary);
     private const string ToImmutableSortedDictionaryMethodName = nameof(ImmutableSortedDictionary.ToImmutableSortedDictionary);
@@ -64,7 +65,8 @@ public static class DictionaryMappingBuilder
             keyMapping,
             valueMapping,
             false,
-            objectFactory: objectFactory);
+            objectFactory: objectFactory,
+            explicitCast: GetExplicitIndexer(ctx));
     }
 
     public static IExistingTargetMapping? TryBuildExistingTargetMapping(MappingBuilderContext ctx)
@@ -87,7 +89,8 @@ public static class DictionaryMappingBuilder
             ctx.Source,
             ctx.Target,
             keyMapping,
-            valueMapping);
+            valueMapping,
+            explicitCast: GetExplicitIndexer(ctx));
     }
 
     private static (ITypeMapping, ITypeMapping)? BuildKeyValueMapping(MappingBuilderContext ctx)
@@ -148,6 +151,14 @@ public static class DictionaryMappingBuilder
         return (enumeratedType.TypeArguments[0], enumeratedType.TypeArguments[1]);
     }
 
+    private static INamedTypeSymbol? GetExplicitIndexer(MappingBuilderContext ctx)
+    {
+        if (ctx.Target.ImplementsGeneric(ctx.Types.IDictionaryT, SetterIndexerPropertyName, out var typedInter, out var isExplicit) && !isExplicit)
+            return null;
+
+        return typedInter;
+    }
+
     private static LinqDicitonaryMapping? ResolveImmutableCollectMethod(MappingBuilderContext ctx, ITypeMapping keyMapping, ITypeMapping valueMapping)
     {
         if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableSortedDictionaryT))
@@ -159,4 +170,5 @@ public static class DictionaryMappingBuilder
 
         return null;
     }
+
 }
--- a/src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/ForEachSetDictionaryExistingTargetMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/ForEachSetDictionaryExistingTargetMapping.cs
@@ -12,25 +12,40 @@ namespace Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
 public class ForEachSetDictionaryExistingTargetMapping : ExistingTargetMapping
 {
     private const string LoopItemVariableName = "item";
+    private const string ExplicitCastVariableName = "targetDict";
     private const string KeyPropertyName = nameof(KeyValuePair<object, object>.Key);
     private const string ValuePropertyName = nameof(KeyValuePair<object, object>.Value);
 
     private readonly ITypeMapping _keyMapping;
     private readonly ITypeMapping _valueMapping;
+    private readonly INamedTypeSymbol? _explicitCast;
 
     public ForEachSetDictionaryExistingTargetMapping(
         ITypeSymbol sourceType,
         ITypeSymbol targetType,
         ITypeMapping keyMapping,
-        ITypeMapping valueMapping)
+        ITypeMapping valueMapping,
+        INamedTypeSymbol? explicitCast)
         : base(sourceType, targetType)
     {
         _keyMapping = keyMapping;
         _valueMapping = valueMapping;
+        _explicitCast = explicitCast;
     }
 
     public override IEnumerable<StatementSyntax> Build(TypeMappingBuildContext ctx, ExpressionSyntax target)
     {
+        if (_explicitCast != null)
+        {
+            var type = FullyQualifiedIdentifier(_explicitCast);
+            var cast = CastExpression(type, target);
+
+            var castedVariable = ctx.NameBuilder.New(ExplicitCastVariableName);
+            target = IdentifierName(castedVariable);
+
+            yield return LocalDeclarationStatement(DeclareVariable(castedVariable, cast));
+        }
+
         var loopItemVariableName = ctx.NameBuilder.New(LoopItemVariableName);
 
         var convertedKeyExpression = _keyMapping.Build(ctx.WithSource(MemberAccess(loopItemVariableName, KeyPropertyName)));
@@ -40,13 +55,10 @@ public class ForEachSetDictionaryExistingTargetMapping : ExistingTargetMapping
             ElementAccess(target, convertedKeyExpression),
             convertedValueExpression);
 
-        return new StatementSyntax[]
-        {
-            ForEachStatement(
+        yield return ForEachStatement(
                 VarIdentifier,
                 Identifier(loopItemVariableName),
                 ctx.Source,
-                Block(ExpressionStatement(assignment))),
-        };
+                Block(ExpressionStatement(assignment)));
     }
 }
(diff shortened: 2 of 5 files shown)


src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -131,4 +134,25 @@ private static (ITypeSymbol, ITypeSymbol)? GetEnumerableKeyValueTypes(MappingBui
 
         return (enumeratedType.TypeArguments[0], enumeratedType.TypeArguments[1]);
     }
+
+    private static INamedTypeSymbol? GetExplicitIndexer(MappingBuilderContext ctx)
+    {
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.IDictionaryT))
```
latonz, line 140: Wouldn't it make sense to create an overload for the existing `SymbolExtensions.ImplementsGeneric` method with the following signature?
```c#
internal static bool ImplementsGeneric(
        this ITypeSymbol t,
        INamedTypeSymbol genericInterfaceSymbol,
        string symbolName,
        [NotNullWhen(true)] out INamedTypeSymbol? genericIntf,
        out bool isExplicit)
```

The same could then be used for the enumerables and replace `HasImplicitInterfaceMethod` in the end.

latonz (changes requested):
Thanks for another great contribution!
The idea is to use a MappingBuilder to construct a Mapping object. The Mapping then should only transform the given information into a syntax. The MappingBuilder should provide all the required information for the Mapping to do only the transformation.

* This does not fix #286 for enumerables, does it?
* This does not fix the case when the source implements `IDictionary<K, V>`, `IReadOnlyDictionary<K, V>` or `IEnumerable<KeyValuePair<K, V>>` explicitly, does it?

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -131,4 +134,25 @@ private static (ITypeSymbol, ITypeSymbol)? GetEnumerableKeyValueTypes(MappingBui
...
+            return null;
+
+        // get typed version of IDictionay<>
+        ctx.Target.ImplementsGeneric(ctx.Types.IDictionaryT, out var typedInter);
+
+        var interfaceMethodSymbol = typedInter!.GetMembers(SetterIndexerPropertyName).OfType<IMethodSymbol>().SingleOrDefault(x => x.Parameters.Length == 2 && x.ReturnsVoid);
```
latonz, line 146: Wouldn't it be more appropriate to use the `this` indexed property instead of the set method?

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/ForEachSetDictionaryExistingTargetMapping.cs
```diff
@@ -12,25 +12,40 @@ namespace Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
...
         : base(sourceType, targetType)
     {
         _keyMapping = keyMapping;
         _valueMapping = valueMapping;
+        _explicitCast = explicitCast;
     }
 
     public override IEnumerable<StatementSyntax> Build(TypeMappingBuildContext ctx, ExpressionSyntax target)
     {
+        if (_explicitCast is not null)
```
latonz, line 38: Nit: I think for null checks this codebase uses `!= null`

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/ForEachSetDictionaryExistingTargetMapping.cs
```diff
@@ -12,25 +12,40 @@ namespace Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
...
+        {
+            var type = FullyQualifiedIdentifier(_explicitCast);
```
latonz, line 40: Could you add a comment how the generated code looks like to make it faster to understand, we started that a while ago and it makes the mappings a lot easier to read.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
TimothyMakkison: I'd have to double check, but I think indexer methods become:
```C#
public V get_Item(K key);
public void set_Item(K key, V value);
```

TimothyMakkison:
> The idea is to use a MappingBuilder to construct a Mapping object. The Mapping then should only transform the given information into a syntax. The MappingBuilder should provide all the required information for the Mapping to do only the transformation.

Gotcha, I'll keep the invocation `GetExplicitIndexer` out of `ForEachDictionary`.

> * This does not fix [Enumerable / Dictionary mappings do not support explicit interface implementations #286](https://github.com/riok/mapperly/issues/286) for enumerables, does it?

Where is this a problem for Enumerables? Arrays aren't affected, I don't think the `IList<T>` indexers are used, foreach will get explicit enumerators and we check for `Add`.

> * This does not fix the case when the source implements `IDictionary<K, V>`, `IReadOnlyDictionary<K, V>` or `IEnumerable<KeyValuePair<K, V>>` explicitly, does it?

Looking at the tests the only thing the source is used for is Enumerating. Do you have an example?

latonz:
Ah yap your totally right, foreach does work even if the `IEnumerable` is implemented explicitly. I had in my mind that the enumerator couldn't be created when `IEnumerable` is implemented explicitly but thats not the case.
So then the only improvement which could be done there would be to support add calls when the interface is implemented explicitly. The dictionary implementation is alright 👍

TimothyMakkison:
> So then the only improvement which could be done there would be to support add calls when the interface is implemented explicitly.

A lot of .Net collections "implement" `ICollection`, not implementing `Add` and making it explicit. ie the Immutable collections, Frozen collections, `Queue<T>`, `Stack<T>`, `LinkedList<T>`, `HashSet<T>`, `SortedSet<T>`, etc. A lot of these types are unlikely to be used with `Mapperly`/some won't reach this section of code #312, but will break if casted.

I created `HasImplicitMethod` to prevent `Add` being called on these types, although in retrospect it would have been simpler to check if Target was a `List<T>`. `SortedSet<T>`, `HashSet<T>` etc and use `Add` accordingly. If I explicitly cast anything implementing `ICollection<T>` we'd have to create a large black list filter.

latonz:
Thank you for the input, this makes totally sense!

However, I don't think removing `HasImplicitMethod` would be any better as then a custom `ICollection<T>` implementation could never be used as a mapping target. This may also be related to https://github.com/riok/mapperly/pull/342#discussion_r1167472826. Don't most of these collections which don't throw on add have ctor with a single enumerable parameter?

TimothyMakkison:
> However, I don't think removing `HasImplicitMethod` would be any better as then a custom `ICollection<T>` implementation could never be used as a mapping target. This may also be related to [#342 (comment)](https://github.com/riok/mapperly/pull/342#discussion_r1167472826). Don't most of these collections which don't throw on add have ctor with a single enumerable parameter?

Good point, #324 would prevent a lot of these cases. Unfortunately the casting problem would still affect `TryBuildExistingTargetMapping` as the constructor version cannot be used when the target collection already exists.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -130,4 +133,25 @@ private static (ITypeSymbol, ITypeSymbol)? GetEnumerableKeyValueTypes(MappingBui
...
         return (enumeratedType.TypeArguments[0], enumeratedType.TypeArguments[1]);
     }
+
+    private static INamedTypeSymbol? GetExplicitIndexer(MappingBuilderContext ctx)
+    {
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.IDictionaryT))
+            return null;
+
+        // get typed version of IDictionay<>
+        ctx.Target.ImplementsGeneric(ctx.Types.IDictionaryT, out var typedInter);
```
latonz, line 143: Instead of the equals on L139, couldn't we just return null here if `ImplementsGeneric` return false? Event the new overload with the `isExplicit` out param could simplify this a lot.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
latonz: I think the indexer property is available under the name `this[]`.

latonz (changes requested):

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
TimothyMakkison: What's the advantage of getting the indexer over the setter? We only care about the setter.
With the indexer property we'd access `SetMethod` getting the exact same value as if we'd used "set_item".

```C#
// var targetIndexer = ctx.Target.GetMembers("this[]").FirstOrDefault(); // is null not sure why, might have to add types
var indexer = typedInter.GetMembers("this[]").First(); // get indexer with typed interface
var setter = indexer.SetMethod;
// here setter is the same as ctx.Target.GetMembers("set_Item").First();
var same = SymbolEqualityComparer.Default.Equals(indexer.SetMethod, setter); // true

```

latonz (dismissed):

latonz:
I rebased the branch via GH and only then saw that there are two commits. Could you squash those two commits together into one feature commit? 😊

latonz (dismissed):

TimothyMakkison:
Okay I'll revase and squish later :part

latonz (approved):

latonz:
@TimothyMakkison this looks ready to merge to me, are you okay with merging?

TimothyMakkison:
Thanks for your help @latonz, you can merge this request!

PR #342, opened 2023-04-15
feat: use IEnumerable constructor when possible (`Queue<T>` and `Stack<T>`)

TimothyMakkison:
[Resolves](https://github.com/riok/mapperly/pull/312#discussion_r1165926743)
- If target is does not exist and target is  `Queue<T>` or `Stack<T>` then use the `IEnumerable` constructors
- Uses global types
- Added 2 unit tests

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
@@ -63,6 +63,10 @@ public static class EnumerableMappingBuilder
         if (immutableLinqMapping is not null)
             return immutableLinqMapping;
 
+        // if target is a type that takes IEnumerable in its constructor
+        if (HasEnumerableConstructor(ctx, elementMapping.TargetType))
+            return BuildLinqConstructorMapping(ctx, elementMapping);
+
         return ctx.IsExpression
             ? null
             : BuildCustomTypeMapping(ctx, elementMapping);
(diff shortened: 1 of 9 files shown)


src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -63,6 +63,11 @@ public static class EnumerableMappingBuilder
         if (immutableLinqMapping is not null)
             return immutableLinqMapping;
 
+        // if target is a type that takes IEnumerable in its constructor
+        if (ctx.Target.ImplementsGeneric(ctx.Types.StackT, out _)
+            || ctx.Target.ImplementsGeneric(ctx.Types.QueueT, out _))
```
latonz, line 68: On a second thought: couldn't we generalize this to check if such a ctor is available instead of checking only those types?

latonz (changes requested):

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
TimothyMakkison: It could cause some unexpected behaviour. It can be slower to use constructors in some cases. For example `List<T>` does not check for `IListProvider` so a foreach loop with `EnsureCapacity` might be faster (`List<T>` can't hit this section of code).

test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
```diff
@@ -704,17 +700,49 @@ public void EnumerableToCreatedQueue()
             "A",
             "B",
             "class A { public IEnumerable<int> Value { get; } }",
-            "class B { public Queue<long> Value { get; set; } }");
+            "class B { public Stack<int> Value { get; set; } }");
```
latonz, line 703: Shouldn't this be a queue?

latonz (changes requested):

latonz (approved):

latonz (approved):

latonz (approved):

CommonGuy (approved):

Issue #329, opened 2023-04-12
Analyzer should ignore readonly properties in target


marcoatribeiro:
When creating a mapping method from `Source` to `Target` classes, where `Target` has a readonly property, analyzer shows the **RMG012** warning, which doesn't make sense, as it would not be possible to map to a readonly property anyway.

**To Reproduce**
1. Create `Source` class with any set of properties (e.g. `FirstName` and `LastName`)
2. Create `Target` class with all the properties from `Source` plus a readonly property (e.g. `FullName => FirstName + LastName`)
3. Create a `Mapper` class and a method to map from `Source` to `Target`
4. Check the analyzer displays _RMG012_ warning

**Expected behavior**
As the target property is readonly (and, thus, cannot be assigned to), the aforementioned warning should not be displayed.

**Environment**
- Mapperly Version: 2.7.0
- .NET Version: .NET 7
- IDE: Visual Studio v17.5.4
- OS: Windows 11

marcoatribeiro:
Hi, @latonz !

I just upgraded Mapperly to version 2.8.0 and, although issue #345 is part of this release, this problem still occurs, as can be seen below:

![image](https://github.com/riok/mapperly/assets/35035369/332a64f6-0dd4-4a85-8ac7-2396eda2b944)

Classes `PersonEntity` and `PersonDto` have settings just like I described above, with both having `FirstName` and `LastName` properties and `PersonEntity` only having `FullName` (readonly) property:

```csharp
public class PersonEntity
{
    public string FirstName { get; set; } = string.Empty;
    public string LastName { get; set; } = string.Empty;
    public string FullName => $"{FirstName} {LastName}";
}
```

```csharp
public class PersonDto
{
    public string FirstName { get; set; } = string.Empty;
    public string LastName { get; set; } = string.Empty;
}
```

Can you please advise?

TimothyMakkison:
Hey @marcoatribeiro, is this problem present in the prerelease version?

marcoatribeiro:
Hi, @TimothyMakkison and @latonz!

Please disregard my previous comment. I just check it again (with version 2.8.0) and it is working as expected now.
Yesterday I tested right after I updated the package, so, most likely the alert was still showing because VS was still referencing the previous version (even though I had built the whole solution).

PR #345, opened 2023-04-16
fix: do not diagnostic property not found if target is read only

latonz:
Closes https://github.com/riok/mapperly/issues/329

### DIFF
--- a/test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
@@ -236,7 +236,7 @@ public class EnumerableTest
     }
 
     [Fact]
-    public void ArrayToArrayOfReadonlyStruct()
+    public void ArrayToArrayOfReadOnlyStruct()
     {
         var source = TestSourceBuilder.Mapping(
             "A[]",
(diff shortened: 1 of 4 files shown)


test/Riok.Mapperly.Tests/Mapping/EnumerableTest.cs
```diff
@@ -848,102 +848,6 @@ public Task MapToReadOnlyCollectionProperty()
...
     }
 
-    [Fact]
-    public void EnumerableToReadonlyImmutableListShouldDiagnostic()
-    {
-        var source = TestSourceBuilder.Mapping(
-            "A",
-            "B",
-            "class A { public IEnumerable<int> Value { get; } }",
-            "class B { public System.Collections.Immutable.ImmutableList<int> Value { get; } }");
```
CommonGuy, line 858: IMO these test cases should still emit a warning. Having a matching source and target property, but the target property is read-only/immutable, is probably a bug in most cases.

Would still emit the warning by default. Users can ignore it by ignoring the target property via configuration

latonz: Added a warning when trying to map to a readOnly immutable collection type.

CommonGuy (approved):

PR #351, opened 2023-04-17
feat: add support for Immutable dictionaries

### LINKED ISSUE #250: Support System.Collections.Immutable types

TimothyMakkison:
Resolve the last two types from #250 
- Added support for `ImmutableDictionary` and `ImmutableSortedDictionary`
- Added guards to `DictionaryMappingBuilder` to prevent existing immutable types from being assigned to in a foreach loop. - I did not copy the approach of checking for a implicit setter from the Immutable Enumerable types as this would conflict with #341.
- Added tests.

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
@@ -1,3 +1,4 @@
+using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Descriptors.Mappings;
@@ -10,6 +11,9 @@ namespace Riok.Mapperly.Descriptors.MappingBuilders;
 public static class DictionaryMappingBuilder
 {
     private const string CountPropertyName = nameof(IDictionary<object, object>.Count);
+
+    private const string ToImmutableDictionaryMethodName = nameof(ImmutableDictionary.ToImmutableDictionary);
+    private const string ToImmutableSortedDictionaryMethodName = nameof(ImmutableSortedDictionary.ToImmutableSortedDictionary);
 
     public static ITypeMapping? TryBuildMapping(MappingBuilderContext ctx)
     {
@@ -37,6 +41,11 @@ public static class DictionaryMappingBuilder
                 targetDictionarySymbol,
                 dictionaryObjectFactory);
         }
+
+        // if target is an immutable dictionary then use LinqDictionaryMapper
+        var immutableLinqMapping = ResolveImmutableCollectMethod(ctx, keyMapping, valueMapping);
+        if (immutableLinqMapping != null)
+            return immutableLinqMapping;
 
         // the target is not a well known dictionary type
         // it should have a an object factory or a parameterless public ctor
@@ -66,6 +75,14 @@ public static class DictionaryMappingBuilder
         if (BuildKeyValueMapping(ctx) is not var (keyMapping, valueMapping))
             return null;
 
+        // if target is an immutable dictionary then don't create a foreach loop
+        if (ctx.Target.OriginalDefinition.ImplementsGeneric(ctx.Types.IImmutableDictionaryT, out _))
+        {
+            ctx.ReportDiagnostic(DiagnosticDescriptors.CannotMapToReadOnlyMember);
+            return null;
+        }
+
+        // add values to dictionary by setting key values in a foreach loop
         return new ForEachSetDictionaryExistingTargetMapping(
             ctx.Source,
             ctx.Target,
@@ -129,5 +146,17 @@ public static class DictionaryMappingBuilder
             return null;
 
         return (enumeratedType.TypeArguments[0], enumeratedType.TypeArguments[1]);
+    }
+
+    private static LinqDicitonaryMapping? ResolveImmutableCollectMethod(MappingBuilderContext ctx, ITypeMapping keyMapping, ITypeMapping valueMapping)
+    {
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableSortedDictionaryT))
+            return new LinqDicitonaryMapping(ctx.Source, ctx.Target, ctx.Types.ImmutableSortedDictionary.GetStaticGenericMethod(ToImmutableSortedDictionaryMethodName)!, keyMapping, valueMapping);
+
+        // if taget is an ImmutableDictionary or implements interface IImmutableDictionary
+        if (ctx.Target.OriginalDefinition.ImplementsGeneric(ctx.Types.IImmutableDictionaryT, out _))
+            return new LinqDicitonaryMapping(ctx.Source, ctx.Target, ctx.Types.ImmutableDictionary.GetStaticGenericMethod(ToImmutableDictionaryMethodName)!, keyMapping, valueMapping);
+
+        return null;
     }
 }
--- a/src/Riok.Mapperly/Descriptors/Mappings/LinqDictionaryMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/LinqDictionaryMapping.cs
@@ -1,0 +1,56 @@
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Descriptors.Mappings;
+
+/// <summary>
+/// Represents an enumerable mapping which works by using linq (select + collect).
+/// </summary>
+public class LinqDicitonaryMapping : TypeMapping
+{
+    private const string LambdaParamName = "x";
+
+    private const string KeyPropertyName = nameof(KeyValuePair<object, object>.Key);
+    private const string ValuePropertyName = nameof(KeyValuePair<object, object>.Value);
+
+    private readonly IMethodSymbol _collectMethod;
+    private readonly ITypeMapping _keyMapping;
+    private readonly ITypeMapping _valueMapping;
+
+    public LinqDicitonaryMapping(
+        ITypeSymbol sourceType,
+        ITypeSymbol targetType,
+        IMethodSymbol collectMethod,
+        ITypeMapping keyMapping,
+        ITypeMapping valueMapping)
+        : base(sourceType, targetType)
+    {
+        _collectMethod = collectMethod;
+        _keyMapping = keyMapping;
+        _valueMapping = valueMapping;
+    }
+
+    public override ExpressionSyntax Build(TypeMappingBuildContext ctx)
+    {
+        var lambdaParamName = ctx.NameBuilder.New(LambdaParamName);
+
+        // if key and value types do not change then use a simple call
+        // ie: source.ToImmutableDictionary();
+        if (_keyMapping.IsSynthetic && _valueMapping.IsSynthetic)
+            return StaticInvocation(_collectMethod, ctx.Source);
+
+        // create expressions mapping the key and value and then create the final expression
+        // ie: source.ToImmutableDictionary(x => x.Key, x=> (int)x.Value);
+        var keyMapExpression = _keyMapping.Build(ctx.WithSource(MemberAccess(lambdaParamName, KeyPropertyName)));
+        var keyExpression = SimpleLambdaExpression(Parameter(Identifier(lambdaParamName)))
+            .WithExpressionBody(keyMapExpression);
+
+        var valueMapExpression = _valueMapping.Build(ctx.WithSource(MemberAccess(lambdaParamName, ValuePropertyName)));
+        var valueExpression = SimpleLambdaExpression(Parameter(Identifier(lambdaParamName)))
+            .WithExpressionBody(valueMapExpression);
+
+        return StaticInvocation(_collectMethod, ctx.Source, keyExpression, valueExpression);
+    }
+}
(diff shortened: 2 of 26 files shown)


src/Riok.Mapperly/Descriptors/Mappings/LinqDictionaryMapping.cs
```diff
@@ -0,0 +1,55 @@
...
+
+/// <summary>
+/// Represents an enumerable mapping which works by using linq (select + collect).
+/// </summary>
+public class LinqDicitonaryMapping : TypeMapping
+{
+    private const string LambdaParamName = "x";
+
+    private const string KeyValueKeyPropertyName = nameof(KeyValuePair<object, object>.Key);
+    private const string KeyValueValuePropertyName = nameof(KeyValuePair<object, object>.Value);
```
latonz, line 16: I would simplify these names to `KeyPropertyName` and `ValuePropertyName`

src/Riok.Mapperly/Descriptors/Mappings/LinqDictionaryMapping.cs
```diff
@@ -0,0 +1,55 @@
...
+        _valueMapping = valueMapping;
+    }
+
+    public override ExpressionSyntax Build(TypeMappingBuildContext ctx)
+    {
+        var lambdaParamName = ctx.NameBuilder.New(LambdaParamName);
+
+        // if key and value types do not change then use a simple call
+        if (SymbolEqualityComparer.Default.Equals(_keyMapping.TargetType, _keyMapping.SourceType)
+            && SymbolEqualityComparer.Default.Equals(_valueMapping.TargetType, _valueMapping.SourceType))
```
latonz, line 41: Instead of checking the type, it should be ok to check `IsSynthetic` on the mapping.

test/Riok.Mapperly.Tests/Mapping/DictionaryTest.cs
```diff
@@ -250,6 +250,90 @@ public void ReadOnlyDictionaryToCustomTypeReadOnlyReadOnlyDictionaryShouldIgnore
...
+    public void DictionaryToImmutableDictionaryCasted()
+    {
+        var source = TestSourceBuilder.Mapping(
+            "Dictionary<string, string>",
+            "System.Collections.Immutable.ImmutableDictionary<int, int>");
+        TestHelper.GenerateMapper(source)
+            .Should()
+            .HaveMapMethodBody(
+                """
+                return global::System.Collections.Immutable.ImmutableDictionary.ToImmutableDictionary(source, x => int.Parse(x.Key), x => int.Parse(x.Value));
```
latonz, line 291: This is not really a casted value as indicated by the name of the test.

test/Riok.Mapperly.Tests/Mapping/DictionaryTest.cs
```diff
@@ -250,6 +250,90 @@ public void ReadOnlyDictionaryToCustomTypeReadOnlyReadOnlyDictionaryShouldIgnore
...
+    public void DictionaryToImmutableDictionaryCastedValue()
+    {
+        var source = TestSourceBuilder.Mapping(
+            "Dictionary<string, string>",
+            "System.Collections.Immutable.ImmutableDictionary<string, int>");
+        TestHelper.GenerateMapper(source)
+            .Should()
+            .HaveMapMethodBody(
+                """
+                return global::System.Collections.Immutable.ImmutableDictionary.ToImmutableDictionary(source, x => x.Key, x => int.Parse(x.Value));
```
latonz, line 305: This is not really a casted value as indicated by the name of the test.

test/Riok.Mapperly.Tests/Mapping/DictionaryTest.cs
```diff
@@ -250,6 +250,90 @@ public void ReadOnlyDictionaryToCustomTypeReadOnlyReadOnlyDictionaryShouldIgnore
...
+                """);
+    }
+
+    [Fact]
+    public void EnumerableToReadonlyImmutableDictionary()
```
latonz, line 310: Add the ShouldDiagnostic suffix.

Nit: `Readonly` => `ReadOnly`

test/Riok.Mapperly.Tests/Mapping/DictionaryTest.cs
```diff
@@ -250,6 +250,90 @@ public void ReadOnlyDictionaryToCustomTypeReadOnlyReadOnlyDictionaryShouldIgnore
...
+            "class A { public Dictionary<string, string> Value { get; } }",
+            "class B { public System.Collections.Immutable.ImmutableDictionary<string, string> Value { get; } }");
+
+        TestHelper.GenerateMapper(source, TestHelperOptions.AllowInfoDiagnostics)
+            .Should()
+            .HaveDiagnostic(new(DiagnosticDescriptors.SourceMemberNotFound));
+    }
+
+    [Fact]
+    public void EnumerableToReadonlyImmutableSortedDictionary()
```
latonz, line 324: Add the ShouldDiagnostic suffix.
Nit: `Readonly` => `ReadOnly`

src/Riok.Mapperly/Descriptors/Mappings/LinqDictionaryMapping.cs
```diff
@@ -0,0 +1,55 @@
...
+        var keyMapExpression = _keyMapping.Build(ctx.WithSource(MemberAccess(lambdaParamName, KeyValueKeyPropertyName)));
+        var keyExpression = SimpleLambdaExpression(Parameter(Identifier(lambdaParamName)))
+            .WithExpressionBody(keyMapExpression);
+
+        var valueMapExpression = _valueMapping.Build(ctx.WithSource(MemberAccess(lambdaParamName, KeyValueValuePropertyName)));
+        var valueExpression = SimpleLambdaExpression(Parameter(Identifier(lambdaParamName)))
+            .WithExpressionBody(valueMapExpression);
+
+        return StaticInvocation(_collectMethod, ctx.Source, keyExpression, valueExpression);
+    }
```
latonz, line 54: Add a comment with a sample of the generated code.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -38,6 +42,13 @@ public static class DictionaryMappingBuilder
                 dictionaryObjectFactory);
         }
 
+        // if target is an immutable dictionary then use LinqDictionaryMapper
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableDictionaryT))
```
latonz, line 46: Couldn't we add the `IImmutableDictionary<K, V>` to the check here?

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -38,6 +42,13 @@ public static class DictionaryMappingBuilder
...
+            return new LinqDicitonaryMapping(ctx.Source, ctx.Target, ResolveStaticMethod(ctx.Types.ImmutableDictionary, ToImmutableDictionaryMethodName)!, keyMapping, valueMapping);
```
latonz, line 47: Instead of asserting non null, I'd add this to the if condition.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -38,6 +42,13 @@ public static class DictionaryMappingBuilder
...
+
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableSortedDictionaryT))
+            return new LinqDicitonaryMapping(ctx.Source, ctx.Target, ResolveStaticMethod(ctx.Types.ImmutableSortedDictionary, ToImmutableSortedDictionaryMethodName)!, keyMapping, valueMapping);
```
latonz, line 50: Instead of asserting non null, I'd add this to the if condition.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -66,6 +77,11 @@ public static class DictionaryMappingBuilder
         if (BuildKeyValueMapping(ctx) is not var (keyMapping, valueMapping))
             return null;
 
+        // if target is an immutable dictionary then don't a foreach loop
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableDictionaryT)
+            || SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableSortedDictionaryT))
```
latonz, line 82: Instead of checking the immutable types, couldn't we check if the `IImmutableDictionary<TKey, TValue>` is implemented? 🤔 
Also I think we should use the same approach as for the enumerables. Eg. if we do the denylist here, we should adjust the enumerables too.

latonz (changes requested):

src/Riok.Mapperly/Descriptors/Mappings/LinqDictionaryMapping.cs
TimothyMakkison: I meant to ask about this,, for some reason `_keyMapping` and `_valueMapping` `IsSynthetic` kept returning true.

latonz: `IsSynthetic` is truethy if the mapping does not produce any code and could be omitted. Do you observed `IsSynthetic` beeing true for a mapping which produces code?

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
TimothyMakkison: Would this change anythng? We'd then have to check if it's an `ImmutableDictionary` or `ImmutableSortedDictionary` to pass the correct collect method name.

I could extract the logic like `TryBuildImmutableLinqMapping` in `EnumerableMappingBuilder`.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
TimothyMakkison: I assume you want something like:
```C#
if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableSortedDictionaryT) 
&& ctx.Types.ImmutableSortedDictionary.GetStaticGenericMethod(ToImmutableSortedDictionaryMethodName) is IMethodSymbol method)
            return new LinqDicitonaryMapping(ctx.Source, ctx.Target, method, keyMapping, valueMapping);
```

Currently if the condition is true then we know 100% that it will `ResolveStaticMethod` will not return null. I'm not a fan of  putting logic in a condition; It is misleading because it makes it seem like it could fail. I could change `ResolveStaticMethod` to throw if returning null or keep the nullable suppression operator.

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
TimothyMakkison: Good point I'll make the change.

 I'll add the denylist in #341

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -66,6 +75,10 @@ public static class DictionaryMappingBuilder
         if (BuildKeyValueMapping(ctx) is not var (keyMapping, valueMapping))
             return null;
 
+        // if target is an immutable dictionary then don't create a foreach loop
+        if (ctx.Target.OriginalDefinition.ImplementsGeneric(ctx.Types.IImmutableDictionaryT, out _))
+            return null;
```
latonz, line 80: I think a diagnostic here would be super helpful, we recently implemented  a similar thing for enumerables: 
https://github.com/riok/mapperly/pull/345/files#diff-d0e8d4a6f1b4bc7cabf6bf20484b1ff8e57a8fde79753fa4bf2d563df38965dbR94-R97

src/Riok.Mapperly/Descriptors/MappingBuilders/DictionaryMappingBuilder.cs
```diff
@@ -130,4 +143,15 @@ private static (ITypeSymbol, ITypeSymbol)? GetEnumerableKeyValueTypes(MappingBui
 
         return (enumeratedType.TypeArguments[0], enumeratedType.TypeArguments[1]);
     }
+
+    private static LinqDicitonaryMapping? ResolveImmutableCollectMethod(MappingBuilderContext ctx, ITypeMapping keyMapping, ITypeMapping valueMapping)
+    {
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableDictionaryT))
+            return new LinqDicitonaryMapping(ctx.Source, ctx.Target, ctx.Types.ImmutableDictionary.GetStaticGenericMethod(ToImmutableDictionaryMethodName)!, keyMapping, valueMapping);
+
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.ImmutableSortedDictionaryT) && ctx.Types.ImmutableSortedDictionary.GetStaticGenericMethod(ToImmutableSortedDictionaryMethodName) is IMethodSymbol method)
```
latonz, line 152: Now we have the styles mixed (one time non-null assertion, one time in the if condition). You convinced me to assert non-null. Could you move this to match the stile on L150?

latonz (changes requested):
Thank you for the updates!
Sorry for the late feedback. 2-3 smaller points to discuss / fix, otherwise it looks really good to me 😊 
Could you then also rebase (I think there may be some conflicts 🤔) and squash into one feature commit? Thanks!

latonz (approved):

PR #343, opened 2023-04-15
chore: add chsarpier and husky

latonz:

PR #361, opened 2023-04-19
fix: add `EnsureCapacity` for dictionary types

### LINKED ISSUE #251: Dictionary and Enumerable Mappings: Call `EnsureCapacity` if source count is known and target has a `EnsureCapacity(int)` method

TimothyMakkison:
Resolves #251

- Modifed a bit of #312 to work with dictionaries. 
- Adjusted `ImplementsGeneric` to support interface inheritance, added a guard to `TryBuildEnsureCapacity` because `Dictionary` does have `EnsureCapacity` for older versions.
- Added 3 tests

### DIFF
--- a/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/Enumerables/EnsureCapacity/EnsureCapacityBuilder.cs
@@ -42,6 +42,10 @@ public static class EnsureCapacityBuilder
                 x => x.ReturnType.SpecialType == SpecialType.System_Boolean && x.IsStatic && x.Parameters.Length == 2 && x.IsGenericMethod
             );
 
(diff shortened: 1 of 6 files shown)


latonz (approved):
@TimothyMakkison thank you for this contribution, looks good!
https://github.com/riok/mapperly/pull/343 got merged in the meantime. Rebasing should work  without too many conflicts if you first run `dotnet csharpier .`. We introduced csharpier to unify source code formatting and eliminate discussions around the source code format.

latonz (approved):

PR #364, opened 2023-04-20
fix: enable static user methods for instance mappers

### LINKED ISSUE #200: Support static user implemented mapping methods in non-static mappers
Rel. https://github.com/riok/mapperly/discussions/199 3. and https://github.com/riok/mapperly/issues/193.

A non static mapper does not support static user implemented mapping methods.
```csharp
[Mapper]
public class CarMapper
{
    public partial CarDto ToDto(Car car);

    private static CarDtoId ToDtoId(CarId carId)
        => CarDtoId.FromId(carId);
}
```

Mapperly does not take the `ToDtoId` into account, although there is no obvious reason for not calling the user implemented method to map from `CarId` to `CarDtoId`.

TimothyMakkison:
Resolve #200
- Edit `BuildUserImplementedMapping` to allow static instance members.
- Add unit test to `UserMethodTest` and renamed some other tests, deleting the unused snapshots

Is it possible to get the diagnostic `PartialInstanceMethodInStaticMapper`? It looks like [line 83](https://github.com/TimothyMakkison/mapperly/blob/a819909cab9bd4089fd64386aa3e453be70951bb/src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs#L83) should be impossible to hit in normal C# code because you can't have an instance method in a static class.


src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
```diff
@@ -67,7 +67,7 @@ bool isStatic
             BuildParameters(ctx, method, out var parameters)
             && !method.ReturnsVoid
             && (allowPartial || !method.IsPartialDefinition)
-            && isStatic == method.IsStatic;
+            && !(isStatic && !method.IsStatic);
```
latonz, line 70: IMO a `||` condition would be easier to read here: `(!isStatic || method.IsStatic)`

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
```diff
@@ -44,14 +44,37 @@ public Task InstanceMapperShouldEmitDiagnosticForStaticMethods()
...
+        var source =
+            @"
+using System;
+using System.Collections.Generic;
+using Riok.Mapperly.Abstractions;
+
+[Mapper]
+public partial class MyMapper
+{
+    public static int StaticMapper(int s);
```
latonz, line 67: I'd try to keep the test cases valid c# when easily possible. Probably add a method expression here (eg. `=> s + 10`).

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
```diff
@@ -44,14 +44,37 @@ public Task InstanceMapperShouldEmitDiagnosticForStaticMethods()
...
+
+    public partial B AToB(A s);
```
latonz, line 69: In most test cases this method is simply called `Map`

latonz (changes requested):
Thank you for all those great contributions 😊 

I can't think of a scenario which triggers `PartialInstanceMethodInStaticMapper / RMG019`. There is also no test for it, no idea why I added it in the first place 🤷  Feel free to remove it.

TimothyMakkison:
Ok,, I've deleted the `DiagnosticDescriptor` and updated the shipped file.

latonz (approved):

latonz (approved):

PR #366, opened 2023-04-20
feat: support derived type mappings

### LINKED ISSUE #330: Automatically handle `KnownTypeAttribute`
[KnownType] attribute is used on a base type to inform callers/serializers etc what its known derived types are, Mapperly could use this to automatically generate bindings for known types

latonz:
Closes #330

* implements support to map derived types


TimothyMakkison:
This is a really cool addition, I love how elegant the generated code is.

I wasn't sure where to put this but after reading the original issue I got thinking about adding generic attributes without breaking anything. Because generic attributes are a C#11 feature it mighr be possible to support them for compatible programs by doing as follows:

- Set the language version of `Riok.Mapperly.Abstractions` to 11
- Create a copy of `MapDerivedTypeAttribute` with 2 generic parameters and an additional empty constructor.
- Inside `DerivedTypeMappingBuilder` concat configs to a call to below.
- Programs using C# 10 or lower can use the `typeof` attribute constructor, getting an error if they try to use the generic attribute. C#11 programs can use either constructors.


```C#
// Check is MapDerivedTypeAttribute...

var attrDatas = symbol.GetAttributes().Where(x => x.AttributeClass.TypeArguments.Count() == 2);

foreach (var attrData in attrDatas)
{
    var namedSymbol = attrData.AttributeClass;
    yield return new MapDerivedType(namedSymbol.TypeArguments[0], namedSymbol.TypeArguments[1]);
}
```

I've hacked together a [working version here,](https://github.com/TimothyMakkison/mapperly/tree/derived_type) it appears to work with ,NET Framework and Core. I've probably overlooked something, perhaps it only works on modern .NET Sdk's or recent versions of Visual Studio. Hopefully this helps 🤞

docs/docs/02-configuration/10-derived-type-mapping.md
```diff
@@ -0,0 +1,32 @@
...
+abstract class Car {}
+class Audi : Car {}
+class Porsche : Car {}
+
+abstract class CarDto {}
+class AudiDto : CarDto {}
+class PorscheDto : CarDto {}
+```
+
+Mapperly implements a type switch with an arm for each source type.
```
CommonGuy, line 27: Is "with an arm" correct? Even if it is, I don't think that is a statement that a lot of developers will understand.

Maybe showing the generated code is easier in this case? Does not have to be the complete, accurate generated code, just a sample

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
```diff
@@ -79,6 +79,9 @@ RMG030  | Mapper   | Error    | Reference loop detected while mapping to an init
 RMG031  | Mapper   | Warning  | Reference loop detected while mapping to a constructor member
 RMG032  | Mapper   | Warning  | The enum mapping strategy ByName cannot be used in projection mappings
 RMG033  | Mapper   | Info     | Object mapped to another object without deep clone
+RMG034  | Mapper   | Error    | Derived source type is duplicated
```
CommonGuy, line 82: "Duplicated" could lead to confusion. Maybe
> Derived source type is specified multiple times. You may only specify it once.

docs/docs/02-configuration/10-derived-type-mapping.md
latonz: Fixed.

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
latonz: Adjusted.

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -293,4 +293,31 @@ internal static class DiagnosticDescriptors
         DiagnosticSeverity.Info,
         true
     );
+
+    public static readonly DiagnosticDescriptor DerivedSourceTypeDuplicated = new DiagnosticDescriptor(
+        "RMG034",
+        "Derived source type is specified multiple times, a source type may only be specified once",
+        "Derived source type {0} is duplicated",
```
CommonGuy, line 300: I think this message should also be changed

latonz: Fixed.

CommonGuy (approved):

PR #392, opened 2023-04-30
feat: add IgnoreObsoleteMembers and `MapProperty` maps ignored Target

### LINKED ISSUE #284: Add configuration option to exclude deprecated members
Currently deprecated members are included unless ignored by a `MapperIgnore*Attribute`.
It would be nice to have an attribute `[MapperIgnoreObsoleteMembers]` and a configuration property `IgnoreObsoleteMembers` on the `MapperAttribute` to ignore all obsolete properties.

TimothyMakkison:
Resolves #284 and #450.

- Created `MapperIgnoreObsoleteMembersAttribute` and added `IgnoreObsoleteStrategy` to `MapperAttribute`. Updated `MembersMappingBuilderContext` to add the derecated members to there respective `Ignored*MemberNames`.
- Updated `FieldMember` and `PropertyMember` to expose the underlying symbol.
- Added unit tests
- Mapper attribute config is overriden by the method configuration.
- By default `IgnoreObsoleteStrategy` is none so deprecated members are treated as normal.
- The method attribute has a default parameters of `Both`  
```C#
[MapperIgnoreObsoleteMembers] partial B Map(A source);`
```

>  blanket ignore all obsolete members except when [MapProperty] is used?

Haven't done this yet, I'm not 100% sure how `MemberConfigsByRootTargetName` works to match the obsolete members to the respective explicitly mapped member.

I'll wait for #390 and then update the relevant logic.

### DIFF
--- a/src/Riok.Mapperly.Abstractions/MapperIgnoreObsoleteMembersAttribute.cs
+++ b/src/Riok.Mapperly.Abstractions/MapperIgnoreObsoleteMembersAttribute.cs
@@ -1,0 +1,22 @@
+﻿namespace Riok.Mapperly.Abstractions;
+
+/// <summary>
+/// Specifies options for obsolete ignoring strategy.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method)]
+public sealed class MapperIgnoreObsoleteMembersAttribute : Attribute
+{
+    /// <summary>
+    /// Specifies options for obsolete ignoring strategy.
+    /// </summary>
+    /// <param name="ignoreObsoleteStrategy">The strategy to be used to map <see cref="ObsoleteAttribute"/> marked members. Defaults to <see cref="IgnoreObsoleteMembersStrategy.Both"/>.</param>
+    public MapperIgnoreObsoleteMembersAttribute(IgnoreObsoleteMembersStrategy ignoreObsoleteStrategy = IgnoreObsoleteMembersStrategy.Both)
+    {
+        IgnoreObsoleteStrategy = ignoreObsoleteStrategy;
+    }
+
+    /// <summary>
+    /// The strategy used to map <see cref="ObsoleteAttribute"/> marked members.
+    /// </summary>
+    public IgnoreObsoleteMembersStrategy IgnoreObsoleteStrategy { get; }
+}
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
@@ -1,3 +1,4 @@
+using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Configuration;
 using Riok.Mapperly.Descriptors.Mappings;
 using Riok.Mapperly.Diagnostics;
@@ -21,11 +22,17 @@ public abstract class MembersMappingBuilderContext<T> : IMembersBuilderContext<T
     {
         BuilderContext = builderContext;
         Mapping = mapping;
+        MemberConfigsByRootTargetName = GetMemberConfigurations();
 
         _unmappedSourceMemberNames = GetSourceMemberNames();
         TargetMembers = GetTargetMembers();
 
-        IgnoredSourceMemberNames = builderContext.Configuration.Properties.IgnoredSources;
+        IgnoredSourceMemberNames = builderContext.Configuration.Properties.IgnoredSources
+            .Concat(GetIgnoredObsoleteSourceMembers())
+            .ToHashSet();
+        var ignoredTargetMemberNames = builderContext.Configuration.Properties.IgnoredTargets
+            .Concat(GetIgnoredObsoleteTargetMembers())
+            .ToHashSet();
 
         _ignoredUnmatchedSourceMemberNames = InitIgnoredUnmatchedProperties(IgnoredSourceMemberNames, _unmappedSourceMemberNames);
         _ignoredUnmatchedTargetMemberNames = InitIgnoredUnmatchedProperties(
@@ -34,9 +41,13 @@ public abstract class MembersMappingBuilderContext<T> : IMembersBuilderContext<T
         );
 
         _unmappedSourceMemberNames.ExceptWith(IgnoredSourceMemberNames);
-        TargetMembers.RemoveRange(builderContext.Configuration.Properties.IgnoredTargets);
 
         MemberConfigsByRootTargetName = GetMemberConfigurations();
+
+        // remove explicitly mapped ignored targets from ignoredTargetMemberNames
+        // then remove all ignored targets from TargetMembers, leaving unignored and explicitly mapped ignored members
+        ignoredTargetMemberNames.ExceptWith(MemberConfigsByRootTargetName.Keys);
+        TargetMembers.RemoveRange(ignoredTargetMemberNames);
     }
 
     public MappingBuilderContext BuilderContext { get; }
@@ -64,6 +75,32 @@ public abstract class MembersMappingBuilderContext<T> : IMembersBuilderContext<T
         var unmatched = new HashSet<string>(allProperties);
         unmatched.ExceptWith(mappedProperties);
         return unmatched;
+    }
+
+    private IEnumerable<string> GetIgnoredObsoleteTargetMembers()
+    {
+        var obsoleteStrategy = BuilderContext.Configuration.Properties.IgnoreObsoleteMembersStrategy;
+
+        if (!obsoleteStrategy.HasFlag(IgnoreObsoleteMembersStrategy.Target))
+            return Enumerable.Empty<string>();
+
+        return BuilderContext.SymbolAccessor
+            .GetAllAccessibleMappableMembers(Mapping.TargetType)
+            .Where(x => BuilderContext.SymbolAccessor.HasAttribute<ObsoleteAttribute>(x.MemberSymbol))
+            .Select(x => x.Name);
+    }
+
+    private IEnumerable<string> GetIgnoredObsoleteSourceMembers()
+    {
+        var obsoleteStrategy = BuilderContext.Configuration.Properties.IgnoreObsoleteMembersStrategy;
+
+        if (!obsoleteStrategy.HasFlag(IgnoreObsoleteMembersStrategy.Source))
+            return Enumerable.Empty<string>();
+
+        return BuilderContext.SymbolAccessor
+            .GetAllAccessibleMappableMembers(Mapping.SourceType)
+            .Where(x => BuilderContext.SymbolAccessor.HasAttribute<ObsoleteAttribute>(x.MemberSymbol))
+            .Select(x => x.Name);
     }
 
     private HashSet<string> GetSourceMemberNames()
--- a/src/Riok.Mapperly/Symbols/IMappableMember.cs
+++ b/src/Riok.Mapperly/Symbols/IMappableMember.cs
@@ -12,6 +12,8 @@ public interface IMappableMember
 
     ITypeSymbol Type { get; }
 
+    ISymbol MemberSymbol { get; }
+
     bool IsNullable { get; }
 
     bool IsIndexer { get; }
--- a/test/Riok.Mapperly.Tests/Mapping/IgnoreObsoleteTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/IgnoreObsoleteTest.cs
@@ -1,0 +1,379 @@
+﻿using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+[UsesVerify]
+public class IgnoreObsoleteTest
+{
+    private const string _classA = """
+        class A
+        {
+            public int Value { get; set; }
+
+            [Obsolete]
+            public int Ignored { get; set; }
+        }
+        """;
+
+    private const string _classB = """
+        class B
+        {
+            public int Value { get; set; }
+
+            [Obsolete]
+            public int Ignored { get; set; }
+        }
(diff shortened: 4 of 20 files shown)


TimothyMakkison:
Does Mapperly support ignoring nested properties?

```C#
// Where ProducerDto has getter setter methods and a parameterless constructor.
[MapperIgnoreSource("Manufacturer.Id")]
[MapperIgnoreTarget("Producer.Id")]
```
### Generates
```C#
 public static partial global::Riok.Mapperly.Sample.CarDto MapCarToDto(global::Riok.Mapperly.Sample.Car car)
{
    var target = new global::Riok.Mapperly.Sample.CarDto();
    if (car.Manufacturer != null)
    {
        target.Producer = MapToProducerDto(car.Manufacturer);
    }

    target.Name = car.Name;
//...
private static global::Riok.Mapperly.Sample.ProducerDto MapToProducerDto(global::Riok.Mapperly.Sample.Manufacturer source)
{
    var target = new global::Riok.Mapperly.Sample.ProducerDto();
    target.Id = source.Id;
    target.Name = source.Name;
    return target;
}
```

latonz:
@TimothyMakkison nested ignored properties are currently not supported. Is there a use-case for this?

test/Riok.Mapperly.Tests/Mapping/IgnoreObsoleteTest.cs
```diff
@@ -0,0 +1,227 @@
+﻿using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class IgnoreObsoleteTest
+{
+    private readonly string _classA = """
```
latonz, line 8: Couldn't this be a const?

test/Riok.Mapperly.Tests/Mapping/IgnoreObsoleteTest.cs
```diff
@@ -0,0 +1,227 @@
...
+                target.Ignored = source.Ignored;
+                return target;
+                """
+            );
+    }
+
+    [Fact]
+    public void MethodAttributeIgnoreObsoleteBoth()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes("[MapperIgnoreObsoleteMembers()] partial B Map(A source);", _classA, _classB);
```
latonz, line 148: Nit: rm the unneeded parenthesis

test/Riok.Mapperly.Tests/Mapping/IgnoreObsoleteTest.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            public int Value { get; set;}
+
+            [Obsolete]
+            public int Ignored { get; set;}
+        }
+        """;
+
+    [Fact]
+    public void ClassAttributeIgnoreObsoleteNone()
```
latonz, line 29: Add a test case which explicitly maps a obsolete property via `[MapProperty]` which should overwrite the obsolete member strategy.

TimothyMakkison:
> @TimothyMakkison nested ignored properties are currently not supported. Is there a use-case for this?

Mb I thought it did, wasn't sure if i had messed up the member paths. I wanted to reuse some of the logic to let explicit mappings override ignored deprecated members.

I think nested ignores would be a good addition to mapperl, It already has nested explicit mappings.

latonz (changes requested):
Thank you for another great contribution 🌞 
* Good looking tests 😊
* Could you also update the documentation?
* TBD: Should obsolete enum members also be ignored (to be discussed, if implemented, implement it in another PR). Should it be automatically or via a separate configuration? I think the best way would be to do it the exact same way, without any additional configuration. If it should be overwritten for an enum, it could get overwritten on an enum mapping method with the same attribute. WDYT?

latonz:
@TimothyMakkison 
> I think nested ignores would be a good addition to mapperl, It already has nested explicit mappings.

Can you show a concrete use-case for this? Probably open a new discussion about this topic to not bloat this PR.

src/Riok.Mapperly.Abstractions/MapperIgnoreObsoleteMembersAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Riok.Mapperly.Abstractions;
+
+/// <summary>
+/// Specifies options for obsolete ignoring strategy.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
```
latonz, line 6: `AllowMultiple` can be removed as it is the default anyway.

latonz:
@TimothyMakkison FYI: an issue to the nested ignored properties just got opened: https://github.com/riok/mapperly/issues/395

latonz:
@TimothyMakkison feel free to re-request a review when ready 😊

TimothyMakkison:
> ignore all obsolete members except when [MapProperty] is used?


Discovered #450 when trying to implement this. I could fix the bug so `MapProperty` will map ignored target properties just like obsolete target properties.

src/Riok.Mapperly/Symbols/IMappableMember.cs
```diff
@@ -9,6 +9,8 @@ public interface IMappableMember
 
     ITypeSymbol Type { get; }
 
+    ISymbol MemberSymbol { get; }
```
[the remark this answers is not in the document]
TimothyMakkison: I originally didn't do this because:
- This property is only used when `IgnoreObsolete` is enabled and could slow down generation.
- Requires a `WellKnownTypes` instance being passed to `MappableMember.Create` / `new PropertyMember`.

WDYT?

TimothyMakkison:
- `MapProperty` now always maps ignored obsolete members.
  - In the process resolved #450 decided to make it so `MapProperty` overrides ingored target properties just like it already did for ignored source proeprties.
- Updated `PublicAPI.Shipped` file.
- Added `MapProperty` tests.
- I haven't added to the docs, should I create a brief summary, and if so where?

src/Riok.Mapperly/Helpers/DictionaryExtensions.cs
```diff
@@ -18,4 +18,19 @@ public static class DictionaryExtensions
             dict.Remove(key);
         }
     }
+
+    public static Dictionary<TKey, TValue> ExtractRange<TKey, TValue>(this IDictionary<TKey, TValue> dict, IEnumerable<TKey> keys)
```
latonz, line 22: Could you unit test this?

src/Riok.Mapperly/Symbols/IMappableMember.cs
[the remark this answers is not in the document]
latonz: I think your points make sense, exposing it is okay with me.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
```diff
@@ -66,6 +75,40 @@ private HashSet<string> InitIgnoredUnmatchedProperties(IEnumerable<string> allPr
...
+        {
+            var obsoleteAttribute = BuilderContext.Types.Get<ObsoleteAttribute>();
+
+            return Mapping.TargetType
+                .GetAccessibleMappableMembers()
+                .Where(x => x.MemberSymbol.HasAttribute(obsoleteAttribute))
+                .Select(x => x.Name);
+        }
+
+        return Enumerable.Empty<string>();
```
latonz, line 92: I'd invert the if condition, same for `GetIgnoredObsoleteSourceMembers` (IMO improves the readability).

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
```diff
@@ -66,6 +66,19 @@ private static void BuildInitOnlyMemberMappings(INewInstanceBuilderContext<IMapp
...
+            : ctx.IgnoredTargetMembers.Values.Where(x => x.CanOnlySetViaInitializer()).ToArray();
+        foreach (var targetMember in initOnlyIgnoredTargetMembers)
+        {
+            ctx.IgnoredTargetMembers.Remove(targetMember.Name);
+
+            if (ctx.MemberConfigsByRootTargetName.Remove(targetMember.Name, out var memberConfigs))
+            {
+                BuildInitMemberMapping(ctx, targetMember, memberConfigs);
+            }
+        }
```
latonz, line 81: Could you add a comment what this does and why it is needed? Does this work nicely with required init properties?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/ObjectMemberMappingBodyBuilder.cs
```diff
@@ -66,6 +66,20 @@ public static void BuildMappingBody(IMembersContainerBuilderContext<IMemberAssig
             }
         }
 
+        foreach (var targetMember in ctx.IgnoredTargetMembers.Values)
+        {
+            if (ctx.MemberConfigsByRootTargetName.Remove(targetMember.Name, out var memberConfigs))
+            {
+                // add all configured mappings
```
latonz, line 73: probably add ` even if they are ignored` and move it up to L69 as this is where the section begins IMO.

latonz (changes requested):
Thank you for the updates, I added my feedback.
Could you also update the documentation (I would add a new section in `01-mapper.md` in `## Properties / fields`).

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
TimothyMakkison: Added comment and added required/init tests

TimothyMakkison:
Note that in order to support `MapProperty` mapping obsolete targets, I fixed #450 in this PR. 

Now if a target is Ignored but is explicitly mapped with `MapProperty` a mapping will be created.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
```diff
@@ -66,6 +66,19 @@ private static void BuildInitOnlyMemberMappings(INewInstanceBuilderContext<IMapp
...
+        var initOnlyIgnoredTargetMembers = includeAllMembers
+            ? ctx.IgnoredTargetMembers.Values.ToArray()
+            : ctx.IgnoredTargetMembers.Values.Where(x => x.CanOnlySetViaInitializer()).ToArray();
+        foreach (var targetMember in initOnlyIgnoredTargetMembers)
+        {
+            if (ctx.MemberConfigsByRootTargetName.Remove(targetMember.Name, out var memberConfigs))
+            {
+                BuildInitMemberMapping(ctx, targetMember, memberConfigs);
+            }
+        }
```
latonz, line 81: Instead of this, wouldn't it be much simpler to just add `ignoredTargetMemberNames.ExceptWith(MemberConfigsByRootTargetName.Keys);` in `MembersMappingBuilderContext` before initialising `IgnoredTargetMembers` (L45)?

latonz (changes requested):
I'd add a warning diagnostic if an explicit ignored property is mapped manually (but not if a property is ignored because it is obsolete). I think this could be done in another PR.

Could you adjust the PR body that it resolves https://github.com/riok/mapperly/issues/450?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
TimothyMakkison: It would save time by removing ignored targets that are not explicitly mapped. I don't think it would replace this logic tho

latonz (approved):

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
```diff
@@ -66,6 +76,37 @@ private HashSet<string> InitIgnoredUnmatchedProperties(IEnumerable<string> allPr
...
+            .GetAccessibleMappableMembers()
+            .Where(x => x.MemberSymbol.HasAttribute(obsoleteAttribute))
+            .Select(x => x.Name);
+    }
+
+    private IEnumerable<string> GetIgnoredObsoleteSourceMembers()
+    {
+        var obsoleteStrategy = BuilderContext.Configuration.Properties.IgnoreObsoleteMembersStrategy;
+
+        if (obsoleteStrategy.HasFlag(IgnoreObsoleteMembersStrategy.Source))
```
latonz, line 97: I'd invert this condition to improve the readability. Similar to the style which is implemented by `GetIgnoredObsoleteTargetMembers`.

latonz (changes requested):
@TimothyMakkison I created https://github.com/riok/mapperly/issues/512 to implement the diagnostic.
Whats still missing:
* Update the documentation (I would add a new section in `01-mapper.md` in `## Properties / fields`)
* Update integration tests

Otherwise this MR looks good 😊

latonz (dismissed):

latonz:
@TimothyMakkison looks good to me. A few linter issues need to be fixed...

TimothyMakkison:
> @TimothyMakkison looks good to me. A few linter issues need to be fixed...

Thanks, please don't merge this just yet, I haven't added the integration tests

latonz:
Oh thanks, I would have missed that 🙈

latonz (approved):

PR #405, opened 2023-05-03
feat: Multiple ignore properties with one attribute

tborozan:
Support for ignoring multiple source and target properties with a single attribute has been added. The new `MapperIgnoreSourcesAttribute` and `MapperIgnoreTargetsAttribute` allow multiple properties to be specified for ignoring in a single attribute. This makes mapper classes cleaner when there are many repeated ignored properties, such as when many objects inherit from a class with properties that need to be ignored.”


latonz:
Thank you for your PR, looks pretty good. However, we want to keep the API surface of Mapperly small and we think that the benefit with this new attribute vs. multiple ignore attributes with a single property each isn't that big that it is worth introducing additional attributes.

PR #411, opened 2023-05-07
fix: lambdas should use correct scopes to generate names

latonz:
The name builder created for lambda parameters wasn't used in the context created to build mappings inside those lambdas. This could lead to name conflicts.


latonz:
@TimothyMakkison as you contributed several cool features and fixes in the recent past and got to know the code base pretty good, if you feel like it, an additional review from you would be appreciated 😊.

ni507 (approved):

PR #414, opened 2023-05-08
fix: throw NullReferenceException for element access

### LINKED ISSUE #413: Mapping string[] to string[] throws CS8081
**Describe the bug**
Mapping string array to a string array throws a CS8081: "Expression does not have a name".

The picture describes the error:
afbeelding

**To Reproduce**
1. Use the code snippets and check the generated code

**Expected behavior**
No CS8081 exception in the generated code.
Also the generated code is different with no generated code header comment which is strange behavior?

**Code snippets**
A.cs:
``` cs
namespace Namespace;

using System;

public class A
{
	public string[] Items { get; set; }

	public A()
	{
		Items = Array.Empty<string>();
	}

	public A(string[] array)
	{
		Items = array;
	}
}
```
B.cs:
``` cs
// //------------------------------------------------------------------------------
// // <auto-generated>
// //     This code was generated by a tool.
// //
// //     Changes to this file may cause incorrect behavior and will be lost if
// //     the code is regenerated.
// // </auto-generated>
// //------------------------------------------------------------------------------
namespace AnotherNamespace
{
	using System;

	public class A
	{
		public string[] Items { get; set; }

		public A()
		{
			Items = Array.Empty<string>();
		}

		public A(string[] array)
		{
			Items = array;
		}
	}
}
```
Mapper.cs:
``` cs
namespace Namespace;

[Mapper]
public partial class Mapper
{
	public partial A AtoA(AnotherNamespace.A abcde);
}
```

**Environment:**
- Mapperly Version: 2.8.0
- .NET Version: .NET SDK 7.0.203
- Target Framework: net6.0
- IDE: Visual Studio v17.5.5
- OS: Windows 11 22H2

TimothyMakkison:
### Description

Added a `NullReferenceException` with the message `"Sequence {nameof(source)}, contained a null value"` arm to `NullSubstitute` for element accessors. 
The error message could be improved, I considered `"Null value found inside sequence {nameof(source)}"`.

Fixes #413.

### DIFF
--- a/src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
+++ b/src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
@@ -77,6 +77,12 @@ public static class SyntaxFactoryHelper
             NullFallbackValue.Default => DefaultLiteral(),
             NullFallbackValue.EmptyString => StringLiteral(string.Empty),
             NullFallbackValue.CreateInstance => CreateInstance(t),
+            _ when argument is ElementAccessExpressionSyntax memberAccess
+                => ThrowNullReferenceException(
+                    InterpolatedString(
+                        $"Sequence {NameOf(memberAccess.Expression)}, contained a null value at index {memberAccess.ArgumentList.Arguments[0].Expression}."
+                    )
+                ),
             _ => ThrowArgumentNullException(argument),
         };
     }
@@ -162,6 +168,13 @@ public static class SyntaxFactoryHelper
     {
         return ThrowExpression(
             ObjectCreationExpression(IdentifierName(NullReferenceExceptionClassName)).WithArgumentList(ArgumentList(StringLiteral(message)))
+        );
+    }
+
+    public static ThrowExpressionSyntax ThrowNullReferenceException(ExpressionSyntax arg)
+    {
+        return ThrowExpression(
+            ObjectCreationExpression(IdentifierName(NullReferenceExceptionClassName)).WithArgumentList(ArgumentList(arg))
         );
     }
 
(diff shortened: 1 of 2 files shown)


src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
```diff
@@ -77,6 +77,8 @@ public static ExpressionSyntax NullSubstitute(ITypeSymbol t, ExpressionSyntax ar
             NullFallbackValue.Default => DefaultLiteral(),
             NullFallbackValue.EmptyString => StringLiteral(string.Empty),
             NullFallbackValue.CreateInstance => CreateInstance(t),
+            _ when argument is ElementAccessExpressionSyntax memberAccess
+                => ThrowNullReferenceException(InterpolatedString($"Sequence {NameOf(memberAccess.Expression)}, contained a null value.")),
```
latonz, line 81: I think including the index in the exception message could help a lot when troubleshooting such issues (eg. `Sequence {nameof(source)} contained a null value at index {index}`.

src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
```diff
@@ -158,13 +160,39 @@ public static InterpolatedStringExpressionSyntax InterpolatedString(FormattableS
 
     public static InvocationExpressionSyntax NameOf(ExpressionSyntax expression) => Invocation(_nameofIdentifier, expression);
 
+    public static ExpressionSyntax ConstantNameOf(ExpressionSyntax expression)
```
latonz, line 163: I don't think nameof of an element access expression ever makes sense, therefore I don't think this method should be needed at all. In what case does this make sense?

latonz (changes requested):

src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
TimothyMakkison: mb accidentally added it when I made a backup. I'll rmeove it

latonz (approved):

latonz (approved):

latonz (approved):

ni507 (approved):

PR #419, opened 2023-05-10
feat: added source generator benchmark

TimothyMakkison:
### Description

Adapted code from [Mediator](https://github.com/martinothamar/Mediator/tree/main/benchmarks/Mediator.Benchmarks/SourceGenerator) to benchmark Mapperly source generator performance. This benchmark uses the sample project although a different more complex file could be used instead.

### DIFF
--- a/benchmarks/Riok.Mapperly.Benchmarks/MappingBenchmarks.cs
+++ b/benchmarks/Riok.Mapperly.Benchmarks/MappingBenchmarks.cs
@@ -1,0 +1,37 @@
+﻿using BenchmarkDotNet.Attributes;
+using Riok.Mapperly.IntegrationTests;
+using Riok.Mapperly.IntegrationTests.Dto;
+using Riok.Mapperly.IntegrationTests.Mapper;
+using Riok.Mapperly.IntegrationTests.Models;
+
+namespace Riok.Mapperly.Benchmarks;
+
+[MemoryDiagnoser]
+[InProcess]
+[HtmlExporter]
+[JsonExporterAttribute.Full]
+public class MappingBenchmarks
+{
+    private readonly TestObject _testObject;
+    private readonly TestObjectDto _testObjectDto;
+    private readonly IdObject _idObject;
+
+    public MappingBenchmarks()
+    {
+        _testObject = BaseMapperTest.NewTestObj();
+        _testObjectDto = StaticTestMapper.MapToDto(_testObject);
+        _idObject = new IdObject { IdValue = 143 };
+    }
+
+    [Benchmark(Description = "MapComplexToDto")]
+    public TestObjectDto MapComplexToDto() => StaticTestMapper.MapToDto(_testObject);
+
+    [Benchmark(Description = "MapComplexToExistingDto")]
+    public void MapComplexToExistingDto() => StaticTestMapper.UpdateDto(_testObject, _testObjectDto);
+
+    [Benchmark(Description = "MapSimpleDeepClone")]
+    public IdObject MapSimpleDeepClone() => DeepCloningMapper.Copy(_idObject);
+
+    [Benchmark(Description = "MapComplexDeepClone")]
+    public TestObject MapComplexDeepClone() => DeepCloningMapper.Copy(_testObject);
+}
--- a/benchmarks/Riok.Mapperly.Benchmarks/Riok.Mapperly.Benchmarks.csproj
+++ b/benchmarks/Riok.Mapperly.Benchmarks/Riok.Mapperly.Benchmarks.csproj
@@ -1,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <DebugSymbols>true</DebugSymbols>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.13.5" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.6.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.6.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.6.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.6.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.6.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.Build.Locator" Version="1.5.5" />
+    <PackageReference Include="NuGet.Frameworks" Version="6.6.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Riok.Mapperly\Riok.Mapperly.csproj" />
+    <ProjectReference Include="..\..\src\Riok.Mapperly\Riok.Mapperly.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\..\src\Riok.Mapperly.Abstractions\Riok.Mapperly.Abstractions.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="true" />
+    <ProjectReference Include="..\..\test\Riok.Mapperly.IntegrationTests\Riok.Mapperly.IntegrationTests.csproj" />
+  </ItemGroup>
+</Project>
--- a/Riok.Mapperly.sln
+++ b/Riok.Mapperly.sln
@@ -1,33 +1,34 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30114.105
+# Visual Studio Version 17
+VisualStudioVersion = 17.5.33424.131
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{B65AF89A-4A3B-473C-83C8-5F0CB0EED30E}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Riok.Mapperly", "src\Riok.Mapperly\Riok.Mapperly.csproj", "{CB991FD7-B9B7-47C0-A060-66EBE136DBDA}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Riok.Mapperly", "src\Riok.Mapperly\Riok.Mapperly.csproj", "{CB991FD7-B9B7-47C0-A060-66EBE136DBDA}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{3598BE50-28D5-4BF4-BEA7-09E5FEA2910C}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Riok.Mapperly.IntegrationTests", "test\Riok.Mapperly.IntegrationTests\Riok.Mapperly.IntegrationTests.csproj", "{FDA97A46-DB21-4B72-9958-6D61C508B1CD}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Riok.Mapperly.IntegrationTests", "test\Riok.Mapperly.IntegrationTests\Riok.Mapperly.IntegrationTests.csproj", "{FDA97A46-DB21-4B72-9958-6D61C508B1CD}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Riok.Mapperly.Abstractions", "src\Riok.Mapperly.Abstractions\Riok.Mapperly.Abstractions.csproj", "{E45D5E6D-8CC9-4DAD-8E1C-723625475744}"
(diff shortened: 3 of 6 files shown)


benchmarks/Riok.Mapperly.Benchmarks/SourceGeneratorBenchmark.cs
```diff
@@ -0,0 +1,80 @@
...
+        _workspace.WorkspaceFailed += (sender, args) =>
+        {
+            ConsoleLogger.Default.WriteLineError("-------------------------");
+            ConsoleLogger.Default.WriteLineError(args.Diagnostic.ToString());
+            ConsoleLogger.Default.WriteLineError("-------------------------");
+        };
+
+        var projectFile = Path.Combine(currentDirectory, "../../samples/Riok.Mapperly.Sample/Riok.Mapperly.Sample.csproj");
+        if (!File.Exists(projectFile))
+            throw new Exception("Project doesnt exist");
```
odaiwa, line 41:
```suggestion
            throw new Exception("Project doesn't exist");
```

benchmarks/Riok.Mapperly.Benchmarks/SourceGeneratorBenchmark.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.MSBuild;
+
+namespace Riok.Mapperly.Benchmarks;
+
+[MemoryDiagnoser]
+[InProcess]
+//[EventPipeProfiler(EventPipeProfile.CpuSampling)]
+//[EtwProfiler]
+//[DisassemblyDiagnoser]
```
latonz, line 14: Shouldn't this commented code be removed before going to main? Or add comments explaining why it is commented and what activating these attributes does.

benchmarks/Riok.Mapperly.Benchmarks/SourceGeneratorBenchmark.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        _workspace = MSBuildWorkspace.Create();
+        _workspace.WorkspaceFailed += (sender, args) =>
+        {
+            ConsoleLogger.Default.WriteLineError("-------------------------");
+            ConsoleLogger.Default.WriteLineError(args.Diagnostic.ToString());
+            ConsoleLogger.Default.WriteLineError("-------------------------");
+        };
+
+        var projectFile = Path.Combine(currentDirectory, "../../samples/Riok.Mapperly.Sample/Riok.Mapperly.Sample.csproj");
```
latonz, line 39: Extract this magic string into a named constant.

latonz (changes requested):
Thank you for this contribution.
* adding a `launchSettings.json` would be helpful as several IDE's load the launchsettings as run configurations into the IDE
* I think a benchmark of the generated mapping at runtime would be helpful too, wdyt? (probably benchmark `Riok.Mapperly.IntegrationTests.Mapper.StaticTestMapper.MapToDto` as it contains a bit of everything?)
* To be discussed: would it make sense to continuously monitor the performance? Eg. via https://github.com/benchmark-action/github-action-benchmark? (I never used that before, just found it by doing a quick google search)

benchmarks/Riok.Mapperly.Benchmarks/SourceGeneratorBenchmark.cs
```diff
@@ -0,0 +1,80 @@
...
+        if (!File.Exists(projectFile))
+            throw new Exception("Project doesn't exist");
+        else
+            ConsoleLogger.Default.WriteLine("Project exists!");
+
+        Project? project = null;
+        try
+        {
+            ConsoleLogger.Default.WriteLine("Loading project!!!");
```
latonz, line 48: Nit: I don't think these exclamation marks at the end add any value. These messages don't seem to important to me, that they would need a high attention.

TimothyMakkison:
> * adding a `launchSettings.json` would be helpful as several IDE's load the launchsettings as run configurations into the IDE

What should be inside this file? afaik this benchmark doesn't use configuration for args.

> * To be discussed: I think a benchmark of the generated mapping at runtime would be helpful too, wdyt? (probably benchmark `Riok.Mapperly.IntegrationTests.Mapper.StaticTestMapper.MapToDto` as it contains a bit of everything?)

I thought it would be really useful although it probably won't be worth `github-action-benchmark` on this. I also thought we could add a benchmark comparing Mapperly to other .NET benchmarking libraries.

> * To be discussed: would it make sense to continuously monitor the performance? Eg. via https://github.com/benchmark-action/github-action-benchmark? (I never used that before, just found it by doing a quick google search)

Good idea, I think I've use this in a different project. I'll try to add a .yml file.

benchmarks/Riok.Mapperly.Benchmarks/MapperComparison/BenchmarkAll.cs
```diff
@@ -0,0 +1,80 @@
...
+[MemoryDiagnoser]
+[InProcess]
+[HtmlExporter]
+[JsonExporterAttribute.Full]
+public class BenchmarkAll
+{
+    private readonly Foo _fooInstance = TestHelper.SetupFooInstance();
+    private readonly Customer _customerInstance = TestHelper.SetupCustomerInstance();
+
+    [Params(100_000)] //, 1_000_000)]
```
latonz, line 15: Remove commented code.

benchmarks/Riok.Mapperly.Benchmarks/MapperComparison/Models/Customer.cs
```diff
@@ -0,0 +1,55 @@
+﻿namespace Riok.Mapperly.Benchmarks.MapperComparison.Models;
+
+public class Address
```
latonz, line 3: This code base usually creates one file per type.

benchmarks/Riok.Mapperly.Benchmarks/SourceGeneratorBenchmark.cs
```diff
@@ -0,0 +1,88 @@
...
+    private static string CallerFilePath([CallerFilePath] string callerFilePath = default!) => callerFilePath;
+
+    [GlobalSetup]
+    public void Setup() => SetupAsync().GetAwaiter().GetResult();
+
+    private async Task SetupAsync()
+    {
+        var currentDirectory = CallerFilePath();
+
+        ConsoleLogger.Default.WriteLine($"Starting at {currentDirectory}");
```
latonz, line 32: I don't think this adds any benefit.

.github/workflows/sourcegenerator-benchmark.yml
```diff
@@ -0,0 +1,50 @@
...
+          dotnet-version: "7.0.5" # SDK Version to use. keep in line with examples/benchmarkdotnet/global.json
+      - name: Run benchmark
+        run: cd benchmarks/Riok.Mapperly.Benchmarks && dotnet run --exporters json --filter '*'
+
+      - name: Store benchmark result
+        uses: rhysd/github-action-benchmark@v1
+        with:
+          name: Benchmark.Net Benchmark
+          tool: "benchmarkdotnet"
+          output-file-path: examples/benchmarkdotnet/BenchmarkDotNet.Artifacts/results/Sample.Benchmarks-report-full-compressed.json
```
latonz, line 28: Is this really correct?

.github/workflows/sourcegenerator-benchmark.yml
```diff
@@ -0,0 +1,50 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: "200%"
+          comment-on-alert: true
+          fail-on-alert: true
+          alert-comment-cc-users: "@ktrz"
```
latonz, line 35: I don't think we need to cc users, as this project still has only a handful of maintainers and each PR is reviewed anyway.

.github/workflows/sourcegenerator-benchmark.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Store benchmark result - separate results repo
```
latonz, line 37: Isn't this a duplicate of the above step?

latonz (changes requested):
> What should be inside this file? afaik this benchmark doesn't use configuration for args.

Rider shows nice ready-to run configurations if a `launchSettings.json` exists, can't tell what VS does. Therefore it helps at least for Rider users.

benchmarks/Riok.Mapperly.Benchmarks/Riok.Mapperly.Benchmarks.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.5.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.5.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.5.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.5.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" PrivateAssets="all" />
+
+    <PackageReference Include="ExpressionDebugger" Version="2.2.1" />
+    <PackageReference Include="ExpressionTranslator" Version="2.5.0" />
+    <PackageReference Include="AutoMapper" Version="10.1.1" />
+    <PackageReference Include="FastExpressionCompiler" Version="3.2.1" />
```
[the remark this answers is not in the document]
TimothyMakkison: Okay I'll remove the comparisons 👍  tbh they were really handy, I found two bugs!

TimothyMakkison:
Thanks for your help, I've updated the benchmark, removed the comparisons and modified the action file. 

tbh this is my second time writing a github action and It looks like it will someone with admin level control to setup. Shall I delete the benchmark.yml file and let you or someone with more experience set it up?

.github/workflows/sourcegenerator-benchmark.yml
```diff
@@ -0,0 +1,50 @@
+name: SourceGenerator Benchmark
+on:
+  push:
+    branches:
+      - main
```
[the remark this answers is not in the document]
TimothyMakkison: It will take a while to un eahc time but I've changed it to:
```yml
workflow_call:
pull_request:
  branches:
    - "**"
push:
  branches:
    - main
```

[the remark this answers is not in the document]
TimothyMakkison: See also https://github.com/benchmark-action/github-action-benchmark#caveats

TimothyMakkison:
> You now removed the benchmark of the generated code of Mapperly too, did you? I think it would make sense to keep that one.

Readded the benchamarks from Mapster but removed comparisons to the other mappers,

> To be discussed: I think a benchmark of the generated mapping at runtime would be helpful too, wdyt? (probably benchmark Riok.Mapperly.IntegrationTests.Mapper.StaticTestMapper.MapToDto as it contains a bit of everything?)

Added additional benchmark to `SourceGeneratorBenchmarks` where the time to generate `Riok.Mapperly.IntegrationTests` is measured. I tried adding `StaticTestMapper.MapToDto` as a benchmark but I kept getting a metadata error about not being able to access .NET 7 dll. I suspect the multi targeting logic in the integration tests causes this.

image

benchmarks/Riok.Mapperly.Benchmarks/MappingBenchmarks.cs
```diff
@@ -0,0 +1,118 @@
...
+[MemoryDiagnoser]
+[InProcess]
+[HtmlExporter]
+[JsonExporterAttribute.Full]
+public class MappingBenchmarks
+{
+    private readonly Foo _foo;
+    private readonly Customer _customer;
+
+    public MappingBenchmarks()
```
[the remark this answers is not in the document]
TimothyMakkison: mb I thought you requested that I readd the mapster comparison benchmarks but without the benchmarks for the other libraries.  

> another Mapper around which would need updates with new features, etc.?

Im not sure what you mean by this, why would the current `MappingBenchmarks` need updates?

>  You mention a dll error, do you have more information on that? Probably I'll give it a try in the upcoming days.

I'll see if I can find my stash for this 👍. It would be nice to have a large complex mapper to track.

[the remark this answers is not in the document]
latonz: > Im not sure what you mean by this, why would the current MappingBenchmarks need updates?

If a new feature is implemented (which should usually then also be in the benchmark and integration tests), `Foo`, `Customer`, `CustomerDto`, `MapperlyMapper` and `MappingBenchmarks` probably all need to be updated whereas if the integration test mappers could be reused changes would only have to be done once for a new feature.

TimothyMakkison:
I've removed the Mapster benchmarks and adapted the integration tests into `MapComplexToDto`, `MapComplexToExistingDto`, `MapSimpleDeepClone` and `MapComplexDeepClone`. 

>  You mention a dll error, do you have more information on that? Probably I'll give it a try in the upcoming days.

I wrote and ran these tests in Rider, I suspect this is a Visual Studio issue, I can't even recreate the dll error because VS won't build any of the recent branches. I keep getting the following instead:
`Generator 'MapperGenerator' failed to generate source. Exception was of type 'InvalidOperationException' with message 'Type Riok.Mapperly.Abstractions.EnumMappingStrategy, Riok.Mapperly.Abstractions not found`
Probably caused by installing the latest version of VS or .NET 🤷 

> If a new feature is implemented (which should usually then also be in the benchmark and integration tests), Foo, Customer, CustomerDto, MapperlyMapper and MappingBenchmarks probably all need to be updated whereas if the integration test mappers could be reused changes would only have to be done once for a new feature.

Gotcha 👍

TimothyMakkison:
This PR should be finished, it stopped working at one point so I added some packages to fix this I hope this isn't an issue.

benchmarks/Riok.Mapperly.Benchmarks/Riok.Mapperly.Benchmarks.csproj
```diff
@@ -0,0 +1,55 @@
...
+    <None Remove="BenchmarkDotNet.Artifacts\Riok.Mapperly.Benchmarks.SourceGeneratorBenchmarks-20230530-115845.log" />
+    <None Remove="BenchmarkDotNet.Artifacts\Riok.Mapperly.Benchmarks.SourceGeneratorBenchmarks-20230530-115624.log" />
+    <None Remove="BenchmarkDotNet.Artifacts\Riok.Mapperly.Benchmarks.SourceGeneratorBenchmarks-20230530-113524.log" />
+    <None Remove="BenchmarkDotNet.Artifacts\Riok.Mapperly.Benchmarks.SourceGeneratorBenchmarks-20230530-113300.log" />
+    <None Remove="BenchmarkDotNet.Artifacts\Riok.Mapperly.Benchmarks.SourceGeneratorBenchmarks-20230529-231934.log" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="BenchmarkDotNet.Artifacts\results\" />
+  </ItemGroup>
```
latonz, line 53: Is this really needed?

latonz (changes requested):

latonz:
@TimothyMakkison could you rebase this? I'll merge it afterwards.

latonz (approved):

Issue #421, opened 2023-05-11
Invalid code generation with target type param method


tvardero:
**Describe the bug**
Invalid generation (invalid C#) of mapper when mapper has method with provided target type. Helper method with generic type parameter should be ignored.

**To Reproduce**
Given the code: 
```csharp
[Mapper]
public partial class MapperlyMapper
{
    public partial object? Map(object? source, Type targetType);

    public partial CompanyDto MapToDto(Company company);

    public partial EmployeeDto MapToDto(Employee employee);

    // Helper method
    public TDestination? Map<TDestination>(object? source) => (TDestination?) Map(source, typeof(TDestination));
}
```

Expected (company and employee mappings are omitted for brevity):
```csharp
public partial object? Map(object? source, global::System.Type targetType)
{
    return source switch
    {
        global::MapperlySandbox.Company x when targetType.IsAssignableFrom(typeof(global::MapperlySandbox.CompanyDto)) => MapToDto(x),
        global::MapperlySandbox.Employee x when targetType.IsAssignableFrom(typeof(global::MapperlySandbox.EmployeeDto)) => MapToDto(x),
        null => default,
        _ => throw new System.ArgumentException($"Cannot map {source.GetType()} to {targetType} as there is no known type mapping", nameof(source)),
    };
}
```

Actual:
```csharp
public partial object? Map(object? source, global::System.Type targetType)
{
    return source switch
    {
        global::MapperlySandbox.Company x when targetType.IsAssignableFrom(typeof(global::MapperlySandbox.CompanyDto)) => MapToDto(x),
        global::MapperlySandbox.Employee x when targetType.IsAssignableFrom(typeof(global::MapperlySandbox.EmployeeDto)) => MapToDto(x),
        object x when targetType.IsAssignableFrom(typeof(TDestination)) => MapA(x),
        null => default,
        _ => throw new System.ArgumentException($"Cannot map {source.GetType()} to {targetType} as there is no known type mapping", nameof(source)),
    };
}
```

Line `object x when targetType.IsAssignableFrom(typeof(TDestination)) => MapA(x),` is invalid since `TDestination` is unknown symbol in the current scope.

**Environment (please complete the following information):**
- Mapperly Version: 2.9.0-next.1
- .NET Version: 7.0.203
- Target Framework: net7.0
- IDE: Rider latest
- OS: win11 x64

tvardero:
Workaround: user should not write helper methods in the mapper class. Instead, use extensions methods for mapper:
[...]

Issue #357, opened 2023-04-18
Generic mapping methods


Fube:
**Is your feature request related to a problem? Please describe.**
I have an interface with many implementations.

I want to define one (user) mapping for all of them by doing something like:
```csharp
public partial TheTarget GeneratedMapping (TheSource source);
private TheTargetType MyMapping (TheInterface source) => // some logic here
```

The issue is that on my source type (`TheSource`) the property is not typed as the interface, but as one of its implementations.
As such, the user mapping does not get picked up and one gets created instead.

**Describe the solution you'd like**
Currently, the object factory allows for generics with constraints. I think something like that for user mapping would work.

`private TheTargetType MyMapping<T> (T source) where T : TheInterface => // some logic here`

**Describe alternatives you've considered**
Currently, I name every single type that implements `TheInterface` and create a user mapping per type, but they all have the same exact logic.

**Additional context**
I have written some (failing) tests to more clearly demonstrate the problem, I hope it helps

```csharp
    [Fact]
    public void SourceHasDerivedInterfaceTest()
    {
        var source = TestSourceBuilder.CSharp("""
            using System;
            using Riok.Mapperly.Abstractions;

            public interface ITheInterface
            {
                long SomeProp { get; }
            }

            public class TheImpl : ITheInterface
            {
                public long SomeProp { get; set; }
            }

            public class TheOutsider
            {
                public long SomeProp { get; set; }
            }

            public class TheTarget
            {
                public TheOutsider Me { get; set; }
            }

            public class TheSource
            {
                public TheImpl Me { get; set; }
            }

            [Mapper]
            public partial class TheMapper
            {
                public partial TheTarget TheMappingMethod(TheSource source);
                private TheOutsider TheUserMapping(ITheInterface source) => source is null ? null : new TheOutsider();
            }
        """);

        TestHelper.GenerateMapper(source)
            .Should()
            .HaveSingleMethodBody("var target = new global::TheTarget();\ntarget.Me = TheUserMapping(source.Me);\nreturn target;");
    }


    [Fact]
    public void SourceHasDerived_ButUserMappingIsGeneric_InterfaceTest()
    {
        var source = TestSourceBuilder.CSharp("""
            using System;
            using Riok.Mapperly.Abstractions;

            public interface ITheInterface
            {
                long SomeProp { get; }
            }

            public class TheImpl : ITheInterface
            {
                public long SomeProp { get; set; }
            }

            public class TheOutsider
            {
                public long SomeProp { get; set; }
[...]

PR #423, opened 2023-05-11
fix: ignore user implemented generic methods

latonz:
Closes https://github.com/riok/mapperly/issues/421


tvardero:
There might be a case when user will want to generate generic mapping. For example:
`public partial CarDto MapToDto<TCar>(TCar car) where TCar : Car;`
This PR will disable generation for such mappings.

As a alternative for #421 what about disabling generation for overloads for _the method_? (I mean the method that accepts target type as parameter)

latonz:
@tvardero Mapperly currently does not support generic user implemented mappings. Generic generated mappings are planned in https://github.com/riok/mapperly/issues/357. Therefore it makes sense to completely ignore them for now, as Mapperly does not know how to call them and would generate invalid code. Usually this does not happen, since no type matches the generic type. But it is cleaner to ignore them completely as long as they are not supported by Mapperly.

hoyau (approved):

PR #425, opened 2023-05-13
Add Span and Memory mapping support

### LINKED ISSUE #391: Deep clone does not deep copy `Span<T>` or `ReadOnlySpan<T>` elements in ref structs
**Describe the bug**
When `UseDeepCloning` is enabled `Span<T>` and ReadOnlySpan<T>` elements are not correctly copied.

**To Reproduce**
- Create a ref struct with a `Span<T>` or `ReadOnlySpan<T>` field.
```C#
[Mapper(UseDeepCloning =true)]
{
    public static partial MyRefStruct Map(MyRefStruct source);
}
public ref struct MyRefStruct
{
    public Span<int> Span;
}
```

Will generate:
```C#
public static partial global::Riok.Mapperly.Sample.MyRefStruct Map(global::Riok.Mapperly.Sample.MyRefStruct source)
{
    var target = new global::Riok.Mapperly.Sample.MyRefStruct();
    target.Span = source.Span;
    return target;
}
```

**Expected behavior**
Mapperly should clone the source span into a new array.
```C#
public static partial global::Riok.Mapperly.Sample.MyRefStruct Map(global::Riok.Mapperly.Sample.MyRefStruct source)
{
    var target = new global::Riok.Mapperly.Sample.MyRefStruct();
    target.Span = new int[source.Length];
    source.Span.CopyTo(target.Span);
    return target;
}
```

**Note**
- In the case of complex objects like managed structs or classes, a for loop should be generated to copy over the source span, similar to how arrays are handled.
- ~Potential optimization for `ReadOnlySpan<T>` where T is a primitve or string type (or possibly a readonly copyable struct), could be to just copy the field over. This is because under these circumstances any elements of `ReadOnlySpan` cannot be changed, assuming unsafe code is not used~.

```C#
[Mapper(UseDeepCloning =true)]
{
    public static partial MyRefStruct Map(MyRefStruct source);
}
public ref struct MyRefStruct
{
    public ReadOnlySpan<int> Span;
}
public static partial global::Riok.Mapperly.Sample.MyRefStruct Map(global::Riok.Mapperly.Sample.MyRefStruct source)
{
    var target = new global::Riok.Mapperly.Sample.MyRefStruct();
    target.Span = source.Span;
    return target;
}
```

TimothyMakkison:
### Description
Resolves #391. 

- Added `SpanMappingBuilder`, mapping is either a cast, `ToArray`/ForLoop, using `EnumerableMappingBuilder` to map an enumerable to an array, or using `ToArray` and then mapping to an enumerable.
- Shares a lot of logic with `EnumerableMappingBuilder`, most logic is different although some minor changes could be made to share logic ie `ResolveImmutableCollectMethod` or `TryBuildExistingTargetMapping`.
- Added two new TypeMapping although I suspect `DelegateMapping` already exists.

Haven't added documentation or `MapperConversionType`

**Mostly complete notes:**
image


TimothyMakkison:
Update: Added Memory mapping.

- Handles mapping to Memory from Span, Memory, Array and Enumerable, also supports mapping to Span, Array and Enumerable from Memory.
- Supports existing target mapping for Memory to Enumerable
- Added unit tests
- Is using Method Mapping encouraged? For most `Memory -> ?` mappings I use `Memory.Span` and pass this into a `SpanTo?` mapping method.

latonz (changes requested):
Thank you for another great contribution 😊
I added my feedback.

src/Riok.Mapperly/Descriptors/Mappings/AsSpanMethodMapping.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Descriptors.Mappings;
+
+public class AsSpanMethodMapping : TypeMapping
+{
+    private const string SpanMemberName = "Span";
```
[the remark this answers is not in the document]
TimothyMakkison: Good idea, wish I'd thought of that.

src/Riok.Mapperly/Descriptors/MappingBuilders/MemoryMappingBuilder.cs
```diff
@@ -0,0 +1,261 @@
...
+namespace Riok.Mapperly.Descriptors.MappingBuilders;
+
+public static class MemoryMappingBuilder
+{
+    private const string SpanMemberName = nameof(Memory<int>.Span);
+    private const string ToArrayMethodName = nameof(Enumerable.ToArray);
+
+    public static TypeMapping? TryBuildMapping(MappingBuilderContext ctx)
+    {
+        if (GetCollectionInfo(ctx.Source, ctx) is not { } sourceInfo)
```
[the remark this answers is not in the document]
TimothyMakkison: I'll add 2 flags for `Span` and `Memory`. 

How specific should the checks be? For instance for should I check for `Memory`,`Enumerable` and `Span` for mapping `Memory` -> `Array`?

TimothyMakkison:
Need to rename / cleanup stuff don't bother reviewing or merging. Thanks 👃

src/Riok.Mapperly/Descriptors/Mappings/SourceObjectMethodDelegateMapping.cs
```diff
@@ -0,0 +1,34 @@
...
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+using static Riok.Mapperly.Emit.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Descriptors.Mappings;
+
+/// <summary>
+/// Represents a mapping which works by invoking an instance method on the source object
+/// and then mapping the result with an optional mapper.
+/// </summary>
+public class SourceObjectMethodDelegateMapping : TypeMapping
```
latonz, line 12: Would exclude the delegate part from the name as long as it is optional.

src/Riok.Mapperly/Descriptors/WellKnownTypes.cs
```diff
@@ -17,6 +18,16 @@ internal WellKnownTypes(Compilation compilation)
...
     public INamedTypeSymbol? TimeOnly => TryGet("System.TimeOnly");
 
+    public IArrayTypeSymbol GetArrayType(ITypeSymbol type) =>
+        _compilation.CreateArrayTypeSymbol(type, elementNullableAnnotation: type.NullableAnnotation);
+
+    public bool IsSpanType(ITypeSymbol symbol) =>
+        SymbolEqualityComparer.Default.Equals(symbol.OriginalDefinition, Get(typeof(Span<>)))
+        || SymbolEqualityComparer.Default.Equals(symbol.OriginalDefinition, Get(typeof(ReadOnlySpan<>)));
+
+    public ITypeSymbol? GetEnumeratedType(ITypeSymbol type) =>
```
latonz, line 28: This could and probably should be used here too:
* `Riok.Mapperly.Descriptors.MappingBuilders.DictionaryMappingBuilder.GetEnumerableKeyValueTypes`
* `Riok.Mapperly.Descriptors.Enumerables.EnsureCapacity.EnsureCapacityBuilder.TryBuildEnsureCapacity`

src/Riok.Mapperly/Descriptors/MappingBuilders/SpanMappingBuilder.cs
```diff
@@ -0,0 +1,270 @@
...
+        MappingBuilderContext ctx
+    )
+    {
+        var enumeratedType = ctx.Types.GetEnumeratedType(symbol);
+        if (enumeratedType != null)
+        {
+            if (symbol.IsArrayType())
+                return (CollectionType.Array, enumeratedType);
+
+            return (CollectionType.Enumerable, enumeratedType);
```
latonz, line 245: I think readability could be improved here by using `?:`.

src/Riok.Mapperly/Descriptors/MappingBuilders/SpanMappingBuilder.cs
```diff
@@ -0,0 +1,270 @@
...
+
+        if (symbol is not INamedTypeSymbol namedTypeSymbol || namedTypeSymbol.TypeArguments.Length != 1)
+            return null;
+
+        var typeSymbol = namedTypeSymbol.TypeArguments[0];
+        if (SymbolEqualityComparer.Default.Equals(symbol.OriginalDefinition, ctx.Types.Get(typeof(Span<>))))
+            return (CollectionType.Span, typeSymbol);
+
+        if (SymbolEqualityComparer.Default.Equals(symbol.OriginalDefinition, ctx.Types.Get(typeof(ReadOnlySpan<>))))
+            return (CollectionType.ReadOnlySpan, typeSymbol);
```
latonz, line 256: I think I'd move the Span/ReadOnlySpan before the array / enumerable checks because it seems to be more intuitive to me. It is not entirely clear that `GetEnumeratedType` just works on `IEnumerable<>` implementations but not on types implementing `GetEnumerator` which would be implemented by span. Therefore moving these checks before the array/enumerable check IMO would make it a lot easier to understand.

src/Riok.Mapperly/Descriptors/MappingBuilders/SpanMappingBuilder.cs
```diff
@@ -0,0 +1,270 @@
...
+        var typedArray = ctx.Types.GetArrayType(enumeratedType);
+        if (ctx.FindOrBuildMapping(typedArray.NonNullable(), ctx.Target) is not { } arrayMapping)
+            return null;
+
+        return new SourceObjectMethodDelegateMapping(ctx.Source, ctx.Target, ToArrayMethodName, arrayMapping);
+    }
+
+    private static bool IsTargetImmutableCollection(MappingBuilderContext ctx)
+    {
+        if (SymbolEqualityComparer.Default.Equals(ctx.Target.OriginalDefinition, ctx.Types.Get(typeof(ImmutableArray<>))))
```
latonz, line 201: Similar checks in the `EnumerableMappingBuilder` and `MemoryMappingBuilder`. I think it would make sense to extract the enum `CollectionType` and create a method in `WellKnownTypes` `CollectionType GetCollectionType(ITypeSymbol type)`. `CollectionType` could include all collection types handled by Mapperly (eg. `List, ImmutableList, Array, Span, ReadOnlySpan, Memory, ...`). The enum could then have extension methods (eg. `bool IsSpan()`, `bool IsImmutable()`, ...).

src/Riok.Mapperly/Descriptors/MappingBuilders/MemoryMappingBuilder.cs
```diff
@@ -0,0 +1,236 @@
...
+
+            // if target is Array
+            // either use ToArray or MapToArray(memory.Span)
+            (_, CollectionType.Array) => BuildMemoryToArrayMapping(ctx, elementMapping),
+
+            // if target is Enumerable
+            // use MapToEnumerable(memory.Span)
+            (_, CollectionType.Enumerable) => BuildMemoryToEnumerableMapping(ctx, elementMapping),
+
+            _ => throw new Exception("Unreachable exception.")
```
latonz, line 73: Instead of this I'd just convert to last case to `_ => ...`

src/Riok.Mapperly/Descriptors/MappingBuilders/MemoryMappingBuilder.cs
```diff
@@ -0,0 +1,236 @@
...
+    {
+        if (!elementMapping.IsSynthetic || ctx.MapperConfiguration.UseDeepCloning)
+        {
+            var targetArray = ctx.Types.GetArrayType(elementMapping.TargetType);
+            if (ctx.FindOrBuildMapping(ctx.Source, targetArray.NonNullable()) is not { } arrayMapping)
+                return null;
+            return new CastMapping(ctx.Source, ctx.Target, arrayMapping);
+        }
+
+        return new SourceObjectMethodDelegateMapping(ctx.Source, ctx.Target, ToArrayMethodName);
```
latonz, line 118: IMO the readability of this method could be improved if the first if is inverted and returned early and the second one is changed into `?:` syntax.

latonz (changes requested):
Thanks for the updates, I added my review comments 😊 don't hesitate to reach out if you have different opinions or questions 😊

test/Riok.Mapperly.IntegrationTests/_snapshots/NET_48/StaticMapperTest.SnapshotGeneratedSource.verified.cs
```diff
@@ -228,6 +232,11 @@ public static partial class StaticTestMapper
             target.NestedNullableTargetNotNullable = MapToTestObjectNested(dto.NestedNullableTargetNotNullable);
             target.StringNullableTargetNotNullable = dto.StringNullableTargetNotNullable;
             target.SourceTargetSameObjectType = dto.SourceTargetSameObjectType;
+            if (dto.SpanValue != null)
+            {
+            }
```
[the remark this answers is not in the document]
TimothyMakkison: mb forgot to create an issue for this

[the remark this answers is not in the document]
TimothyMakkison: See #497

src/Riok.Mapperly/Descriptors/MappingBuilders/SpanMappingBuilder.cs
latonz: While addressing https://github.com/riok/mapperly/issues/498 and https://github.com/riok/mapperly/issues/499 I tried sth similar here: https://github.com/riok/mapperly/pull/501.

TimothyMakkison: Nice, looks super helpful. Should I add the various `Span` & `Memory` types to it? Not sure if there is enough overlap to justify it.

latonz: I think adding `Span`, `Memory`, ... makes sense as the checks etc. are the same and these mappings are usually either from and to a span/memory or include at least one collection.

src/Riok.Mapperly/Descriptors/Mappings/ExistingTarget/SourceObjectMemberDelegateExistingTargetMapping.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
+
+/// <summary>
+/// Represents a mapping which works by accessing an instance member on the source object
+/// and then mapping the result to the target with an existing target mapper.
+/// </summary>
+public class SourceObjectMemberDelegateExistingTargetMapping : ExistingTargetMapping
+{
+    private readonly string _memberName;
+    private readonly IExistingTargetMapping _delegateMapping;
```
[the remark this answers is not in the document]
TimothyMakkison: What would the optional form look like? Because its an existing target mapping it currectly creates `MapToSomething(source.MyMember, target)`. It can't create something like `target = source.MyMember`.

src/Riok.Mapperly/Descriptors/MappingBuilders/SpanMappingBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+            return null;
+
+        return new CastMapping(ctx.Source, ctx.Target, arrayMapping);
+    }
+
+    private static TypeMapping? MapSpanArrayToEnumerableMethod(MappingBuilderContext ctx)
+    {
+        var enumeratedType = ctx.CollectionInfos!.Source.EnumeratedType;
+        var typedArray = ctx.Types.GetArrayType(enumeratedType);
+        if (ctx.FindOrBuildMapping(typedArray.NonNullable(), ctx.Target) is not { } arrayMapping)
```
latonz, line 203: Is the `NonNullable()` really correct here? `ToArray()` can never return `null`, can it? Probably `GetArrayType` should always return non-nullable type (while still copying the element nullability). Same on L193.

src/Riok.Mapperly/Descriptors/WellKnownTypes.cs
```diff
@@ -17,6 +17,13 @@ internal WellKnownTypes(Compilation compilation)
 
     public INamedTypeSymbol? TimeOnly => TryGet("System.TimeOnly");
 
+    public IArrayTypeSymbol GetArrayType(ITypeSymbol type) =>
+        _compilation.CreateArrayTypeSymbol(type, elementNullableAnnotation: type.NullableAnnotation);
+
+    public bool IsSpanType(ITypeSymbol symbol) =>
```
latonz, line 23: To be discussed: this information could instead probably resolved via the `CollectionInfo`...

latonz (changes requested):

src/Riok.Mapperly/Descriptors/WellKnownTypes.cs
TimothyMakkison: This is only used in `EnsureCapacityBuilder.TryGetNonEnumeratedCount`. One of the main functions doesn't take `CollectionInfo` as a parameter.

TimothyMakkison: Its only used once in `BuildSpanToEnumerable`. I can use `TryBuildMapping` to generate it instead.

latonz:
@TimothyMakkison is this ready to be merged? Could you rebase it?

TimothyMakkison:
I've rebased it and it should be ready to be merge.

I've got a bad vide [with line](https://github.com/riok/mapperly/pull/425/files#diff-c6aa5add0ebfc429802588710a9bca6220e0c082261fbbe3a8836a9362e93195R100), I'm not sure if it should check for `Span`, `ReadOnlySpan`, `Memory` etc. The next step double checks the collection type so it should filter out most false positives.

latonz:
@TimothyMakkison on a second thought I think checking for Span, ReadOnlySpan, Memory is a good idea 👍

TimothyMakkison:
> @TimothyMakkison on a second thought I think checking for Span, ReadOnlySpan, Memory is a good idea 👍

Done, thanks for the help 🙏 

Might be able to speed this check up in #530

latonz (approved):

PR #467, opened 2023-05-29
feat: added tuple mapping support

### LINKED ISSUE #448: Improve tuple support
**Describe the bug**
Mapperly incorrectly attempts to construct tuples via simplified form. Further when named fields are used it will attempt to map both the named field and the default field identifier. See also #447.

**To Reproduce**
```C#
[Mapper(UseDeepCloning = true)]
public static partial class Map
{
    public static partial (int Value, int Num) TupleMapper(this (int Value, int Num) input);
}
```
```C#
public static partial (int Value, int Num) TupleMapper(this (int Value, int Num) input)
{
    var target = new (int Value, int Num)(); // this is an error
    target.Item1 = input.Item1;  // maps both default names and aliased named
    target.Value = input.Value;
    target.Item2 = input.Item2;
    target.Num = input.Num;
    return target;
}
```

**Expected behavior**
Tuples should be generated with `new ValueTuple<...>(...)` syntax and Mappely should track when fields are mapped. Mapperly should only try to map via the aliased named first, only using the positional parameters as a fallback.

TimothyMakkison:
**Description**
Fixes #448 Added support for tuple mapping, supported mapping attributes and repeat member mappings are no longer created.

- `Class -> Tuple` and `Tuple -> Class` will map via field names, or if a field doesn't have a name its default positional name is used.
- `Tuple` to `Tuple` will try to map via field names, otherwise it will attempt to match fields by position.
- Support for `IgnoreSource`, `IgnoreProperty` and `MapProperty`.
- - `IgnoreSource` and `MapProperty` will only use field names when available, otherwise if the field doesn't have a name it will use the default positional name.

Question: currently a simple tuple expression is used, should the targets field names be added?
```C#
public (int A, int B) Map((int C, int D) source)
{
  // current mapping
  return (source.C, source.D);

  // slightly verbose alternative mapping
  return (A: source.C, B: source.D);
}
```

I'll update the integration test when we decide on this behaviour


src/Riok.Mapperly/Descriptors/Mappings/NewTupleExpressionMapping.cs
```diff
@@ -0,0 +1,30 @@
...
+    public NewTupleExpressionMapping(ITypeSymbol sourceType, ITypeSymbol targetType)
+        : base(sourceType, targetType) { }
+
+    public void AddConstructorParameterMapping(TupleConstructorParameterMapping mapping) => _constructorPropertyMappings.Add(mapping);
+
+    public override ExpressionSyntax Build(TypeMappingBuildContext ctx)
+    {
+        // (Name:.. ,..);
+        var ctorArgs = _constructorPropertyMappings.Select(x => x.BuildArgument(ctx)).ToArray();
+        var objectCreationExpression = SyntaxFactory.TupleExpression(SyntaxFactory.SeparatedList(ctorArgs));
```
latonz, line 26: This does not work in expressions, does it?

src/Riok.Mapperly/Descriptors/Mappings/NewTupleExpressionMapping.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        return objectCreationExpression.WithLeadingTrivia();
```
latonz, line 28: Is the empty leading trivia really required?

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -83,11 +83,20 @@ internal static IEnumerable<ISymbol> GetAllMembers(this ITypeSymbol symbol)
         IEqualityComparer<string> comparer
     )
     {
+        if (symbol.IsTupleType && symbol is INamedTypeSymbol namedType)
+            return namedType.TupleElements
```
latonz, line 87: Add braces around multiline if statements.

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/TupleConstructorParameterMapping.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    public IFieldSymbol Parameter { get; }
+
+    public NullMemberMapping DelegateMapping { get; }
+
+    public ArgumentSyntax BuildArgument(TypeMappingBuildContext ctx)
+    {
+        var argumentExpression = DelegateMapping.Build(ctx);
+        var arg = SyntaxFactory.Argument(argumentExpression);
+        return arg;
```
latonz, line 23: Statically use `SyntaxFactory` and return directly.

src/Riok.Mapperly/Descriptors/Mappings/NewTupleExpressionMapping.cs
latonz, line 26: Statically import SyntaxFactory...

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/INewTupleExpressionBuilderContext.cs
```diff
@@ -0,0 +1,14 @@
+using Riok.Mapperly.Descriptors.Mappings;
+using Riok.Mapperly.Descriptors.Mappings.MemberMappings;
+
+namespace Riok.Mapperly.Descriptors.MappingBodyBuilders.BuilderContext;
+
+/// <summary>
+/// A <see cref="INewTupleExpressionBuilderContext{T}"/> for mappings which create the target object via ().
```
latonz, line 7: Is the reference to `INewTupleExpressionBuilderContext` correct here? Shouldn't it be `IMembersBuilderContext`?

latonz, line 7: I find the `... create the target object via ()` part confusing as the first time reading it I was not sure what `()` was referring to. Probably rephrase it to sth. like `... create the target tuple via a tuple expression (eg. (A: MapToA(src.A), B: MapToB(src.B)))`

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/NewTupleExpressionBuilderContext.cs
```diff
@@ -0,0 +1,22 @@
...
+/// An implementation of <see cref="INewTupleExpressionBuilderContext{T}"/>.
+/// </summary>
+/// <typeparam name="T">The type of the mapping.</typeparam>
+public class NewTupleExpressionBuilderContext<T> : MembersMappingBuilderContext<T>, INewTupleExpressionBuilderContext<T>
+    where T : INewTupleExpressionMapping
+{
+    public NewTupleExpressionBuilderContext(MappingBuilderContext builderContext, T mapping)
+        : base(builderContext, mapping) { }
+
+    public void AddConstructorParameterMapping(TupleConstructorParameterMapping mapping)
```
latonz, line 16: I'd rename to more specific `AddTupleConstructorParameterMapping`

src/Riok.Mapperly/Descriptors/MappingBuilders/ExplicitCastMappingBuilder.cs
```diff
@@ -16,6 +16,9 @@ public static class ExplicitCastMappingBuilder
         if (ctx.MapperConfiguration.UseDeepCloning && !ctx.Source.IsImmutable() && !ctx.Target.IsImmutable())
             return null;
 
+        if (ctx.Target.IsTupleType)
```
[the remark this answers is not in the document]
TimothyMakkison: Added comment:
ClassifyConversion does not check if tuple field member names are the same, if tuple check isn't done then `(A: int, B: int) -> (B: int, A: int)` would be mapped with:
`return ((int, int))source;` instead of `return (B: source.A, A: source.B);`

src/Riok.Mapperly/Descriptors/Mappings/NewTupleExpressionMapping.cs
TimothyMakkison: I'm not sure what you mean, could you elaborate?

TimothyMakkison:
Thanks for the review 🙏 
I've added the named fields, opting to not add the name colon when the mapping is positional so: `(int, int A) -> (int, int A)` becomes: `return (source.Item1, A: source.A)`

I'll squash the commits and faff around with the integration tests when you're happy with these changes 👍

src/Riok.Mapperly/Descriptors/Mappings/NewTupleExpressionMapping.cs
latonz: Mapperly supports Mappings in expressions trees for Queryable projection mappings (see [docs](https://mapperly.riok.app/docs/configuration/queryable-projections)). AFAIK tuple expressions (eg. `(A: 10, B: 20)`) are not valid in expression trees (eg. this does not work: `Expression<Func<(int, int)>> x = () => (10, 20);`.
The above mapping is constructed even if the mapping takes place in an expression tree. I think this should be differentiated: in expression trees `new ValueTuple<int, int>(10, 20)` should be used, outside `(A: 10, B: 20)` should be used for increased readability.

latonz:
Looks good for me 😊 
Only open discussion to be resolved: https://github.com/riok/mapperly/pull/467#discussion_r1211783801

TimothyMakkison:
> tuple expressions (eg. (A: 10, B: 20)) are not valid in expression trees (eg. this does not work: Expression<Func<(int, int)>> x = () => (10, 20);.

Interesting, I wonder why they enver added it.

I've partially added `ValueTuple` support although I haven't cleaned up or properly added the integration tests, so please don't review this.

I added `ValueTuple` by adding an `isExpression` bool in the constructor, this is used to determine whether a tuple expression or ValueTuple should be used. `valueTupleSymbol` is used to get the identifier for `ValueTuple`, I know passing `ISymbol` to a mapping is an anti pattern, should a constant string be used instead? Note `isExpression` could probably be removed with the nullability of `valueTupleSymbol` be used instead.

This is very hacky, should I create `NewTupleExpressionMapping` and `VewValueTupleMapping` instead and reuse the mapping logic?

latonz:
@TimothyMakkison I think separating the two mappings makes the code easier to read, therefore I'd separate them as `ValueTupleExpressionMapping` (`(A: 10, B: 20)`) and `ValueTupleConstructorMapping` (`new ValueTuple<int, int>(10, 10)`).

TimothyMakkison:
Rebased, renamed files and updated tests. 

I wonder if it's possible to configure `Verify` to show the line number and text diff of any issues. It might make updating the tests faster.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewValueTupleMappingBodyBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+                    StringComparer.OrdinalIgnoreCase,
+                    out sourcePath
+                )
+            )
+                return true;
+
+            // if standard matching fails, use the underlying field
+            // if source is a tuple compare the underlying field ie, Item1, Item2
+            // if source is not a tuple compare top level members
+            if (ctx.Mapping.SourceType.IsTupleType)
```
latonz, line 138: I'd invert this check and return early, probably even combine with the check on L140.

TimothyMakkison:
This PR is near completion.

I've added support for `MapProperty` making `TupleExpression`  a method mapping if thats okay. I've played around with making tuple methods either directly return the tuple if it doesn't map to any child members, otherwise it uses `var target = (src.A, src.B); target.B.C = src.C; return target;` syntax. Ideally I'd skip creating a method when it doesn't use `MapProperty` but doing it to mapperly feel a little hacky.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewValueTupleMappingBodyBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+                var mappableMember = namedType.TupleElements
+                    .Where(
+                        x =>
+                            x.CorrespondingTupleField != default
+                            && !ctx.IgnoredSourceMemberNames.Contains(x.Name)
+                            && string.Equals(
+                                field.CorrespondingTupleField!.Name,
+                                x.CorrespondingTupleField!.Name,
+                                StringComparison.OrdinalIgnoreCase
+                            )
```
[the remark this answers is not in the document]
TimothyMakkison: Added `PropertyMappingStrategy` for the normal fields, this section is used for positional mapping ie `"Item1" == "Item1"` so letter casing doesn't matter

TimothyMakkison:
- Added `MappingConversionType.Tuple`
- Added full `MapProperty` support, including for nested properties. - The logic is a little hacky but could be reused for new instance initializers.
- Tuple expression will either be a normal expression (`target.T = (src.A, ...)`) or generate a method if `MapProperty` is used on nested properties (`target.T = MapToTuple(src)`). I achieved this by making `MethodDelcarationSyntax` nullable, I suspect a neater api change could be done instead.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewValueTupleMappingBodyBuilder.cs
```diff
@@ -0,0 +1,225 @@
...
+        if (!TryBuildTupleConstructorMapping(ctx, namedTargetType, out var constructorParameterMappings, out var mappedTargetMemberNames))
+        {
+            ctx.BuilderContext.ReportDiagnostic(DiagnosticDescriptors.NoConstructorFound, ctx.BuilderContext.Target);
+            return;
+        }
+
+        var removableMappedTargetMemberNames = mappedTargetMemberNames.Where(x =>
+        {
+            ctx.MemberConfigsByRootTargetName.TryGetValue(x, out var value);
+            return !value?.Any() ?? true;
```
latonz, line 47: Couldn't `ContainsKey` be used here instead?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/NewValueTupleExpressionBuilderContext.cs
```diff
@@ -0,0 +1,26 @@
...
+    where T : INewValueTupleMapping, IMemberAssignmentTypeMapping
+{
+    public NewValueTupleExpressionBuilderContext(MappingBuilderContext builderContext, T mapping)
+        : base(builderContext, mapping) { }
+
+    public void AddTupleConstructorParameterMapping(ValueTupleConstructorParameterMapping mapping)
+    {
+        if (MemberConfigsByRootTargetName.TryGetValue(mapping.Parameter.Name, out var value))
+        {
+            value.RemoveAll(x => x.Target.Path.Count == 1);
```
latonz, line 20: If `value` is empty after removing these values, we should remove it from `MemberConfigsByRootTargetName` to make the code at other locations easier.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewValueTupleMappingBodyBuilder.cs
```diff
@@ -0,0 +1,225 @@
...
+        mappedTargetMemberNames = new HashSet<string>();
+        constructorParameterMappings = new HashSet<ValueTupleConstructorParameterMapping>();
+
+        foreach (var targetMember in namedTargetType.TupleElements)
+        {
+            if (!ctx.TargetMembers.ContainsKey(targetMember.Name))
+            {
+                return false;
+            }
+            if (!TryFindConstructorParameterSourcePath(ctx, targetMember, out var sourcePath))
```
latonz, line 73: Nit: add an empty line after a block.

src/Riok.Mapperly/Descriptors/MappingBuilders/NewInstanceObjectPropertyMappingBuilder.cs
```diff
@@ -26,6 +26,14 @@ public static class NewInstanceObjectPropertyMappingBuilder
...
             return null;
 
+        if (ctx.Target is INamedTypeSymbol { IsTupleType: true })
+        {
+            // inline expressions don't support tuple expressions so ValueTuple is used instead
+            return ctx.IsExpression
+                ? new NewValueTupleConstructorMapping(ctx.Source, ctx.Target, ctx.Types.Get<ValueTuple>())
+                : new NewValueTupleExpressionMapping(ctx.Source, ctx.Target);
+        }
+
```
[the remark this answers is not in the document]
latonz: This isn't resolved, is it?

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/ValueTupleConstructorParameterMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+    public IFieldSymbol Parameter { get; }
+
+    public NullMemberMapping DelegateMapping { get; }
+
+    public ArgumentSyntax BuildArgument(TypeMappingBuildContext ctx, bool emitFieldName)
+    {
+        var argumentExpression = DelegateMapping.Build(ctx);
+        var argument = Argument(argumentExpression);
+
+        // tuples inside expression cannot use the expression form (A: .., ..) instead new ValueTuple<>(..) must be used instead
```
latonz, line 24: Nit: `instead` is repeated twice

src/Riok.Mapperly/Descriptors/Mappings/NewValueTupleConstructorMapping.cs
```diff
@@ -0,0 +1,33 @@
...
+    public void AddConstructorParameterMapping(ValueTupleConstructorParameterMapping mapping) => _constructorPropertyMappings.Add(mapping);
+
+    public override ExpressionSyntax Build(TypeMappingBuildContext ctx)
+    {
+        // new ValueTuple<T..>(ctorArgs)
+        var ctorArgs = _constructorPropertyMappings.Select(x => x.BuildArgument(ctx, emitFieldName: false)).ToArray();
+        var genericName = GenericName(ValueTupleName);
+        var typeArguments = TypeArgumentList((TargetType as INamedTypeSymbol)!.TypeArguments.Select(NonNullableIdentifier));
+        var typedValue = genericName.WithTypeArgumentList(typeArguments);
+        return ObjectCreationExpression(typedValue).WithArgumentList(ArgumentList(ctorArgs));
```
latonz, line 31: Instead of collecting `ctorArgs` to array, `ArgumentList` could just expose an overload accepting an enumerable, as the called `CommaSeparatedList` only works on the enumerable anyway.

src/Riok.Mapperly/Descriptors/Mappings/NewValueTupleConstructorMapping.cs
```diff
@@ -0,0 +1,33 @@
...
+        : base(sourceType, targetType) { }
+
+    public void AddConstructorParameterMapping(ValueTupleConstructorParameterMapping mapping) => _constructorPropertyMappings.Add(mapping);
+
+    public override ExpressionSyntax Build(TypeMappingBuildContext ctx)
+    {
+        // new ValueTuple<T..>(ctorArgs)
+        var ctorArgs = _constructorPropertyMappings.Select(x => x.BuildArgument(ctx, emitFieldName: false)).ToArray();
+        var genericName = GenericName(ValueTupleName);
+        var typeArguments = TypeArgumentList((TargetType as INamedTypeSymbol)!.TypeArguments.Select(NonNullableIdentifier));
```
latonz, line 29: Instead of `(TargetType as INamedTypeSymbol)!` a simple cast can be used.

src/Riok.Mapperly/Descriptors/Mappings/NewValueTupleExpressionMapping.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        // generate method call if it's a complex mapping
+        if (HasMemberMappings())
+        {
+            return base.Build(ctx);
+        }
+
+        // (Name:.. ,..);
+        var ctorArgs = _constructorPropertyMappings.Select(x => x.BuildArgument(ctx, emitFieldName: true)).ToArray();
+        return TupleExpression(SeparatedList(ctorArgs));
```
latonz, line 52: As `SeparatedList` accepts an enumerable as parameter there is no need to collect into an array.

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -185,6 +185,11 @@ private IEnumerable<ISymbol> GetAllMembersCore(ITypeSymbol symbol)
 
     private IEnumerable<IMappableMember> GetAllAccessibleMappableMembersCore(ITypeSymbol symbol)
     {
+        if (symbol.IsTupleType && symbol is INamedTypeSymbol namedType)
+        {
+            return namedType.TupleElements.Where(x => x is { Kind: SymbolKind.Field }).Select(MappableMember.Create).WhereNotNull();
```
latonz, line 190: Is the where part here needed, aren't these filtered by create+whereNotNull anyway? Does it bring performance benefits? If so could we move it to the top of the create method and return null? Also for single field assertions this code base usually uses the `x.Kind == SymbolKind.Field` style instead of pattern matching.

test/Riok.Mapperly.Tests/Mapping/ValueTupleTest.cs
```diff
@@ -0,0 +1,512 @@
...
+    }
+
+    [Fact]
+    public void TupleToTupleWithIgnoredSource()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+                [MapperIgnoreSource("C")]
+                partial (int, string) Map((int A, string B, int C) source);
+                """
```
latonz, line 167: Nit: Alignment. (the same at several other lines in this file...)

latonz (changes requested):
Thank you for the updates.
Added a few smaller feedbacks...
 
Regarding the inlined method mapping: For now I'd keep it simple and always generate a method for a tuple expression mapping. The compiler will probably inline it anyway. If we come to the conclusion it is worth to inline these one-liner methods, IMO we should do so in a separate PR, which could probably also affect other method mappings and maybe even be handled on a `MethodMapping` level.

TimothyMakkison:
Thanks will make the changes later. 👍 

> Regarding the inlined method mapping: For now I'd keep it simple and always generate a method for a tuple expression mapping. The compiler will probably inline it anyway. If we come to the conclusion it is worth to inline these one-liner methods, IMO we should do so in a separate PR, which could probably also affect other method mappings and maybe even be handled on a MethodMapping level.

I already added a slightly hacky way of method inlining. Do you want me to remove it?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewValueTupleMappingBodyBuilder.cs
TimothyMakkison: Needs to check if the key doesn't exist or that there aren't any values in the list.

latonz: Why do you need to check whether the list is empty? There should never be an empty list to a key. If the list gets empty it should be removed from the dictionary.

TimothyMakkison:
Thanks for the review, I've removed method inlining and resolved all your suggested changes. 

I'm converting this pr to a draft until I proof read it.

latonz (approved):

Issue #477, opened 2023-06-02
Run Benchmarks in GitHub Actions


latonz:
Run Benchmarks in GitHab Actions to detect impact of a PR on the performance. 

Probably https://github.com/benchmark-action/github-action-benchmark could be used.

latonz:
Implemented by https://github.com/riok/mapperly/pull/632

PR #504, opened 2023-06-15
chore: add benchmark action

TimothyMakkison:
# Added benchmark workflow

## Description
Added `github-action-benchmark` to run for changes to main. I've tried to include the changes from the previous discussion here #419.

Resovles #477

This needs a branch called `gh-pages` to visualize the benchmarks, [See here](https://github.com/benchmark-action/github-action-benchmark#charts-on-github-pages-1) for more information.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,35 @@
+name: Benchmark
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  benchmark:
+    name: Run Benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


TimothyMakkison:
[From comment](https://github.com/riok/mapperly/pull/419#discussion_r1195375495)
> IMO the benchmarks should run on each PR shouldn't they?
But the results should only be updated when merged onto main. WDYT?

Enabling the benchmark for all pull requests will result in the file being written to and is not recommended, I could either add a separate workflow for pull requests or try to use `save-data-file` with `github.event_name != 'pullrequest'` (not sure if this is possible)

TimothyMakkison:
Why did it take 6 hours for the CI to fail?

latonz:
@TimothyMakkison no Idea, which job took so long?

TimothyMakkison:
Looks like the test. Wonder if github was having issues yesterday.
image

PR #517, opened 2023-06-23
chore: attribute data accessor should resolve arguments based on target type

### LINKED ISSUE #486: `AttributeDataAccessor` fails to construct `MapEnumValue`
**Describe the bug**
I noticed the benchmarks were only using 6000KB of memory. When I ran `SourceGeneratorBenchmarks.LargeCompile()` I got the following message:

`System.MissingMethodException: Constructor on type 'Riok.Mapperly.Configuration.MapEnumValue' not found.
   at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture)
   at System.Activator.CreateInstance(Type type, Object[] args)
   at Riok.Mapperly.Configuration.AttributeDataAccessor.Access[TAttribute,TData](WellKnownTypes knownTypes, ISymbol symbol)+MoveNext() in /_/src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs:line 41
   at Riok.Mapperly.Descriptors.MappingBuilders.EnumMappingBuilder.BuildExplicitValueMapping(MappingBuilderContext ctx) in /_/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumMappingBuilder.cs:line 144`

Bizarrely the integration or unit tests don't fail. I assume this is due to #478 and `MapEnumValue` being both a record type and attribute. This could be caused by `Activator` trying to create `MapEnumValue(TypedConstant, TypedConstant)` from `ctorArguments` where `BuildArgumentValue` have cast the values to integer objects.

latonz:
Closes #486

### DIFF
--- a/src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
+++ b/src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
@@ -50,10 +50,8 @@ internal class AttributeDataAccessor
 
         foreach (var attrData in attrDatas)
         {
-            var typeArguments = attrData.AttributeClass?.TypeArguments ?? Enumerable.Empty<ITypeSymbol>();
-            var ctorArguments = attrData.ConstructorArguments.Select(BuildArgumentValue);
-            var newInstanceArguments = typeArguments.Concat(ctorArguments).ToArray();
-            var attr = (TData)Activator.CreateInstance(dataType, newInstanceArguments);
+            var typeArguments = (IReadOnlyCollection<ITypeSymbol>?)attrData.AttributeClass?.TypeArguments ?? Array.Empty<ITypeSymbol>();
+            var attr = Create<TData>(typeArguments, attrData.ConstructorArguments);
 
             foreach (var namedArgument in attrData.NamedArguments)
             {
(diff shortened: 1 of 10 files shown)


latonz:
@TimothyMakkison do the benchmarks work on your machine with this patch?

TimothyMakkison:
Looks like it should but I haven't tried it yet.

You should be able to check by running the benchmarks before and after commenting out the map enum value / fallback enum integration code. You would see an increase in memory and time with them commented out. We're You able to reproduce it?

latonz:
@TimothyMakkison seems to work on my machine, however, I'd be glad if you could verify 😊

TimothyMakkison:
> @TimothyMakkison seems to work on my machine, however, I'd be glad if you could verify 😊

It works, no exceptions and the benchmarks look the same.
Thanks for making the pull request, hopefully #504 works.

TimothyMakkison:
I'm assuming you don't expect attributes to have both generic and constructor arguments or `*Configuration` to have generic parameters right? Otherwise there are two logic bugs.

latonz:
@TimothyMakkison thank you for the hint, I fixed the logic that it works correctly if the attribute has generic parameters which are passed to the data class constructor as `ITypeSymbol`.

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
```diff
@@ -50,67 +50,103 @@ public IEnumerable<TAttribute> Access<TAttribute>(ISymbol symbol)
...
-            ?? throw new InvalidOperationException("Array typed constant is not of type " + nameof(IArrayTypeSymbol));
+        if (arg.Kind == TypedConstantKind.Array)
+        {
+            var values = arg.Values.Select(x => (string?)BuildArgumentValue(x, typeof(string))).WhereNotNull().ToList();
+            return new StringMemberPath(values);
+        }
+
+        if (arg is { Kind: TypedConstantKind.Primitive, Value: string v })
+        {
+            return new StringMemberPath(v.Split('.'));
```
ni507, line 120: Extract the '.' into a utils or somewhere else to reuse it for the 'StringMemberPath' class.

ni507 (changes requested):

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
latonz: Fixed.

ni507 (approved):

TimothyMakkison:
Nice, thanks for adding this. Might be able to add #490 now.

PR #518, opened 2023-06-23
feat: introduce experimental full nameof support

latonz:
# introduce experimental full nameof support

## Description

Introduces experimental full nameof support.
See #401.

This would allow simpler `MapProperty` attributes for nested properties:
`[MapProperty(nameof(@A.B.C), nameof(@D.E.F)]` would be the same as `[MapProperty(new[]{nameof(A.B), nameof(A.B.C)}, new[]{nameof(D.E), nameof(D.E.F)}]`.

## Checklist

### DIFF
--- a/docs/docs/configuration/flattening.md
+++ b/docs/docs/configuration/flattening.md
@@ -14,9 +14,31 @@ by either using the source and target property path names as arrays or using a d
 [MapProperty(new[] { nameof(Car.Make), nameof(Car.Make.Id) }, new[] { nameof(CarDto.MakeId) })]
 // Or alternatively
 [MapProperty("Make.Id", "MakeId")]
+// Or
+[MapProperty($"{nameof(Make)}.{nameof(Make.Id)}", "MakeId")]
 partial CarDto Map(Car car);
 ```
 
 :::info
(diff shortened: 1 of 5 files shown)


TimothyMakkison:
Nice, what made you reconsider this?

I'm not sure if it's possible, but is it possible to check for and warn the user when a qualified namespace is used?

latonz:
@TimothyMakkison this is just an experiment 😊 I wanted to poke around and see what it gets. It is still far from ready to be merged, there is still a lot to consider and work out. Not sure if it should relay on the @ prefix, how to handle the namespace thing, ..., ...

TimothyMakkison:
I thought the @ symbol is a pretty clever compromise. The namespace thing isn't a big deal. Hopefully the error and documentation would inform people.
Perhaps when a lookup fails mapperly could scan all available assemblies and types looking for a matching namespace. Not sure how slow this is

CommonGuy (approved):

latonz:
`CallerArgumentExpressionAttribute` does not work, because the Roslyn apis resolve only the compile type value provided to the constructor. Mapperly then constructs the `PropertyMappingConfiguration` type via reflection. Since it is initialized via reflection the `CallerArgumentExpressionAttribute` cannot be applied.

PR #524, opened 2023-06-27
chore: refactor MapperGenerator to cache outputs.

### LINKED ISSUE #72: Improve incremental source generator support

TimothyMakkison:
# Refactored `MapperGenerator` to use pipeline caching

## Description
Refactored `MapperGenerator` to use pipeline caching based upon technique used [here](https://github.com/CommunityToolkit/dotnet/blob/main/src/CommunityToolkit.Mvvm.SourceGenerators/Input/RelayCommandGenerator.cs#L23). 

Because of the use of `Compilation` it is hard to cache any of the `Descriptor` logic, instead I moved the bulk of the generation and creation logic into a single pipeline step. All descriptors are calculated together yielding `(ImmutableArray<(MemberDeclarationSyntax, string)>` and `ImmutableArray<Diagnostic>` warnings/errors. The warning are reported and the mapping name pairs are split into individual values. Finally `RegisterImplementationSourceOutput` runs `NormalizeWhitespace`, creates a `CompilationUnitSyntax`, gets `SourceText.From` and adds the source. 

Hopefully subsequent compilations will recognise that `(MemberDeclarationSyntax, string)` is the same, skipping `AddSource` and the costly `NormalizeWhitespace` (usually 20% of the memory usage).

This doesn't appear to change any of the benchmarks but should improve IDE responsiveness 🤞 . 

Please double check my logic. Is there a way to see if caching is occuring? Source generators are still a black box to me. I might have to return `EquatableArrays` from `BuildDescriptors`

Helps #72

## Checklist


src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -22,39 +22,67 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
         var mapperClassDeclarations = SyntaxProvider.GetClassDeclarations(context);
 
         var compilationAndMappers = context.CompilationProvider.Combine(mapperClassDeclarations.Collect());
-        context.RegisterImplementationSourceOutput(compilationAndMappers, static (spc, source) => Execute(source.Left, source.Right, spc));
+        IncrementalValueProvider<(
+            ImmutableArray<(MemberDeclarationSyntax, string)> Mappers,
+            ImmutableArray<Diagnostic> Diagnostics
+        )> mappersWithErrors = compilationAndMappers.Select((x, token) => BuildDescriptors(x.Left, x.Right, token));
```
latonz, line 28: I think it would make sense to create a class for the Mappers / Diagnostic combination.

src/Riok.Mapperly/Emit/SourceEmitter.cs
```diff
@@ -8,16 +8,19 @@ namespace Riok.Mapperly.Emit;
...
     {
         var sourceEmitterContext = new SourceEmitterContext(descriptor.Symbol.IsStatic, descriptor.NameBuilder);
         MemberDeclarationSyntax member = ClassDeclaration(descriptor.Syntax.Identifier)
             .WithModifiers(descriptor.Syntax.Modifiers)
             .WithMembers(List(BuildMembers(descriptor, sourceEmitterContext)));
 
         member = WrapInClassesAsNeeded(descriptor.Symbol, member);
-        member = WrapInNamespaceIfNeeded(descriptor.Namespace, member);
+        return WrapInNamespaceIfNeeded(descriptor.Namespace, member);
+    }
```
latonz, line 20: Couldn't the `CompilationUnitSyntax` be returned here, just without the `NormalizeWhiteSpace`? Then the MapperGenerator could call `NormalizeWhiteSpace` in the very last step.

latonz (changes requested):
Thank you for opening this PR.

To test each intermediate step you can use `Microsoft.CodeAnalysis.GeneratorRunResult.TrackedSteps` / `Microsoft.CodeAnalysis.GeneratorRunResult.TrackedOutputSteps` (eg. `Riok.Mapperly.Tests.TestHelper.Generate().GetRunResult()..Results[0].TrackedSteps`). We should definitely test whether the caching between the steps works. Eg. the reason on each step can be used to assert this.

TimothyMakkison:
Thanks for the information, I wish incremental generators had better documentation. 

- Added incremental source generator tests
- Added `MapperResult` and `MapperNode` types, I've opted to use `IsEquivalentTo` for the node comparisons, I'm not sure if this is a good idea, it's probably slow in large projects/mappers.

test/Riok.Mapperly.Tests/Generator/IncrementalGeneratorTest.cs
```diff
@@ -0,0 +1,161 @@
...
+            namespace Test.A
+            {
+                [Mapper]
+                internal partial class FooBarMapper
+                {
+                    internal partial string FooToBar(string value);
+                }
+            }
+            """
+        );
```
latonz, line 99: Instead of these mapper sources couldn't `Riok.Mapperly.Tests.TestSourceBuilder.Mapping("string", "string")` be used?

test/Riok.Mapperly.Tests/Generator/IncrementalGeneratorTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        var syntaxTree = CSharpSyntaxTree.ParseText(source, CSharpParseOptions.Default);
+        var compilation1 = TestHelper.BuildCompilation(TestHelperOptions.NoDiagnostics.NullableOption, syntaxTree);
+
+        var driver1 = TestHelper.TrackingGeneratorDriver().RunGenerators(compilation1);
+
+        var compilation2 = compilation1.AddSyntaxTrees(CSharpSyntaxTree.ParseText("struct MyValue {}", CSharpParseOptions.Default));
```
latonz, line 33: I think it would make sense to add a `SyntaxTree` method (and probably additional overloads to `Mapping`: `MappingSyntaxTree`) to the `TestSourceBuilder` to improve the readability of the tests. Eg. `TestSourceBuilder.SyntaxTree("struct MyValue {}")`. This would also improve IDE support as some IDE's recognise the `StringSyntax` `csharp` (eg. Rider), which `ParseText` doesn't have.

test/Riok.Mapperly.Tests/TestHelper.cs
```diff
@@ -80,4 +71,27 @@ private static CSharpCompilation BuildCompilation(NullableContextOptions nullabl
 
         return CSharpCompilation.Create("Tests", syntaxTrees, references, compilationOptions);
     }
+
+    public static GeneratorDriver TrackingGeneratorDriver()
```
latonz, line 75: Instead of this method, I think `GenerateTracked(Compilation)` could be exposed, which calls `RunGenerators` on the driver (each caller of this method calls `RunGenerators` anyway).

test/Riok.Mapperly.Tests/Generator/IncrementalGeneratorTest.cs
```diff
@@ -0,0 +1,161 @@
...
+        var newRoot = syntaxTree.GetCompilationUnitRoot().ReplaceNode(classDeclaration, updatedClass);
+        var newTree = syntaxTree.WithRootAndOptions(newRoot, syntaxTree.Options);
+
+        var compilation2 = compilation1.ReplaceSyntaxTree(compilation1.SyntaxTrees.First(), newTree);
+        var driver2 = driver1.RunGenerators(compilation2);
+
+        AssertRunResults(driver2, IncrementalStepRunReason.New);
+    }
+
+    private static void AssertRunResults(GeneratorDriver driver, params IncrementalStepRunReason[] runReasons)
```
latonz, line 154: Diagnostics should probably also be tested. FYI the `IncrementalValuesProvider` has a `WithTrackingName` method to set step tracking names.

src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -22,39 +22,63 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
         var mapperClassDeclarations = SyntaxProvider.GetClassDeclarations(context);
 
         var compilationAndMappers = context.CompilationProvider.Combine(mapperClassDeclarations.Collect());
-        context.RegisterImplementationSourceOutput(compilationAndMappers, static (spc, source) => Execute(source.Left, source.Right, spc));
+        var mappersWithErrors = compilationAndMappers.Select((x, token) => BuildDescriptors(x.Left, x.Right, token));
```
latonz, line 25: I'd rename to `mappersWithDiagnostics` as these are not necessarily errors.

src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -22,39 +22,63 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+
+        // output the diagnostics
+        context.ReportDiagnostics(mappersWithErrors.Select(static (source, _) => source.Diagnostics));
+
+        // split into mapper name pairs
+        var mappers = mappersWithErrors.SelectMany((x, _) => x.Mappers);
```
latonz, line 31: This select many lambda could also be static, couldn't it?

src/Riok.Mapperly/MapperGenerator.cs
latonz, line 25: This select many lambda could also be static, couldn't it?

src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -22,39 +22,63 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+
+        context.RegisterImplementationSourceOutput(
+            mappers,
+            static (spc, source) =>
+            {
+                spc.AddSource(source.FileName, SourceText.From(source.Body.NormalizeWhitespace().ToFullString(), Encoding.UTF8));
```
latonz, line 37: A lot going on on this line, probably split it into multiple lines to increase readability.

src/Riok.Mapperly/MapperGenerator.cs
latonz, line 25: Would rename `token` to `ct` or `cancellationToken` as only token seems more syntax related to me in this context.

latonz (changes requested):

src/Riok.Mapperly/MapperResults.cs
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public ImmutableArray<MapperNode> Mappers { get; }
+    public ImmutableArray<Diagnostic> Diagnostics { get; }
+
+    public static MapperResults Empty() => new(ImmutableArray<MapperNode>.Empty, ImmutableArray<Diagnostic>.Empty);
+
+    public bool Equals(MapperResults other)
+    {
+        return Mappers.SequenceEqual(other.Mappers) && Diagnostics.SequenceEqual(other.Diagnostics);
```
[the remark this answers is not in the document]
TimothyMakkison: `MapperResult` is a value type therfore `ReferenceEquals` doesn't work. i can make it a class

test/Riok.Mapperly.Tests/Generator/IncrementalGeneratorTest.cs
TimothyMakkison: `WithTrackingName` is not available in older versions of roslyn. I could use the auto generated name or wrap everything in `#if` preproccessors

TimothyMakkison: No idea how to use `WithTrackingName` with older roslyn versions

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -35,13 +35,13 @@ WellKnownTypes wellKnownTypes
...
             _mapperDescriptor,
-            sourceContext,
+            _diagnostics,
             new MappingBuilder(_mappings),
             new ExistingTargetMappingBuilder(_mappings)
         );
     }
 
-    public MapperDescriptor Build()
+    public (MapperDescriptor descriptor, List<Diagnostic> errors) Build()
```
latonz, line 44: IMO the name errors is not appropriate here, as it can contain any diagnostics.

src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -22,39 +22,66 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
         var mapperClassDeclarations = SyntaxProvider.GetClassDeclarations(context);
 
         var compilationAndMappers = context.CompilationProvider.Combine(mapperClassDeclarations.Collect());
-        context.RegisterImplementationSourceOutput(compilationAndMappers, static (spc, source) => Execute(source.Left, source.Right, spc));
+        var mappersWithDiagnostics = compilationAndMappers.Select(
+            static (x, cancellationToken) => BuildDescriptors(x.Left, x.Right, cancellationToken)
+        );
+
+        // output the diagnostics
+        context.ReportDiagnostics(mappersWithDiagnostics.Select(static (source, _) => source.Diagnostics).WithTrackingName("diagnostics"));
```
latonz, line 30: Since `WithTrackingName` is not available in older Roslyn versions we could probably just add an extension method `IncrementalValueProvider<TSource> WithTrackingName<TSource>(this IncrementalValueProvider<TSource> source, string name) => this;` to mock this for older roslyn versions.

src/Riok.Mapperly/MapperNode.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public bool Equals(MapperNode other) => Body.IsEquivalentTo(other.Body) && FileName == other.FileName;
+
+    public override bool Equals(object? obj) => obj is MapperNode other && Equals(other);
+
+    public override int GetHashCode()
+    {
+        unchecked
+        {
+            return (Body.GetHashCode() * 397) ^ FileName.GetHashCode();
```
latonz, line 24: Couldn't this return another hash code for entities which are actually equal? (If the `Body` is another instance which is equal to each other?)

src/Riok.Mapperly/MapperResults.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+
+namespace Riok.Mapperly;
+
+public readonly struct MapperResults : IEquatable<MapperResults>
+{
+    public static readonly MapperResults Empty = new(ImmutableArray<MapperNode>.Empty, ImmutableArray<Diagnostic>.Empty);
+
+    public MapperResults(ImmutableArray<MapperNode> mappers, ImmutableArray<Diagnostic> diagnostics)
```
latonz, line 10: Couldn't we simply use `IReadOnlyCollection` instead?

src/Riok.Mapperly/MapperResults.cs
```diff
@@ -0,0 +1,50 @@
...
+                hash = Combine(hash, mapper.GetHashCode());
+            }
+            foreach (var diagnostic in Diagnostics)
+            {
+                hash = Combine(hash, diagnostic.GetHashCode());
+            }
+            return hash;
+        }
+
+        static int Combine(int h1, int h2)
```
latonz, line 42: We could probably backport the [HashCode](https://learn.microsoft.com/en-us/dotnet/api/system.hashcode) struct?

src/Riok.Mapperly/MapperResults.cs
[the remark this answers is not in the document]
latonz: My bad, thank you the explanation. IMO struct is a good fit here.

test/Riok.Mapperly.Tests/Generator/IncrementalGeneratorTest.cs
```diff
@@ -0,0 +1,121 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Riok.Mapperly.Tests.Generator;
+
+[UsesVerify]
+public class IncrementalGeneratorTest
+{
+    private const string AddMappersStep = "ImplementationSourceOutput";
+    private const string ReportDiagnosticsStep = "diagnostics";
```
latonz, line 11: I think it could be helpful to define these names as consts inside `Riok.Mapperly` and reference them from the MapperGenerator and here from the tests.

test/Riok.Mapperly.Tests/Generator/IncrementalGeneratorTest.cs
```diff
@@ -0,0 +1,121 @@
...
+
+namespace Test.B
+{
+    [Mapper]
+    internal partial class BarFooMapper
+    {
+        internal partial string BarToFoo(string value);
+    }
+}
+"""
```
latonz, line 51: Indent as other code...

latonz (changes requested):

test/Riok.Mapperly.Tests/Generator/IncrementalGeneratorTest.cs
latonz: You could probably just add an extension method `IncrementalValueProvider<TSource> WithTrackingName<TSource>(this IncrementalValueProvider<TSource> source, string name) => source;` to mock this for older roslyn versions.

TimothyMakkison: My main issue is getting the tests to pass. I can use preprocessors/mock `WithTrackingName`, but I can't get the tests to pass.

For instance if I wrap the tests in Net_7_And_Above the tests will still run for a version of Roslyn that does not add the tracking name

latonz: Hmmm, but the unit tests only run for the latest Roslyn version supported by Mapperly, don't they?

TimothyMakkison: That's what I thought too 😆 

I'll double check I probably made a nistake

src/Riok.Mapperly/MapperNode.cs
TimothyMakkison: Good question, it might.

I don't think incremental generators use the objects hash code or `SyntaxNode` has some custom `GetHashCode` method. Otherwise the tests would never pass and it wouldn't be possible to cache `SyntaxNode`, defeating the purpose of the incremental source generator.

TimothyMakkison:
Added `ImmutableEquatableArray` from `System.Text.Json`, its not needed but its more idiomatic IMO, happy to remove it.

src/Riok.Mapperly/MapperNode.cs
latonz: IMO not using the hash code is an implementation detail of how Roslyn calls the source generators which Mapperly should not rely on. IMO the fix here is also pretty simple:
* Instead of using the `Body.GetHashCode()` use a simpler hash code, eg. `Body.Members.Count.GetHashCode()`
* Remove the hash code part of the body completely

latonz (changes requested):

latonz (approved):

PR #530, opened 2023-06-29
chore: cache members, mappable members, get attribute and optimize `CollectionInfoBu…

TimothyMakkison:
### Description
Modified `WellKnownTypes` to cache more members, mappable members and attributes. Added a couple of checks to `CollectionInfoBuilder` to speed up common paths.

- Added cached methods to `WellKnownTypes` for `GetAttributes`, `GetAllMembers` and `GetAllAccessibleMappableMembers`
- CollectionTypeInfo
  - Converted `CollectionTypeInfo` to a struct
  - Removed duplicate `IImmutableQueue` and added `ImmutableSortedDictionary`
  - Added check to `HasValidAddMethod` for types with known `Add` methods, otherwise all none `None` types return false. `None` types are manually checked for `Add` methods.
  - `IsCountKnown` returns false for `IEnumerable` and true for all none `None` types. `None` types are manually checked for length properties.
  - Added a massive switch expression to `GetImplementedCollectionTypes`, it uses the known `CollectionTypeInfo` to get the implemented types. `None` types are manually checked for implemented types. Could be moved into `_collectionTypeInfos`.
    - Added a special check for `string` to return `IEnumerable`, because string isn't actually a "collection" type but does implement `IEnumerable` and is marked `None` by `GetCollectionTypeInfo`. It would iterate thrrought every possible collection type causing a considerable slow down.
- Moved `GetEnumeratedType` into `BuildCollectionInfos`. This is a quick check to ensure that both the source and target types are actually enumerable, saving a lot of wasted time.

I kept the slightly modified `GetAllMembers`and `GetAccessibleMappableMembers` (they call `WellKnownTypes`) in `SymbolExtensions`. Should I remove them and update all the call sites to use the `WellKnownTypes` version?

#### Optimized
|       Method |      Mean |     Error |    StdDev |      Gen0 |     Gen1 |  Allocated |
|------------- |----------:|----------:|----------:|----------:|---------:|-----------:|
|      Compile |  1.699 ms | 0.0322 ms | 0.0317 ms |   85.9375 |  19.5313 |  180.45 KB |
| LargeCompile | 54.299 ms | 0.9632 ms | 0.9460 ms | 1500.0000 | 375.0000 | 9344.24 KB |

#### Original
|       Method |       Mean |     Error |     StdDev |      Gen0 |     Gen1 |   Allocated |
|------------- |-----------:|----------:|-----------:|----------:|---------:|------------:|
|      Compile |   3.569 ms | 0.2264 ms |  0.6641 ms |  171.8750 |  31.2500 |   311.38 KB |
| LargeCompile | 100.170 ms | 4.6670 ms | 13.3905 ms | 2200.0000 | 800.0000 | 13395.01 KB |

Helps #72

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
@@ -13,6 +13,7 @@ public class DescriptorBuilder
 {
     private readonly MapperDescriptor _mapperDescriptor;
 
+    private readonly WellKnownTypes _wellKnownTypes;
     private readonly MappingCollection _mappings = new();
     private readonly MethodNameBuilder _methodNameBuilder = new();
     private readonly MappingBodyBuilder _mappingBodyBuilder;
@@ -28,6 +29,7 @@ public class DescriptorBuilder
         WellKnownTypes wellKnownTypes
     )
     {
+        _wellKnownTypes = wellKnownTypes;
         _mapperDescriptor = new MapperDescriptor(mapperSyntax, mapperSymbol, _methodNameBuilder);
         _mappingBodyBuilder = new MappingBodyBuilder(_mappings);
         _builderContext = new SimpleMappingBuilderContext(
@@ -77,7 +79,7 @@ public class DescriptorBuilder
 
     private void ReserveMethodNames()
     {
-        foreach (var methodSymbol in _mapperDescriptor.Symbol.GetAllMembers())
+        foreach (var methodSymbol in _mapperDescriptor.Symbol.GetAllMembers(_wellKnownTypes))
         {
             _methodNameBuilder.Reserve(methodSymbol.Name);
         }
--- a/src/Riok.Mapperly/Descriptors/Enumerables/CollectionInfoBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/Enumerables/CollectionInfoBuilder.cs
@@ -6,7 +6,7 @@ namespace Riok.Mapperly.Descriptors.Enumerables;
 
 public static class CollectionInfoBuilder
 {
-    private record CollectionTypeInfo(
+    private readonly record struct CollectionTypeInfo(
         CollectionType CollectionType,
         Type? ReflectionType = null,
         string? TypeFullName = null,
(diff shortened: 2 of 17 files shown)


latonz:
@TimothyMakkison thank you for this PR. Could you split it up into several smaller PR's to make them easier to review? This also makes sure you get prompter feedback and keeps the enhancements isolated. Additionally the performance impact of each change can be measured more easily. Proposed PR's:
* Caching of members
* CollectionTypeInfo general optimisations (struct, Immutable duplicate / sorted dict, GetEnumeratedType in BuildCollectionInfos)
* CollectionTypeInfo GetImplementedCollectionTypes

I'd split GetImplementedCollectionTypes and the general CollectionTypeInfo optimisations as I'm not sure if the massive type switch is worth it, would be nice if we could measure the performance impact on that separately and therefore create an isolated PR.

On a first glance the enhancements look promising 😊  I think I'd remove the member methods from `SymbolExtensions` and move them to the well known types class. Not sure if well known types is still an appropriate name then or if it should be renamed to just `Types` or similar.

TimothyMakkison:
Good idea, I'll split this pr.

> I'm not sure if the massive type switch is worth it, 

It was probably the most impactful change. Iirc Mapperly spent 30% of the time calling GetImplementedCollections. It would also get the type symbol for every possible collection, degrading most use cases where they only simple collections.

I can remove the switch and move it into CollectionInfos as another property

latonz:
@TimothyMakkison oh okay, I didn't expect that. I think I'd keep the current switch as it is easier to read 🤔
Doesn't this deactivate the feature detection? Eg. `IReadOnlySet` is only available for .NET 5+ and not for .NET Standard 2.0, this was detected before, but isn't in the new code. Not sure if this is an actual problem for Mapperly...

TimothyMakkison:
> @TimothyMakkison oh okay, I didn't expect that.

Yeah, I was a little surprised.

> Doesn't this deactivate the feature detection? Eg. `IReadOnlySet` is only available for .NET 5+ and not for .NET Standard 2.0, this was detected before, but isn't in the new code. Not sure if this is an actual problem for Mapperly...

It could be an issue if Mapperly added a non backward compatible mapper for types implementing `IReadOnlySet`. I'm not too concerned by this, ImplementedTypes extracts a bunch of unused information

Issue #453, opened 2023-05-22
Include all members from a nested member


latonz:
If a class has a nested member which is flattened without the member name prefix, each member needs a manual `MapProperty` attribute to be mapped correctly. This could be simplified a lot by including all members of a nested property by one configuration attribute.
Name proposal: `MapNestedPropertiesAttribute(string memberName)`
(probably `MapNestedMembersAttribute` would be better as this includes fields and does not only apply to properties, but for backward compatibility `MapProperty` is called what it is, therefore the name should match).

Example:
```
class Document { string Name; DocumentSettings Settings;}
class DocumentSettings { string Locale; string CurrencySymbol; }
class DocumentDto { string Name; string Locale; string CurrencySymbol; }

[Mapper]
partial class Mapper
{
    [MapNestedProperties(nameof(Document.Settings))]
    public partial DocumentDto Map(Document doc);
    
    // instead of:
    [MapProperty("Settings.Locale", "Locale")]
    [MapProperty("Settings.CurrencySymbol", "CurrencySymbol")]
    public partial DocumentDto Map(Document doc);
}
```

AutoMapper supports this via [IncludeMembers](https://docs.automapper.org/en/stable/Flattening.html#includemembers).
See https://github.com/riok/mapperly/discussions/447#discussioncomment-5952448.

TimothyMakkison:
How should it be used for other mappings? ie should it map `Document.Settings.Locale` -> `DocumentDto.SettingsLocale`

latonz:
@TimothyMakkison good question, didn't thought about this. I think this should diagnostic unless explicitly mapped.

TimothyMakkison:
Should it support nested members? Something like:

```C#
[Mapper]
partial class Mapper
{
    [MapNestedProperties("Settings.Region"))]
    public partial DocumentDto Map(Document doc);
    
    // instead of:
    [MapProperty("Settings.Locale", "Locale")]
    [MapProperty("Settings.CurrencySymbol", "CurrencySymbol")]
    public partial DocumentDto Map(Document doc);
}
```

latonz:
@TimothyMakkison intuitively I'd say it should support nested members.

trejjam:
I am interested in this feature. Can I help?

mindotale:
Hi, what is the progress on this issue? This is a really useful and necessary feature.

PR #534, opened 2023-07-03
feat: add support for "." wildcard

zeidoo:
# Adding support for "." wildcard

## Description
If you don’t want explicitly name all properties from nested source bean, you can use . as target. This will tell Mapperly to map every property from source bean to target object.

Inspired by:
https://mapstruct.org/documentation/stable/reference/html/#mapping-nested-bean-properties-to-current-target

Currently a WIP. Please let me know what to add for this to be merge eventually.

## Checklist


zeidoo:
Please also note I added a test that verifies that a user defined [MapProperty] takes precedence over the PropertyNameMappingStrategy.

latonz:
@zeidoo thank you for opening this PR. Doesn't this address the same use-case as we planned to address with https://github.com/riok/mapperly/issues/453?

zeidoo:
I guess it does. Didn't see that issue.

I would say that mapperly is closer to mapstruct than it is to automapper in its usage. I would favor mapstruct's "." syntax instead of a dedicated attribute. 

Ultimately, both work. Let me know which you prefer and I can change the PR if needed.

Ps: if we follow mapstruct's way, we get free documentation ;)

latonz:
In my opinion, `MapNestedProperties` describes better what happens. It is clearer to a reader of the code what is being configured. I don't think documentation will be a problem, as Mapperly uses its own documentation anyway, which should always be kept up to date and also be the place to go for users. Further, I think more users of Mapperly are familiar with Automapper than with MapStruct. Therefore I prefer the proposed `MapNestedProperties` API design.

wrestlerdude:
+1 

Desperately need this. In my view mapping the properties for like flattening and unflattening with attributes not only defeats the purpose of a mapper in the first place - it is less ergonomic than mapping them manually.

PR #540, opened 2023-07-05
feat: add support for roslyn 3.9

recih:
# add support for roslyn 3.9

## Description

With lower version of roslyn support, mapperly now compatible with Unity 2021.

* add polyfills for `IsPartialDefinition`, `IsRequired`
* use `ISourceGenerator` instead of `IIncrementalGenerator`, later one is not exist in roslyn 3.9
* add `ISourceProductionContext`
* make `ROSLYNx_y_OR_GREATER` defines more reasonable

Fixes # (issue)

## Checklist


src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -57,4 +76,32 @@ private static void Execute(Compilation compilation, ImmutableArray<ClassDeclara
             );
         }
     }
+
+#if ROSLYN4_0_OR_GREATER
+    class SourceProductionContextProxy : ISourceProductionContext
+    {
+        private SourceProductionContext _ctx;
```
latonz, line 83: This field could be set to readonly.

src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -57,4 +76,32 @@ private static void Execute(Compilation compilation, ImmutableArray<ClassDeclara
...
+
+        public void AddSource(string hintName, string source) => _ctx.AddSource(hintName, source);
+
+        public void ReportDiagnostic(Diagnostic diagnostic) => _ctx.ReportDiagnostic(diagnostic);
+    }
+#endif
+
+    class GeneratorExecutionContextProxy : ISourceProductionContext
+    {
+        private GeneratorExecutionContext _ctx;
```
latonz, line 97: This field could be set to readonly.

src/Riok.Mapperly/SymbolExtensions.cs
```diff
@@ -0,0 +1,62 @@
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.CodeAnalysis;
+
+public static class SymbolExtensions
```
latonz, line 6: There is a `SymbolExtensions` class in the `Helpers` directory, I think this could be included there.

src/Riok.Mapperly/SymbolExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    public static bool IsPartialDefinition(this IMethodSymbol method)
+    {
+#if ROSLYN3_10_OR_GREATER
+        return method.IsPartialDefinition;
+#else
+        if (method.Locations.Any(loc => loc.IsInMetadata))
+        {
+            return false;
+        }
```
latonz, line 16: This code base usually omits braces for single line if conditions with single line returns.

src/Riok.Mapperly/SymbolExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        var isPartial = method.DeclaringSyntaxReferences
+            .Select(r => r.GetSyntax())
+            .Where(s => s.IsKind(SyntaxKind.MethodDeclaration))
+            .Cast<MethodDeclarationSyntax>()
+            .Any(s => s.Modifiers.Any(SyntaxKind.PartialKeyword));
+        return isPartial && method.PartialImplementationPart == null && method.PartialDefinitionPart == null;
```
latonz, line 23: Why does the definition part need to be null? Shouldn't it be not null?

src/Riok.Mapperly/SymbolExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+        if (property.Locations.Any(loc => loc.IsInMetadata))
+        {
+            return false;
+        }
+
+        return property.DeclaringSyntaxReferences
+            .Select(r => r.GetSyntax())
+            .Where(s => s.IsKind(SyntaxKind.PropertyDeclaration))
+            .Cast<PropertyDeclarationSyntax>()
+            .Any(s => s.Modifiers.Any(t => t.ToString() == "required"));
```
latonz, line 41: Extract a named constant for the string "required".

src/Riok.Mapperly/SymbolExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+#else
+        if (field.Locations.Any(loc => loc.IsInMetadata))
+        {
+            return false;
+        }
+
+        return field.DeclaringSyntaxReferences
+            .Select(r => r.GetSyntax())
+            .Where(s => s.IsKind(SyntaxKind.PropertyDeclaration))
+            .Cast<PropertyDeclarationSyntax>()
```
latonz, line 58: Shouldn't these be field declarations?

src/Riok.Mapperly/SyntaxProvider.Roslyn3.9.cs
```diff
@@ -0,0 +1,46 @@
+﻿#if ROSLYN3_9_OR_GREATER && !ROSLYN4_0_OR_GREATER
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Riok.Mapperly;
+
+class SyntaxProvider : ISyntaxContextReceiver
```
latonz, line 8: This code base usually always explicitly states the visibility.

src/Riok.Mapperly/SyntaxProvider.Roslyn3.9.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public List<ClassDeclarationSyntax> ClassDeclarations { get; } = new();
```
latonz, line 10: I think this can be converted to a private readonly field and an additionally `IReadOnlyCollection` can be exposed to encapsulate the modification of the collection.

src/Riok.Mapperly/SyntaxProvider.Roslyn3.9.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    public void OnVisitSyntaxNode(GeneratorSyntaxContext context)
+    {
+        if (IsSyntaxTargetForGeneration(context.Node))
```
latonz, line 14: To keep the nesting level low and improve readability an inverted if with an early return could help here.

latonz (changes requested):
Thank you for your contribution to Mapperly!
I added my feedback. Feel free to ask any questions that may arise or discuss the feedback provided.

One big thing that is needed for this to be ready to merge are integration tests. There are several locations of the code which diverge between roslyn 3.9 and later versions and roslyn 3.9 is currently untested.

benchmarks/Riok.Mapperly.Benchmarks/SourceGeneratorBenchmarks.cs
```diff
@@ -70,7 +70,7 @@ private async Task<(Compilation, CSharpGeneratorDriver)> SetupAsync(string proje
         if (compilation == null)
             throw new InvalidOperationException("Compilation returned null");
 
-        var generator = new MapperGenerator().AsSourceGenerator();
+        var generator = new MapperGenerator() as ISourceGenerator;
```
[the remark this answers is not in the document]
recih: This should be fixed later

src/Riok.Mapperly/ISourceProductionContext.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Riok.Mapperly;
+
+public interface ISourceProductionContext
+{
+    public void AddSource(string hintName, Microsoft.CodeAnalysis.Text.SourceText sourceText);
+    public void AddSource(string hintName, string source);
```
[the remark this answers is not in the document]
recih: these three methods are totally comes from `SourceProductionContext`, maybe we should keep them

src/Riok.Mapperly/SymbolExtensions.cs
recih: This is the wierd part, see: https://github.com/dotnet/roslyn/issues/51082
Will add more info in comment

benchmarks/Riok.Mapperly.Benchmarks/SourceGeneratorBenchmarks.cs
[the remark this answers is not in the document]
latonz: Oh sorry if my review was too early. Please put the PR in draft state as long as it is not ready 😊 you can always request a review.

src/Riok.Mapperly/ISourceProductionContext.cs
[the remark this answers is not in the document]
latonz: Since this method is completely unused I‘d remove it for now. It can be readded if it is needed. Otherwise we need to maintain it although it is not used.

recih:
@latonz I need some help.
I have enabled unit testing for different versions of Roslyn (3.9, 4.0, 4.4, 4.5) and discovered that even for Roslyn 4.4, there are some inconsistencies in the generated code.

received:
```c#
    private partial global::System.Linq.IQueryable<global::B> Map(global::System.Linq.IQueryable<global::A> source)
    {
#nullable disable
        return System.Linq.Queryable.Select(source, x => new global::B()
        {StringValue = x.StringValue, IntValue = x.IntValue, CharValue = x.CharValue});
#nullable enable
    }
```

verified:
```c#
    private partial global::System.Linq.IQueryable<global::B> Map(global::System.Linq.IQueryable<global::A> source)
    {
#nullable disable
        return System.Linq.Queryable.Select(source, x => new global::B() { StringValue = x.StringValue, IntValue = x.IntValue, CharValue = x.CharValue });
#nullable enable
    }
```

Any advice for fixing these?

latonz:
@recih Mapperly runs the unit tests with the latest supported Roslyn version, builds the multi-targeted nuget package and runs the integration tests for all supported Roslyn versions with the built nuget package. The idea is to test different Roslyn versions with the integration tests. AFAIK testing the unit tests with several Roslyn versions is not easily possible, as the compiler version of the environment is used, even though Mapperly may reference another Roslyn version.
The difference you found is a formatting (`NormalizeWhitespace`) issue which was present in Roslyn but has been fixed or was at least improved in recent versions of Roslyn. That's why you see the difference in the generated code.
You may need to adjust the `integration-test` github action job to get the integration tests running for a target framework version which uses Roslyn 3.9.
Does this clarify your question?

latonz:
This may bring up merge conflicts with https://github.com/riok/mapperly/pull/524

recih:
@latonz Sorry for the late reply.

> AFAIK testing the unit tests with several Roslyn versions is not easily possible

Yes, it's a bit challenging, but I've managed to get all the unit tests to pass, although I used some tricky methods. I think it's necessary to conduct unit tests on different versions of Roslyn. However, it will indeed increase some maintenance costs.

> The difference you found is a formatting (`NormalizeWhitespace`) issue which was present in Roslyn but has been fixed or was at least improved in recent versions of Roslyn. That's why you see the difference in the generated code.

I found a way to compare code with `SyntaxTree.IsEquivalentTo()`, which will ignore trivial whitespaces. Also I used #if to disable some test code for the `require` keyword. 

Additionally, I patched some test codes so they can pass under Roslyn 3.9, because Roslyn 3.x lacks support for generic attributes and record structs. Here is the tricky part I mentioned earlier. There might be a better solution for this. We can discuss it.

Additionally, I added tests for .net 5.0 in the integration tests, and also disabled some tests related to IncrementalGenerator since Roslyn 3.x doesn't support it.

This PR is done and ready for review. I hope it can be merged soon, because you guys code so fast. I've had to rebase and resolve conflicts at least 3 or 4 times during this period😂.

src/Riok.Mapperly/Riok.Mapperly.Roslyn4.0.props
```diff
@@ -1,5 +1,9 @@
 <Project>
 
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);ROSLYN3_9_OR_GREATER;ROSLYN3_10_OR_GREATER;ROSLYN4_0_OR_GREATER</DefineConstants>
```
latonz, line 4: Here we define a `ROSLYN3_10_OR_GREATER`, however, Mapperly never references an API version of `3.10` of Roslyn. Therefore I would not define this constant. Instead, I think it would make sense to define constants without the `_OR_GREATER` suffix (eg. `ROSLYN_4_0`), this would allow to simplify some conditions.

test/Riok.Mapperly.Tests/MapperGenerationResultAssertions.cs
```diff
@@ -114,7 +120,11 @@ public MapperGenerationResultAssertions HaveOnlyMethods(params string[] methodNa
 
     public MapperGenerationResultAssertions HaveMethodBody(string methodName, [StringSyntax(StringSyntax.CSharp)] string mapperMethodBody)
     {
-        _mapper.Methods[methodName].Body.Should().Be(mapperMethodBody.ReplaceLineEndings().Trim(), $"Method: {methodName}");
+        var actual = CSharpSyntaxTree.ParseText(_mapper.Methods[methodName].Body);
+        var expected = CSharpSyntaxTree.ParseText(mapperMethodBody);
+        Execute.Assertion
+            .ForCondition(actual.IsEquivalentTo(expected))
+            .FailWith($"method body was expected to be equal to `{expected}`, but was `{actual}` instead");
```
latonz, line 127: Extract into a method `void AssertCSharp([StringSyntax(StringSyntax.CSharp)] string expected, [StringSyntax(StringSyntax.CSharp)] string actual)`

test/Riok.Mapperly.Tests/Mapping/IgnoreObsoleteTest.cs
```diff
@@ -345,6 +345,7 @@ class B
             );
     }
 
+#if ROSLYN4_4_OR_GREATER
```
latonz, line 348: IsRequired is polyfilled for `Roslyn3.9`, why doesn't this test work? Does the polyfill work correctly?

latonz (changes requested):
Thank you for the updates.

I'm not really happy running the unit tests on all supported Roslyn versions as IMO this increases the maintenance costs significantly. We wan't to keep the maintenance cost for older versions of Roslyn as small as possible. Especially `PatchForRecordStruct` and similar methods seem super-hacky and magic to me. Older Roslyn versions are fully tested by the integration tests, aren't they? IMO the integration tests should be sufficient to test older versions, otherwise we should probably discuss on how the integration tests could be improved.

latonz:
Closed for now, as this PR has been inactive for almost two months. Feel free to reopen it if you have a reviewable version.

PR #542, opened 2023-07-05
chore: optimize `GetImplementedCollectionTypes`

TimothyMakkison:
# Optimized `GetImplementedCollectionTypes`
## Description
-Added a massive switch expression to GetImplementedCollectionTypes, it uses the known CollectionTypeInfo to get the implemented types. None types are manually checked for implemented types. Could be moved into _collectionTypeInfos.
-Added a special check for string to return IEnumerable, because string isn't actually a "collection" type but does implement IEnumerable and is marked None by GetCollectionTypeInfo. It would iterate thrrought every possible collection type causing a considerable slow down.

Splitting up changes from #530

## Checklist


src/Riok.Mapperly/Descriptors/Enumerables/CollectionInfoBuilder.cs
```diff
@@ -188,21 +188,147 @@ or CollectionType.ReadOnlyMemory
...
+            CollectionType.Span => CollectionType.Span,
+            CollectionType.ReadOnlySpan => CollectionType.ReadOnlySpan,
+            CollectionType.Memory => CollectionType.Memory,
+            CollectionType.ReadOnlyMemory => CollectionType.ReadOnlyMemory,
+
+            CollectionType.None when SymbolEqualityComparer.Default.Equals(type, types.Get<string>()) => CollectionType.IEnumerable,
+            _ => IterateHierarchy(type, types)
+        };
+
+        static CollectionType IterateHierarchy(ITypeSymbol type, WellKnownTypes types)
```
latonz, line 316: This code base usually explicitly states the visibility. I'd set this to private.

latonz, line 316: IMO `IterateHierarchy` is not an optimal name, I'd probably rename to `GetImplementedCollectionTypesByIteratingHierarchy` or similar.

latonz (changes requested):

src/Riok.Mapperly/Descriptors/Enumerables/CollectionInfoBuilder.cs
```diff
@@ -188,21 +188,147 @@ or CollectionType.ReadOnlyMemory
         return null;
     }
 
-    private static CollectionType GetImplementedCollectionTypes(WellKnownTypes types, ITypeSymbol type)
+    private static CollectionType GetImplementedCollectionTypes(WellKnownTypes types, ITypeSymbol type, CollectionType collectionType)
     {
-        var implementedCollectionTypes = type.IsArrayType() ? CollectionType.Array : CollectionType.None;
-
-        foreach (var typeInfo in _collectionTypeInfos)
+        return collectionType switch
```
latonz, line 193: Could you add a comment that the dynamic fallback `IterateHierarchy` is only used, if the type is not well-known and that this is done for performance reasons.

TimothyMakkison: 👍  Changed name to `IterateImplementedTypes` and added comments

latonz (approved):

PR #544, opened 2023-07-06
fix: prevent duplicate attributes and base types for nested mappers

### LINKED ISSUE #543: Compilation fails when defining mapper as inner/nested class and the outer class has an attribute
**Describe the bug**
When defining the mapper as an inner class, the generated mapper code is copying the existing class attributes and the parent class but not copying their respective namespace defined at the bottom. This is creating 2 problems:
1. The compilation fails because the attribute or the parent are not found (since it missing the using namespace). There is an open issue for this already (https://github.com/riok/mapperly/issues/507)
2. By copying the attribute/tag to the generated class, it generates a compilation error `CS0579 Duplicate 'xyz' attribute` since now both the original and the generated class are applying the same attribute/tag.

**To Reproduce**
1. Create a class with an attribute
2. Declare a mapper as an inner class
4. Try to compile the code.

**Expected behavior**
Code compiles without error

**Code snippets**
```csharp
using System.Diagnostics.CodeAnalysis;
namespace Bug_Mapperly;

internal class Program { static void Main(string[] args) { Console.WriteLine("Hello, World!"); } }

[ExcludeFromCodeCoverage] // --The generated code will also have this attribute set and will show the 2 compilations errors mentioned above (Undefined ExcludeFromCodeCoverage because of lack of namespace import. If you use the fqn for ExcludeFromCodeCoverage as in the line below, you will get the duplicate attribute error
//[System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
public partial class CarDto
{
    public string? Maker { get; set; }
    public string? Model { get; set; }
    [Mapper]
    private partial class CarMapper
    {
        public partial CarDto From(Car car);
    }
}
public class Car
{
    public string? Maker { get; set; }
    public string? Model { get; set; }
}
```
**Fix suggestion**
- Mapper should copy the `using namespace` code
- Mapper should not copy outer class attributes when generating the code

- Mapperly Version: 2.9.0-next.2
- .NET Version: .NET 6
- Target Framework: [e.g. .net6.0]
- IDE:  Visual Studio Pro 2022
- OS: Windows 10

**Additional context**
**Related issues:**
- Added support to inner classes: https://github.com/riok/mapperly/issues/350
- Lack of namespace import: https://github.com/riok/mapperly/issues/507

TimothyMakkison:
# Prevent duplicate attributes and base types appearing for nested mappers

## Description
If a class containing a mapper has an attribute, mapperly will create a copy of the attribute in the generated code, not respecting imports or the allow multiple clause, causing compilation issues. This will also occur if the parent class inherit from a base type.

I've changed `WrapInClassesAsNeeded` in `SourceEmitter` to ignore the attributes and base list from the parent class. 
I'm not sure if this is the right fix, perhaps outright creating a new token with the same accessibility tokens would be a better idea.

Fixes #543, #548

## Checklist


latonz (approved):
thank you for this PR!

PR #537, opened 2023-07-04
chore: add member, mappable member and attribute caching

TimothyMakkison:
# Add member, mappable member and attribute caching

## Description
Added member, mappable member and attribute caching to improve performance and reduce memory usage.

Splitting up changes from #530 

## Benchmarks
10ms time and 3000KB memory savings
### Cache
|       Method |      Mean |     Error |    StdDev |      Gen0 |     Gen1 |  Allocated |
|------------- |----------:|----------:|----------:|----------:|---------:|-----------:|
|      Compile |  2.491 ms | 0.0391 ms | 0.0366 ms |  156.2500 |  31.2500 |  284.15 KB |
| LargeCompile | 64.276 ms | 1.2297 ms | 1.2628 ms | 1714.2857 | 714.2857 | 10713.9 KB |

### Original
|       Method |      Mean |     Error |    StdDev |      Gen0 |     Gen1 |   Allocated |
|------------- |----------:|----------:|----------:|----------:|---------:|------------:|
|      Compile |  2.640 ms | 0.0524 ms | 0.0490 ms |  179.6875 |  46.8750 |   320.39 KB |
| LargeCompile | 75.350 ms | 1.4524 ms | 1.3586 ms | 2333.3333 | 833.3333 | 14283.01 KB |
|       Method |      Mean |     Error |    StdDev |      Gen0 |     Gen1 |   Allocated |
|------------- |----------:|----------:|----------:|----------:|---------:|------------:|
|      Compile |  2.627 ms | 0.0428 ms | 0.0379 ms |  179.6875 |  39.0625 |   326.46 KB |
| LargeCompile | 72.979 ms | 1.1195 ms | 0.8740 ms | 2285.7143 | 571.4286 | 14280.24 KB |

## Checklist

- [x] The existing code style is followed
- [x] The commit message follows our guidelines
- [x] Performed a self-review of my code
- [x] Hard-to-understand areas of my code are commented
- [x] The documentation is updated (as applicable)
- [x] Unit tests are added/updated
- [x] Integration tests are added/updated (as applicable, especially if feature/bug depends on roslyn or framework version in use)

Issue #507, opened 2023-06-19
Generate code including using statements


mci-kmd:
**Is your feature request related to a problem? Please describe.**
The generated code does not import any namespaces. This means that every reference to a type is fully qualified. This greatly reduces the readability of the code, especially when you use long namespaces.

**Describe the solution you'd like**
It would be greate if all the relevant using statements could be added to ensure that types can be referenced without namespace qualification. I realize that the current solution sidesteps the issue of class ambiguity, for example when mapping between types of the same name. This could be solved with an aliasing convention, though I'm not sure what the best one would be. A better approach would probably be just to fall back to the old approach for these cases.

latonz:
Thank you for opening this issue. We'll keep the issue around but currently other features have higher priority.
If you are interested in implementing this, free to contribute a concept on how this could be implemented to this issue.

PR #549, opened 2023-07-08
Use `string` for methods instead of `IMethodSymbol`

TimothyMakkison:
# Use `string` for methods instead of `IMethodSymbol`

## Description
Inspired [by Sytem.Text.Json](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/gen/Helpers/KnownTypeSymbols.cs#L237C13-L237C13) to use the known string of a method instead of looking it up via the compilation.

- `CollectionInfoBuilder` - used `SpecialType.System_Int32` instead of `types.Get<int>()`
- `EnumerableMappingBuilder` - use method string instead of retrieving the type and looking for a matching method, this has been done for ImmutableCollections, Collect methods and `Select`
- `DictionaryMappingBuilder`
  - Use switch expression in `ResolveImmutableCollectMethod` to use `CollectionType`
  - Use method string instead of retrieving the type and looking for a matching method.
  - Fix typo `LinqDicitonaryMapping`

Related to #530

Note that I've used a branch where #537 and #542 are added. This shouldn't be merged until after they are added.
By using a known method string it might be possible to remove the qualified type signature and use a using statement. See #507 

## Benchmarks
Saves 0.5MB with a small 5% speed boost
### Original
|       Method |      Mean |     Error |    StdDev |      Gen0 |     Gen1 |  Allocated |
|------------- |----------:|----------:|----------:|----------:|---------:|-----------:|
|      Compile |  1.653 ms | 0.0307 ms | 0.0301 ms |   93.7500 |  23.4375 |   180.8 KB |
| LargeCompile | 58.961 ms | 0.9728 ms | 1.6253 ms | 1571.4286 | 571.4286 | 9712.91 KB |

### Changed
|       Method |      Mean |     Error |    StdDev |      Gen0 |     Gen1 |  Allocated |
|------------- |----------:|----------:|----------:|----------:|---------:|-----------:|
|      Compile |  1.618 ms | 0.0317 ms | 0.0412 ms |   93.7500 |  23.4375 |  174.31 KB |
| LargeCompile | 55.762 ms | 0.8799 ms | 1.2620 ms | 1500.0000 | 500.0000 | 9284.46 KB |

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
+++ b/src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
@@ -299,21 +299,26 @@ public static class SyntaxFactoryHelper
         return InvocationExpression(methodAccess).WithArgumentList(ArgumentList(arguments));
     }
 
-    public static InvocationExpressionSyntax StaticInvocation(IMethodSymbol method, params ExpressionSyntax[] arguments) =>
-        StaticInvocation(
-            FullyQualifiedIdentifierName(method.ReceiverType?.NonNullable()!)
-                ?? throw new ArgumentNullException(nameof(method.ReceiverType)),
-            method.Name,
-            arguments
-        );
+    public static string StaticMethodString(IMethodSymbol method)
+    {
+        var receiver = method.ReceiverType ?? throw new NullReferenceException(nameof(method.ReceiverType) + " is null");
+        var qualifiedReceiverName = FullyQualifiedIdentifierName(receiver.NonNullable());
+        return $"{qualifiedReceiverName}.{method.Name}";
+    }
+
+    public static InvocationExpressionSyntax StaticInvocation(IMethodSymbol method, params ExpressionSyntax[] arguments)
+    {
+        var receiver = method.ReceiverType ?? throw new NullReferenceException(nameof(method.ReceiverType) + " is null");
+        var qualifiedReceiverName = FullyQualifiedIdentifierName(receiver.NonNullable());
+        return StaticInvocation(qualifiedReceiverName, method.Name, arguments);
+    }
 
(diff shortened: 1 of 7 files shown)


src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
```diff
@@ -299,6 +299,14 @@ public static InvocationExpressionSyntax StaticInvocation(string receiverType, s
         return InvocationExpression(methodAccess).WithArgumentList(ArgumentList(arguments));
     }
 
+    public static string StaticMethodString(IMethodSymbol method)
+    {
+        var qualified =
+            FullyQualifiedIdentifierName(method.ReceiverType?.NonNullable()!)
+            ?? throw new ArgumentNullException(nameof(method.ReceiverType));
```
latonz, line 306: I don't think this works as expected, it looks like we have the same problem at other locations in this file.
As far as I understand it what tried achieving here would be sth. like this:
```c#
var receiver = method.ReceiverType ?? throw new NullReferenceException(nameof(method.ReceiverType) + " is null");
var qualifiedReceiverName = FullyQualifiedIdentifierName(receiver.NonNullable());
```

latonz (changes requested):

src/Riok.Mapperly/Emit/SyntaxFactoryHelper.cs
TimothyMakkison: Updated for `StaticInvocation` and `StaticMethodString` methods

latonz (approved):
Looks good to me 😊 Ready to merge after a rebase.

PR #551, opened 2023-07-09
fix: add modifying nested struct diagnostic

### LINKED ISSUE #516: `MapProperty` modifies temporary struct members
**Describe the bug**
`MapProperty` will try to modify struct members, this is invalid because a temporary object is returned and modified, not altering the original member.

**To Reproduce**
```C#
public struct MyStruct
{
}
public class Destination
{
    public MyStruct Nested { get; set; }
}
public record Source(int Value);

[Mapper]
{
    [MapProperty(nameof(Source.Value), "Nested.Value")]
    public static partial Destination Map(Source req);
}
```

```C#
public static partial global::Riok.Mapperly.Sample.Destination Map(global::Riok.Mapperly.Sample.Source req)
{
    var target = new global::Riok.Mapperly.Sample.Destination();
    // Property 'Nested' access returns temporary value. Cannot modify struct member when accessed struct is not classified as a variable
    target.Nested.Value = req.Value;
    return target;
}
```
**Expected behavior**
Determine if `MapProperty` is mutating a struct member. Note that if `MapProperty` is mutating something with even more nesting where a sub member is a class it will be valid again.
ie
```C#
[MapProperty(nameof(Source.Value), "Nested.MyClass.Value")] // valid if MyClass if a class
```

TimothyMakkison:
# Add diagnostic for modifying a temporary value type

## Description
Added check for if a nested value type is being modified.

Probably not a common issue so I went with a diagnostic.
I considered nested chain assignment or a `with` expression ie `target.NestedValue = target.NestedValue with { StringValue = src.StringValue };` (can repeatedly chain)

Fixes #516

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/ObjectMemberMappingBodyBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/ObjectMemberMappingBodyBuilder.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Configuration;
 using Riok.Mapperly.Descriptors.MappingBodyBuilders.BuilderContext;
@@ -95,6 +96,7 @@ public static class ObjectMemberMappingBodyBuilder
         BuildMemberAssignmentMapping(ctx, sourceMemberPath, targetMemberPath);
     }
 
+    [SuppressMessage(" Meziantou.Analyzer", "MA0051:MethodIsTooLong")]
     public static bool ValidateMappingSpecification(
         IMembersBuilderContext<IMapping> ctx,
         MemberPath sourceMemberPath,
@@ -131,6 +133,11 @@ public static class ObjectMemberMappingBodyBuilder
             );
             return false;
         }
+
+        // cannot assign to intermediate value type, error CS1612
+        // invalid mapping a value type has a property set
+        if (!ValidateStructModification(ctx, sourceMemberPath, targetMemberPath))
+            return false;
 
         // a target member path part is init only
         var noInitOnlyPath = allowInitOnlyMember ? targetMemberPath.ObjectPath : targetMemberPath.Path;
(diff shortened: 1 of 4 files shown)


src/Riok.Mapperly/Descriptors/MappingBodyBuilders/ObjectMemberMappingBodyBuilder.cs
```diff
@@ -135,6 +135,27 @@ PropertyMappingConfiguration config
             return false;
         }
 
+        // cannot assign to intermediate value type
+        if (targetMemberPath.Path.Count > 1)
+        {
+            var member = targetMemberPath.Path.Reverse().Skip(1).First();
```
latonz, line 141: This only checks the second last item, but couldn't be any in the path except for the last item be a problem?
Otherwise this could be simplified to `var member = targetMemberPath.Path[^2];` / `var member = targetMemberPath.Path[targetMemberPath.Path.Count - 2];` (Index doesn't work due to net standard 2.0).

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
```diff
@@ -443,4 +443,27 @@ public void ShouldIgnoreStaticConstructorAndDiagnostic()
...
     }
+
+    [Fact]
+    public void ModifyingTemporaryStructShouldDiagnostic()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+            [MapProperty("StringValue", "NestedValue.StringValue")]
+            partial B Map(A src);
+""",
```
latonz, line 454: Weird indention

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/ObjectMemberMappingBodyBuilder.cs
```diff
@@ -135,6 +135,27 @@ PropertyMappingConfiguration config
...
+                    ctx.Mapping.SourceType,
+                    sourceMemberPath.FullName,
+                    sourceMemberPath.Member.Type,
+                    ctx.Mapping.TargetType,
+                    targetMemberPath.FullName,
+                    targetMemberPath.Member.Type,
+                    member.Name,
+                    member.Type
+                );
+                return true;
```
latonz, line 155: As discussed in https://github.com/riok/mapperly/issues/516#issuecomment-1655677725 it would make sense to return false here and not generate this mapping at all, wouldn't it?

latonz (changes requested):

TimothyMakkison:
Sorry, I though I had fixed all of this a week ago. Turns out I forgot to push it 😅

latonz:
@TimothyMakkison ah yep on a second thought I think that makes sense, that only the second last is checked.

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -410,4 +410,13 @@ internal static class DiagnosticDescriptors
         DiagnosticSeverity.Error,
         true
     );
+
+    public static readonly DiagnosticDescriptor CannotMapToTemporarySourceMember = new DiagnosticDescriptor(
+        "RMG047",
+        "Cannot map to member path due to modifying a temporary value",
```
latonz, line 416: I think I'd reference CS1612...

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/ObjectMemberMappingBodyBuilder.cs
```diff
@@ -132,6 +132,28 @@ PropertyMappingConfiguration config
             return false;
         }
 
+        // cannot assign to intermediate value type
+        // invalid mapping when second to last member is a property that returns a value type
```
latonz, line 136: Probably reference CS1612 in the comment.

latonz (changes requested):

TimothyMakkison:
Added CS1612 to error message.

Modified the check to iterate backwards, checking for value type properties. Checking the penultimate member would work if field types didn't exist

latonz (approved):

TimothyMakkison:
Analyzer doesn't like the length of `ValidateMappingSpecification`. I've extracted the struct method but it's still too long.
Might be a bit controversial but I've temporarily disabled the warning, I'd argue a a long method with simple checks is easier to understand than one method calling 6 simple methods. WDYT?

latonz:
IMO its okay here to ignore the warning, could you change it to suppressmessage attribute? Often the readability is improved that when reading a method and submethods are called instead of inlined code, code which I'm not interested in reading can be skipped easier.

TimothyMakkison:
> IMO its okay here to ignore the warning, could you change it to suppressmessage attribute?

Added attribute
> Often the readability is improved that when reading a method and submethods are called instead of inlined code, code which I'm not interested in reading can be skipped easier.

Should I extract all the methods?

latonz (approved):

latonz:
Nope, should be fine like this, just my general opinion on the topic…

PR #554, opened 2023-07-10
fix: private members can override public members

### LINKED ISSUE #521: Private members can prevent the generation of valid mappings
**Describe the bug**
Private fields/properties can prevent a valid mapping being created if it has a similar name.
Mapperly sees a source with a greater priority, generates a memberpath, but then fails when `ObjectMemberMappingBodyBuilder` on line 154 checks if the source member is accessible.

**To Reproduce**
Create a private field or property with a greater name priority than another member.
```C#
[Mapper]
{
    public static partial B Map(A car);
}

public class A
{
    // private property, cannot be the source
    private int MyValue { get; set; }
    public C My { get; set; }
}
public class B
{
    public int MyValue { get; set; }
}
public class C
{
}
```
```C#
// generated
public static partial global::Riok.Mapperly.Sample.B Map(global::Riok.Mapperly.Sample.A car)
{
    var target = new global::Riok.Mapperly.Sample.B();
    return target;
}
```

**Expected behavior**
Mapping should ignore the private property and map via the correct field.

TimothyMakkison:
# Private members with greater flattening priority can override public members

## Description
Under specific circumstances, when Mapperly attempts to map to a target, a private source members with greater flattening priority can override another public member. A later validation check would reject the mapping as the source is inaccessible, emitting a diagnostic.
#537 prevents this by only mapping accessable source members.

This PR contains a test to validate that the bug is fixed.

Fixes #521

## Checklist

### DIFF
--- a/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
@@ -333,6 +333,29 @@ public class ObjectPropertyTest
     }
 
     [Fact]
+    public void PrivateMemberPropertyShouldNotOverride()
+    {
+        var source = TestSourceBuilder.Mapping(
+            "A",
+            "B",
+            "class A { private int MyValue { get; set; } public C My { get; set; } }",
+            "class B { public int MyValue { get; set; } }",
+            "class C { public int Value { get; set; } }"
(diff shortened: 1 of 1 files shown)


latonz:
@TimothyMakkison thank you for this test, could you rebase?

latonz (approved):

PR #556, opened 2023-07-11
fix: required/init properties not using PropertyNameMappingStrategy (…

### LINKED ISSUE #505: Mapping `init` or `required` properties does not use `PropertyNameMappingStrategy`
**Describe the bug**
Mapping to `required` or `init` properties doesn't use the `PropertyNameMappingStrategy`.

**Expected behavior**
Mapping should use `PropertyNameMappingStrategy`

**Code snippets**
```C#
[Mapper(PropertyNameMappingStrategy = PropertyNameMappingStrategy.CaseInsensitive)]
{
    public static partial B Map(A src);
}
public class A
{
}
public class B
{
    // can also be required
    public int value { get; init; }
}
```
**Generated code**
```C#
public static partial global::Riok.Mapperly.Sample.B Map(global::Riok.Mapperly.Sample.A src)
{
    var target = new global::Riok.Mapperly.Sample.B();
    return target;
}
```

cristianUrbina:
# Fix `required` and `init` properties not using `PropertyNameMappingStrategy`

## Description

Use the `MemberPath.TryFind` method overload with an `IEqualityComparer ` parameter inside `NewInstanceObjectMemberMappingBodyBuilder.BuildInitOnlyMemberMappings` and deleted the one it used before (a method overload without a comparer parameter)

Fixes #505 

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
@@ -34,6 +34,11 @@ public static class NewInstanceObjectMemberMappingBodyBuilder
 
     private static void BuildInitOnlyMemberMappings(INewInstanceBuilderContext<IMapping> ctx, bool includeAllMembers = false)
     {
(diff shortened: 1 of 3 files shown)


latonz (approved):
Thank you for this contribution! 🎉

PR #569, opened 2023-07-16
feat: add user implemented void mapping methods

### LINKED ISSUE #439: Pick up user implemented void mapping methods
Currently, Mapperly does not pick up user implemented [void mapping methods](https://mapperly.riok.app/docs/configuration/void-mapping-methods/). For example:

```csharp
[Mapper]
public static partial class MyMapper
{
    public static partial void CopyTo(HttpRequestMessage source, HttpRequestMessage dest);

    private static void CopyTo(HttpRequestHeaders source, HttpRequestHeaders dest)
    {
        foreach (var header in source)
        {
            dest.TryAddWithoutValidation(header.Key, header.Value);
        }
    }
}
```

Mapperly should recognize the user implemented void mapping method and should use it in the generated mapping code.

TimothyMakkison:
# Let mapperly use user implemented existing target methods

## Description
Add user implemented existing target mapping methods. Mapperly will attempt to use the users methods where possible

Fixes #439

## Checklist


TimothyMakkison:
Should I update the docs under `user implemented mappings` or `void mappings`?

latonz:
@TimothyMakkison I'll review the open PR's after 2.9.0 is released (we plan to release it next week). You can probably wait until the code / API changes are reviewed. If sth. fundamental changes after the review on the API surface, you then don't need to rewrite the docs again.

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
```diff
@@ -55,6 +55,29 @@ public record B(int Value);
         return TestHelper.VerifyGenerator(source);
     }
 
+    [Fact]
+    public Task InstanceMapperShouldSupportUserDefinedExistingTargetStaticMethods()
+    {
+        var source =
```
latonz, line 61: Use `Riok.Mapperly.Tests.TestSourceBuilder.CSharp` to get correct syntax highlighting in the IDE and probable use a raw string literal.

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
```diff
@@ -346,4 +411,32 @@ class B { public string Value { get; set; } public short Value2 { get; set; } }
         );
         return TestHelper.VerifyGenerator(source);
     }
+
+    [Fact]
+    public Task ExistingTargetWithMapperClassModifiersShouldCopyModifiersToMapper()
```
latonz, line 416: I don't think this needs an extra test, the code for an existing target mapper regarding the copy of the modifiers is exactly the same.

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
```diff
@@ -70,13 +70,21 @@ SymbolAccessor symbolAccessor
...
     )
     {
-        var valid =
-            method is { ReturnsVoid: false, IsGenericMethod: false }
-            && (allowPartial || !method.IsPartialDefinition)
-            && (!isStatic || method.IsStatic);
-        return valid && BuildParameters(ctx, method, out var parameters)
-            ? new UserImplementedMethodMapping(method, parameters.Source, parameters.ReferenceHandler)
-            : null;
+        var valid = method is { IsGenericMethod: false } && (allowPartial || !method.IsPartialDefinition) && (!isStatic || method.IsStatic);
```
latonz, line 73: Instead of pattern matching for a single property this code base usually uses direct comparisons `!method. IsGenericMethod`.

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedExistingTargetMethodMapping.cs
```diff
@@ -0,0 +1,75 @@
...
+using Riok.Mapperly.Emit;
+using Riok.Mapperly.Helpers;
+using Riok.Mapperly.Symbols;
+
+namespace Riok.Mapperly.Descriptors.Mappings.UserMappings;
+
+/// <summary>
+/// Represents a mapping method on the mapper which is implemented by the user.
+/// </summary>
+public class UserImplementedExistingTargetMethodMapping : ExistingTargetMapping, IUserMapping
```
latonz, line 14: I'm not really happy with this existing target method mapping implementing `ITypeMapping` as an existing target mapping can never support `Riok.Mapperly.Descriptors.Mappings.ITypeMapping.Build`.

I quickly poked around a little and I think the cleanest would be to introduce a new `INewInstanceMapping : ITypeMapping` with the build method. Then let the `IExistingTargetMapping` extend `ITypeMapping`.

You can find what I tried here: [latonz:inewinstancemapping](https://github.com/latonz/mapperly/tree/inewinstancemapping) / https://github.com/latonz/mapperly/commit/d3e786f7256d846fdbcfbfe9cf64d13400a11e7d.

What I did (more or less):
* Rename `ITypeMapping` / `TypeMapping` to `INewInstanceMapping` / `NewInstanceMapping`
* Added a new interface `ITypeMapping` and moved the build method from `ITypeMapping` to this interface
* Set the base interface of `INewInstanceMapping` and `IExistingTargetMapping` to `ITypeMapping`
* Adjusted `RuntimeTargetTypeMappingBodyBuilder ` to only consider `INewInstanceMapping`s.
* Adjusted `MappnigCollection`: Separate `Add`, `AddNewInstanceMapping` and `AddExistingTargetMapping` methods.

WDYT about this approach?

latonz (changes requested):

TimothyMakkison:
- Made suggested chnages
- Reworked `ITypeMapping`

Would it be worth making `AddExisting` and `AddNewInstance` private and using `MappingCollection.Add` directly? 
I haven't updated the docs.

latonz:
@TimothyMakkison feel free to create the issues.
I'd keep `AddExisting` and `AddNewInstance` public and call them if the type is known. This has a little performance benefit since the runtime type switch is not needed.

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedExistingTargetMethodMapping.cs
TimothyMakkison: >  But I think it makes a lot of things easier. The body builders often use the types provided by IMapping.

Yeah I agree. The part I dislike is having to pass down the values/ `ITypeMapping` depending on `IMapping`

I imagined something like the following, where `MappingBuilder` creates an `ITypeMapping` (`INewInstance`) and then creates an object inheriting from `IMapping`, sets `SourceType` and `TargetType` from context and adds `ITypeMapping`.

```C#
interface IMapping
{
    ITypeSymbol SourceType { get; }

    ITypeSymbol TargetType { get; }

    MappingBodyBuildingPriority BodyBuildingPriority { get; }

    ITypeMapping Mapping { get; }
}

tbh the this approach has flaws but it would make it possible to make all the mappings equatable between source generator runs. Not only would this have performance benefits but it would make it easier to do advanced mappings in the future.

When experimenting with reverse mapping, nested ignores/mappings and extra parameters; I would have to modify `MappingCollection` to store mappers based on their type and the `UserImplementedMethodMapping` context. Otherwise, depending on the order of generation, mappers would reuse a `MethodMapping` created by another mapper. Due to my changes the method may ignore or use a parameter that the base `UserImplementedMethodMapping` had not been configured for, resulting in an invalid mapping.

Separating the mappers by context would work but would result in a lot of bloated identical `MethodMappings` being generated. Most of the time these mapping methods would be identical in every way. By making all `ITypeMapping` and `MethodMapping` equatable it would be a lot easier to traverse the "graph" and merge/cull the identical mappers.

latonz (changes requested):

latonz:
2 Nitpicks and rebasing, afterwards this can get merged 😊

TimothyMakkison:
> 2 Nitpicks and rebasing, afterwards this can get merged 😊

Thanks for the review, done the changes. 👍

I think it will be best to update the docs after all the other void mappings are completed

Issue #571, opened 2023-07-17
Cannot map `Value` or `HasValue` Property of `Nullable<T>`


onionhammer:
**Describe the bug**
I upgraded from next-2 to next-3 and now I am receiving a build error

```
ModelMapper.g.cs(705,78): error CS1061: 'JsonElement' does not contain a definition for 'Value' and no accessible extension method 'Value' accepting a first argument of type 'JsonElement' could be found (are you missing a using directive or an assembly reference?)
```

It looks like the generated code is calling `.Value` twice:

```
// Generated code
        {
            var target = new global::SellerAmend()
            {
                Kind = source.Kind,
                SellerPartnerId = source.SellerPartnerId,
                SellerLoanId = source.SellerLoanId,
                DateCreated = source.DateCreated
            };
            if (source.ListingAmendment != null)
            {
                                                                   // .Value.Value should just be .Value
                target.ListingAmendmentValue = source.ListingAmendment.Value.Value.ToString();
            }

            return target;
        }

```

TimothyMakkison:
Haven't recreated it but I'm guessing because `Amendment` is a `Nullable<JsonElement>`, mapperly sees an object with a `Value` field and attempts to map?

onionhammer:
@TimothyMakkison I think its because of the `Value` suffix and some special case with Mapperly?

TimothyMakkison:
> @TimothyMakkison I think its because of the `Value` suffix and some special case with Mapperly for nullables?
> 
> If you rename `AmendmentValue` to `AmendmentVal` it compiles.

Yeah I've updated my comment with more speculation. I'm guessing it only affects nullable value types because its an actual type.
Does this not happen with V2?

onionhammer:
@TimothyMakkison maybe it was just doing a `throw new NotImplementedException` on v2

TimothyMakkison:
It's strange this issue didn't crop up when field mapping was introduced. Iirc one of my nit pick optimizations changed how nullable types were checked, perhaps that broke it.

TimothyMakkison:
I don't think #555 caused it I've been using the following
```C#
[Fact]
public void VaueTest()
{
    var source = TestSourceBuilder.MapperWithBodyAndTypes(
        "partial B Map(A src);",
        "public record A(int? P);",
        "public record B { public int PValue { get; set; } }"
    );
    TestHelper
        .GenerateMapper(source)
        .Should()
        .HaveSingleMethodBody(
            """
            var target = new global::B();
[...]

PR #575, opened 2023-07-18
fix: `Nullable<T>.Value` is no longer treated as a member

### LINKED ISSUE #572: Flattening nested nullable value types does not work
Mapping a flattened property, which has a nullable value type in the flattened path does not work.

Test (fails with RMG012):
```c#
    [Fact]
    public void NestedNullableValueTypeShouldWork()
    {
        var source = TestSourceBuilder.Mapping(
            "A",
            "B",
            "class A { public C? Prop { get; set; } }",
            "class B { public string PropInteger { get; set; } }",
            "struct C { public int Integer { get; set; } }"
        );

        TestHelper.GenerateMapper(source)
            .Should()
            .HaveSingleMethodBody(
                """
                if (source == null)
                    throw new System.ArgumentNullException(nameof(source));
                var target = new global::B();
                if (source.Prop != null)
                {
                    target.PropValue = source.Prop.Value.Integer.ToString();
                }

                return target;
                """);
    }
```

TimothyMakkison:
# `Nullable<T>.Value` is no longer treated as a member

## Description
Added check to `SymbolAccessor` for nullable value types. This way `Nullable<T>.Value` will not be treated as a member resulting in invalid auto resolved flattenings

Fixes #572 & #571

## Checklist

### DIFF
--- a/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFlatteningTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFlatteningTest.cs
@@ -1,3 +1,5 @@
+using Riok.Mapperly.Diagnostics;
+
 namespace Riok.Mapperly.Tests.Mapping;
 
 [UsesVerify]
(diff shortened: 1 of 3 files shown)


test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFlatteningTest.cs
```diff
@@ -126,6 +128,79 @@ public void AutoFlattenedMultiplePropertiesNullablePath()
...
+        TestHelper
+            .GenerateMapper(source, TestHelperOptions.AllowDiagnostics)
+            .Should()
+            .HaveDiagnostic(DiagnosticDescriptors.SourceMemberNotFound)
+            .HaveSingleMethodBody(
+                """
+            var target = new global::B();
+            return target;
+            """
+            );
```
latonz, line 201: This seems to have a weird indention... Same at the other assertion method bodies...

latonz (changes requested):
Thank you for this PR, only little formatting issues...

latonz (approved):

PR #592, opened 2023-07-24
fix: disable warning when iterating nullable array elements

### LINKED ISSUE #416: Converting `Nullable<T>` to `T` generates warning message.
Generated code for #413, generates a `null reference warning`. I'd guess this is because .NET treats each call to the same array index as a potentially different value. I've found two refactors that will suppress this.

```C#
private static string[] MapToStringArray(string?[] source)
{
    var target = new string[source.Length];
    for (var i = 0; i < source.Length; i++)
    {
        // below has the `null reference warning`
        target[i] = source[i] == null ? throw new System.NullReferenceException($"Sequence {nameof(source)}, contained a null value.") : source[i];

        // Reolved with
        // var placeholder = source[i];
        // target[i] = placeholder == null ? throw new System.NullReferenceException($"Sequence {nameof(source)}, contained a null value.") : placeholder;

        // or
        // target[i] = source[i] ?? throw new System.NullReferenceException($"Sequence {nameof(source)}, contained a null value.");
    }

    return target;
}
```

TimothyMakkison:
# Disable warning when accessing nullable elements 

## Description
As suggested added null supression to `NullDelegateMapping` when accessing array elements.

Fixes #416

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
@@ -1,4 +1,5 @@
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Riok.Mapperly.Helpers;
 using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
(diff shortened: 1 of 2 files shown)


latonz (approved):

TimothyMakkison:
Code coverage bot generated a weird message after rebating. Not sure how to rerun the ci pipeline

latonz (approved):

PR #603, opened 2023-07-28
fix: improve performance for member resolution

latonz:
Use a cache dictionary for the mappable members by name. This increases the performance of mapping classes with lots of members, since the lookup by member name is now constant time instead of linear.

### DIFF
--- a/src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
+++ b/src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
@@ -16,6 +16,10 @@ public class SymbolAccessor
     private readonly Dictionary<ITypeSymbol, IReadOnlyCollection<ISymbol>> _allMembers = new(SymbolEqualityComparer.Default);
     private readonly Dictionary<ITypeSymbol, IReadOnlyCollection<IMappableMember>> _allAccessibleMembers =
         new(SymbolEqualityComparer.Default);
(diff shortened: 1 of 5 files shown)


src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -16,6 +16,10 @@ public class SymbolAccessor
     private readonly Dictionary<ITypeSymbol, IReadOnlyCollection<ISymbol>> _allMembers = new(SymbolEqualityComparer.Default);
     private readonly Dictionary<ITypeSymbol, IReadOnlyCollection<IMappableMember>> _allAccessibleMembers =
         new(SymbolEqualityComparer.Default);
+    private readonly Dictionary<ITypeSymbol, IReadOnlyDictionary<string, IMappableMember>> _allAccessibleMembersCaseInsensitive =
```
ni507, line 19: Aren't these dictionary always empty?

latonz: Nope, they are filled on L199: `membersBySymbol[symbol] = symbolMembers = GetAllAccessibleMappableMembers(symbol)...` with `membersBySymbol` beeing one of the two (L193 `var membersBySymbol = ignoreCase ? _allAccessibleMembersCaseInsensitive : _allAccessibleMembersCaseSensitive;`).

ni507: Missed that.

ni507 (approved):

PR #588, opened 2023-07-22
feat: optimize enumerable mappings

TimothyMakkison:
# Replace linq usage with `for` and `foreach` loops

## Description
- Updated logic for implicit enumerable conversions, used when mapping synthetic.
- For known length sources use a `foreach` loop to construct an array or list. 
   - `Array`, `IReadOnlyList`, `IReadOnlyCollection`, `IEnumerable` -> `Array`
   - `List`, `IList`, `ICollection` -> `List`
- Keep linq logic for projection mapping and for `IEnumerable` source/target mapping.

Related #433

## Checklist

- [ ] The existing code style is followed
- [ ] The commit message follows our guidelines
- [ ] Performed a self-review of my code
- [ ] Hard-to-understand areas of my code are commented
- [ ] Unit tests are added/updated
- [ ] Integration tests are added/updated (as applicable, especially if feature/bug depends on roslyn or framework version in use)

PR #621, opened 2023-08-04
fix: arrays use correct element nullability

### LINKED ISSUE #229: Nullability in array mappings not quite correct

TimothyMakkison:
# fix: Array methods use the correct nullability

## Description
Modified `ArrayForMapping` to use the correct element nullability in the method signature.

This could be fixed in `NullableSymbolExtensions` by making it check for `IEnumerable` types. Doing this would create a dependency on `SymbolAccessor` or `WellKnownTypes` and probably create a small performance hit. If #588 has this problem this approach could be revisited.

Fixes #229 

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
@@ -54,7 +54,14 @@ public static class EnumerableMappingBuilder
             // a single Array.Clone / cast mapping call should be sufficient and fast,
             // use a for loop mapping otherwise.
             if (!elementMapping.IsSynthetic)
-                return new ArrayForMapping(ctx.Source, ctx.Target, elementMapping, elementMapping.TargetType);
+            {
+                // upgrade nullability of element type
+                var targetValue =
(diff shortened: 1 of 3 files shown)


src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -54,7 +54,13 @@ or CollectionType.IEnumerable
             // a single Array.Clone / cast mapping call should be sufficient and fast,
             // use a for loop mapping otherwise.
             if (!elementMapping.IsSynthetic)
-                return new ArrayForMapping(ctx.Source, ctx.Target, elementMapping, elementMapping.TargetType);
+            {
+                // upgrade nullability of element type
+                var targetValue = ctx.Target.IsArrayType()
```
latonz, line 59: Couldn't this be simplified to `ctx.CollectionInfos.Target.CollectionType == CollectionType.Array`?

latonz (changes requested):
Thank you for this PR, one little comment 😊

latonz (approved):

latonz (approved):

Issue #103, opened 2022-07-18
Supply additional parameters to the mapping method


cyril265:
Sometimes you want to provide an external value to the mapping method. For example your DTO doesn't have a tenant id. You are getting the tenant id from the currently signed in user and it's the same for every entity inside the current request.

```c#
[Mapper]
public partial class CarMapper
{
    public partial Car Map(CarDto dto, string tenantId);
}

public record Car(string Name, string TenantId);

public record CarDto(string Name);

```
Example usage:

```C#
    public void UseMapper(CarMapper mapper, IEnumerable<CarDto> dtos)
    {
        string tenantId = ""; //GetFromDb;
        var entities = dtos.Select(dto => mapper.Map(dto, tenantId));
    }
```
You could accomplish this by moving `TenantId` to a property and using the after map feature. But I think this solution is cleaner. It doesn't force you to change your domain code for a mapper, it shows the intent more clearly and it's simpler to add.

latonz:
I like this idea. However, it is not that easy to implement. It may take some time until Mapperly supports this.

latonz:
@Fube Sorry, I missed the notification of your comment.
I didn't have the time to think about a concrete architecture yet. Are all arguments considered as mapping sources? Or are the additional arguments only sources for properties with the same name? On a first glance I think if all arguments are considered as mapping sources, the `TypeMapping` needs to be refactored. Otherwise it seems to be limited to the `PropertyMapping` and it's related classes. Tbd is also how to handle it, if another mapping depends on a user defined mapping with additional arguments. 

Feel free to submit an architectural proposal to this issue.

DerMave:
I would like to work on this feature as I have an usecase for it. Have you begun working on this? @latonz @Fube

Fube:
@DerMave 
Unfortunately not, but feel free to takeover.
I investigated the issue, but did not have time to come up with a solution.

As far as I can tell, the `TypeMapping` has to be changed to account for more than one source.

latonz:
@DerMave I didn't have time yet to work on this. Let me know if you start working on this, then I'd assign the issue to you.
However, before you start with an actual implementation, I think it would be useful to discuss a design proposal on how this feature should work exactly and which parts of Mapperly need adjustments.

akoken:
+1 for this feature. I think this is very common use case. As @cyril265 mentioned above, we also obtain tenantId, userId etc. from HttpContext.

DerMave:
[...]

PR #622, opened 2023-08-04
POC Extra parameters 

TimothyMakkison:
POC for #103. This is a simpler version of #550, parameters do not attempt to find a matching memberpath nor are they passed down to sub methods.

- `IMapping` has  a `ImmutableEquatableArray<MethodParameter> Parameters` property
- All user defined mappings should be able to support extra parameters
- Unlike #550, passing down parameters is severly limited. Therefore, the initialization of  `MethodMapping` is not significantly changed.
- Extra parameters can resolve constructor arguements, required/init and object member mapping, with support for `[MapProperty]`.

```C#
public static partial UserDto Map(User src, int id):
```
```C#
public static partial UserDto Map(User src, int id)
{
    var target = new UserDto();
    target.Username = src.Username;
    target.Id = id;
}
``` 

### Parameter usage
This PR has a simple way of using parameters, if a target member has the name `username`, mapperly will attempt to resolve this from the source parameter and then look for a parameter named `username`. 
It will not attempt to a parameter with a matching memberpath i.e. given a member `username` match with parameter `,User user)` where `user.Name -> username`  

This does diverge from how the source parameters are treated, but given that extra parameters use their name to resolve paths, it perhaps makes more sense. This could easily be changed.

### Method resolving
I believe that `user defined` mappers should be able to call other `user defined/implemented` mappers if they have compatible signatures and paramters. I have not decided on the best approach for method resolving. Part of the challenge is choosing a resolving approach that will not prevent future changes.

- Should parameters match using type and name or just name and map to type?
- Similarly to `Parameter usage` should matching member paths be used?

#### Approaches
- Only use another mapper if it has the same signature and parameters.
- Try to match parameters by name while preserving the order of the parameters:
	- (long id, string name, int age) can use (string name, int age)
	- (long id, string name, int age) can't use (int age, long id)
- Match parameters by name - could cause issues if mapperly created methods are allowed to use extra parameters:
	- (long id, string name, int age) can use (string name, int age)
	- (long id, string name, int age) can use (int age, long id)


latonz:
Thank you for this PR 🚀  I did not start the code review yet, I'll review it in the upcoming days. Overall this looks like a promising feature 😊 I prefer the simple approach of not passing the additional parameters, because otherwise it can become really hard to understand when which methods are invoked/generated and why.

Regarding your description of the PR:
> It will not attempt to a parameter with a matching memberpath i.e. given a member username match with parameter ,User user) where user.Name -> username

I don't get this. Do you mean `MemberPathCandidateBuilder.BuildMemberPathCandidates` is not used for additional source parameters? If so, whats the reasoning behind this?

> Method resolving
In a first version of this feature, I would not pass additional parameters anywhere, too keep it simple. Otherwise this could get quite complex when and why another method is called/generated. I'm not even sure if there is demand for this advanced feature. For now if this is needed, it is always possible to make keep a context inside the mapper class, instantiate a mapper with the context and use after map to set these fields.

TimothyMakkison:
> Thank you for this PR 🚀 I did not start the code review yet, I'll review it in the upcoming days. Overall this looks like a promising feature 😊 I prefer the simple approach of not passing the additional parameters, because otherwise it can become really hard to understand when which methods are invoked/generated and why.



> Regarding your description of the PR:
> 
> > It will not attempt to a parameter with a matching memberpath i.e. given a member username match with parameter ,User user) where user.Name -> username
> 
> I don't get this. Do you mean `MemberPathCandidateBuilder.BuildMemberPathCandidates` is not used for additional source parameters? If so, whats the reasoning behind this?

The previous PR would create the following.

```C#
public record UserInfo (int Id);;

public record UserDto(string UserId, ...);

public static partial UserDto MapTo(SourceInfo src, UserInfo user);

// generates the following
public static partial UserDto MapTo(SourceInfo src, UserInfo user)
{
    var target = new UserDto(user.Id, ...);
    return target;
}
```

Here the extra parameter named `user` matches to the target `UserId` because the parameters name combined with the member `Id` match `UserId`.
I thought this behaviour could be confusing as the source parameters name is never combined with its members.
```C#
public static partial DogDto Map(Dog dog)
// does not generate
target.DogOwnder = dog.Owner;
```

On the other hand, extra parameters already diverge from the source parameeter by using there name. So by not having this behaviour could also be confusing 

> > Method resolving
> 
> In a first version of this feature, I would not pass additional parameters anywhere, too keep it simple. Otherwise this could get quite complex when and why another method is called/generated. I'm not even sure if there is demand for this advanced feature. For now if this is needed, it is always possible to make keep a context inside the mapper class, instantiate a mapper with the context and use after map to set these fields.

#518 : thumbs

TimothyMakkison:
> Thank you for this PR 🚀 I did not start the code review yet, I'll review it in the upcoming days. Overall this looks like a promising feature 😊 I prefer the simple approach of not passing the additional parameters, because otherwise it can become really hard to understand when which methods are invoked/generated and why.

Thanks, please don't spend too much time reviewing this, it will likely change if I modify it. I would appreciate some pointers with my abstrctions, i.e. should `IMapping` know about the parameters etc.

> Regarding your description of the PR:
> 
> > It will not attempt to a parameter with a matching memberpath i.e. given a member username match with parameter ,User user) where user.Name -> username
> 
> I don't get this. Do you mean `MemberPathCandidateBuilder.BuildMemberPathCandidates` is not used for additional source parameters? If so, whats the reasoning behind this?

The previous PR would create the following.

```C#
public record UserInfo (int Id);;

public record UserDto(string UserId, ...);

public static partial UserDto MapTo(SourceInfo src, UserInfo user);

// generates the following
public static partial UserDto MapTo(SourceInfo src, UserInfo user)
{
    var target = new UserDto(user.Id, ...);
    return target;
}
```

Here the extra parameter named `user` matches to the target `UserId` because the parameters name combined with the member `Id` match `UserId`.
I thought this behaviour could be confusing as the source parameters name is never combined with its members.
```C#
public static partial DogDto Map(Dog dog)
// does not generate
target.DogOwnder = dog.Owner;
```

On the other hand, extra parameters already diverge from the source parameter by using their name. So not having this behaviour could also be confusing 

> > Method resolving
> 
> In a first version of this feature, I would not pass additional parameters anywhere, too keep it simple. Otherwise this could get quite complex when and why another method is called/generated. I'm not even sure if there is demand for this advanced feature. For now if this is needed, it is always possible to make keep a context inside the mapper class, instantiate a mapper with the context and use after map to set these fields.

I agree 😊. As I found before, passing extra parameters to auto-generated methods means you have to account for scopes/configuration contexts. I had to monkey-patch a way of changing the order of mapping generation from a simple queue into a recursive depth-first generator while also accounting for loops.

I was thinking of how to support extra parameters for user implemented and defined mappers. This wouldn't have the above issues. We can see if this is needed.

For example in the below code, mapperly could use the user defined method `MapProducer` because it has a matching parameter.

```C#
{
    public static partial CarDto MapCarToDto(Car car, string name);
    
    public static ProducerDto MapProducer(Producer src, string name)
    {
        // ...
     }
}
```

latonz:
I converted this PR to draft for now.

> For example in the below code, mapperly could use the user defined method MapProducer because it has a matching parameter.

I think this would still result in weird behaviours. For example if the Car has a property of type Tire which then has a property of type Producer. The mapping of the tire is not user implemented but generated (without the name parameter). Then the user implemented `MapProducer` is never called, because the name parameter got lost in the generated map method for the tire mapping. But if the Producer was directly on the Car, the name parameter would have been passed. IMO this behaviour would be hard difficult to follow.
I think for nested / advanced scenarios a `new MyMapper(myAdditionalContext).Map(...)` with before/after map is still the best solution.

TimothyMakkison:
Great point, this and the method resolution issues could make this pretty janky.

Thankfully `[MapProperty]` can be used with extra parameters.

src/Riok.Mapperly/Descriptors/Mappings/IMapping.cs
```diff
@@ -11,5 +12,7 @@ public interface IMapping
 
     ITypeSymbol TargetType { get; }
 
+    ImmutableEquatableArray<MethodParameter> Parameters { get; }
```
latonz, line 15: `ImmutableEquatableArray<MethodParameter> Parameters { get; }` can then be moved from `IMapping` to `IUserMapping`, the parameters could be passed from `UserMethodMappingBodyBuilder.BuildMappingBody` to the context. I'd also rename to `AdditionalMappingParameters` or similar as this array does not include all parameters of the method.

latonz, line 15: Woulnd't `IReadOnlyCollection<T>` be sufficient here?

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
```diff
@@ -264,27 +268,32 @@ bool isStatic
...
-        }
-
-        parameters = new MappingMethodParameters(sourceParameter.Value, targetParameter, refHandlerParameter);
+        // if (method.Parameters.Length != expectedParameterCount)
+        // {
+        //     parameters = null;
+        //     return false;
+        // }
+        var extraParams = method.Parameters
+            .Skip(expectedParameterCount)
```
latonz, line 291: Instead of skipping the expected parameter count, keep a set of already processed parameter ordinals and check the additional parameter ordinals are not included in this set.

src/Riok.Mapperly/Descriptors/Mappings/TypeMapping.cs
```diff
@@ -12,12 +13,22 @@ protected TypeMapping(ITypeSymbol sourceType, ITypeSymbol targetType)
...
+        SourceType = sourceType;
+        TargetType = targetType;
+        Parameters = parameters;
     }
 
     public ITypeSymbol SourceType { get; }
 
     public ITypeSymbol TargetType { get; }
 
+    public ImmutableEquatableArray<MethodParameter> Parameters { get; }
```
latonz, line 30: With the approach that we never pass parameters to other methods, only user defined mapping methods should not support parameters. For eg. user implemented mappings a diagnostic should be emitted if additional parameters are provided (RMG001).

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
```diff
@@ -162,4 +181,17 @@ private void AddUnmatchedSourceMembersDiagnostics()
             );
         }
     }
+
+    private void AddUnmatchedParametersDiagnostics()
```
latonz, line 185: With the current architecture, if the user defined mapping is not an object mapping (eg. string to int), RMG048 is not emitted.

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
```diff
@@ -60,8 +61,20 @@ ITypeSymbol targetType
...
+            .Zip(ctx.Parameters, (a, b) => (a, b))
+            .Select(x => x.a.WithArgument(x.b))
+            .Cast<MethodArgument?>()
+            .ToArray();
+        return Invocation(
+            MethodName,
+            SourceParameter.WithArgument(ctx.Source),
+            ReferenceHandlerParameter?.WithArgument(ctx.ReferenceHandler),
+            parameters
+        );
```
latonz, line 76: This is probably not needed anymore (also the Wrap10) if we never automatically call methods with additional parameters (`CallableByOtherMappings = false`).

latonz (changes requested):
I did a quick review but I think a lot of code will change when we only consider additional parameters in the user defined mapping methods and don't pass them anywhere. Therefore I'll do a complete review once this is adjusted 😊  
I added the few things I noticed as review feedback.

latonz:
As there seems to be no updates for several weeks I'm closing this for now. Feel free to reopen / open a new PR with updates 😊

PR #637, opened 2023-08-08
chore: increase performance threshold

latonz:
Increase the performance threshold due to flaky results of the GH runner.


TimothyMakkison (commented):
Might be worth disbaling `fail on alert` for now.

I expected the benchmarks to vary a little but this is pretty bad.

TimothyMakkison (approved):

latonz:
@TimothyMakkison I'd try with 200% (this way we still have an alert if the performance decreases a lot). If that still won't work I'll disable it completely.

PR #655, opened 2023-08-18
chore: Remove multi targeting of the source generator and remove InternalsVisibleTo to make VS 17.6+ work again

latonz:
Remove multi targeting of the source generator (now only targets netstandard2.0).
Also needed to remove InternalsVisibleTo from Riok.Mapperly to Riok.Mapperly.Tests. Otherwise the internal polyfills would be visible to the tests. But the tests don't need the polyfills since they target net7.0. Made some Riok.Mapperly API's public to test them. Since Riok.Mapperly doesn't have a public API subject to SemVer (except for the source generator, Riok.Mapperly.Abstractions and the generated code), making these classes public shouldn't really matter.


hoyau (dismissed):
Works in VS 17.7.1:

- MSBuild without Warnings or Errors
- Can Debug the Source Generator
- Tests (759) all green

CommonGuy (approved):

hoyau (approved):

TimothyMakkison (approved):

PR #657, opened 2023-08-18
feat: add default mapper attribute for assemblies

### LINKED ISSUE #201: Default mapper configuration per assembly
Allow to set an attribute on the assembly level to set default configurations applied to all mappers.
These should be overwritable by configurations via a `MapperAttribute` on a mapper class. A mapper class still needs the `MapperAttribute` to mark the class to be taken into account by Mapperly.

The attribute for the assembly level configurations may use another name, since probably (now or in the future) not all and/or different options may be configurable on an assembly level.

ni507:
Fixes #201


src/Riok.Mapperly/AnalyzerReleases.Shipped.md
```diff
@@ -106,3 +106,11 @@ RMG044  | Mapper   | Warning  | An ignored enum member can not be found on the s
 RMG045  | Mapper   | Warning  | An ignored enum member can not be found on the target enum
 RMG046  | Mapper   | Error    | The used C# language version is not supported by Mapperly, Mapperly requires at least C# 9.0
 RMG047  | Mapper   | Error    | Cannot map to member path due to modifying a temporary value, see CS1612
+
+## Release 3.1
```
latonz: This PR will miss 3.1, set it to 3.2

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
```diff
@@ -106,3 +106,11 @@ RMG044  | Mapper   | Warning  | An ignored enum member can not be found on the s
...
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+RMG048  | Mapper   | Error    | Cannot use multiple attributes in the same assembly
```
latonz, line 116: IMO `multiple attributes` can be misleading as only multiple attributes of the type `DefaultMapperAttribute` are not supported.

docs/docs/configuration/mapper.mdx
```diff
@@ -185,3 +185,7 @@ dotnet_diagnostic.RMG020.severity = error # Unmapped source member
 ### Strict enum mappings
 
 To enforce strict enum mappings set `RMG037` and `RMG038` to error, see [strict enum mappings](./enum.mdx).
+
+## Default Mapper configuration
+
+The `DefaultMapperAttribute` allows to set default configurations applied to all mappers on the assembly level.
```
latonz, line 191: Add an example

src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -83,6 +83,13 @@ CancellationToken cancellationToken
             if (!symbolAccessor.HasAttribute<MapperAttribute>(mapperSymbol))
                 continue;
 
+            if (symbolAccessor.HasMultipleAssemblyAttribute<DefaultMapperAttribute>())
+            {
+                diagnostics.Add(
+                    Diagnostic.Create(DiagnosticDescriptors.CannotUseMultipleAssemblyAttributes, null, typeof(DefaultMapperAttribute))
```
latonz, line 89: Doesn't this emit the diagnostic for each mapper class instead of once per assembly?

src/Riok.Mapperly/Configuration/MergedMapperConfiguration.cs
```diff
@@ -0,0 +1,104 @@
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.Descriptors;
+using Riok.Mapperly.Helpers;
+
+namespace Riok.Mapperly.Configuration;
+
+public class MergedMapperConfiguration
```
latonz, line 8: Wouldn't be a name like `MapperConfigurationReader` more appropriate here?

src/Riok.Mapperly/Configuration/MapperConfiguration.cs
```diff
@@ -1,101 +1,85 @@
-using Microsoft.CodeAnalysis;
 using Riok.Mapperly.Abstractions;
-using Riok.Mapperly.Descriptors;
-using Riok.Mapperly.Helpers;
 
 namespace Riok.Mapperly.Configuration;
 
 public class MapperConfiguration
```
latonz, line 5: Add a comment what exactly this class does and why it is almost equal to the MapperAttribute except for the nullability.

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
latonz: Is this diagnostic even requried as AllowMultiple is false on the attribute, isn't there already a diagnostic by the csharp compiler if it is applied multiple times?

src/Riok.Mapperly/Configuration/MergedMapperConfiguration.cs
```diff
@@ -0,0 +1,104 @@
...
+{
+    private readonly MappingConfiguration _defaultConfiguration;
+    private readonly AttributeDataAccessor _dataAccessor;
+
+    public MergedMapperConfiguration(SymbolAccessor symbolAccessor, ISymbol mapperSymbol)
+    {
+        _dataAccessor = new AttributeDataAccessor(symbolAccessor);
+        var mapperConfiguration = _dataAccessor.AccessSingle<MapperAttribute, MapperConfiguration>(mapperSymbol);
+        var defaultMapperConfiguration = _dataAccessor.AccessAssemblyFirstOrDefault<DefaultMapperAttribute, MapperConfiguration>();
```
latonz, line 17: This reads the assembly configuration for each mapper class which could be inefficient for large code bases. Instead we should try to read it once for all mapper classes. This should work similar to how we extract the class declarations in the mapper generator. Eg. for roslyn 4.4 this could look like:
```c#
public static IncrementalValuesProvider<CompilationUnitSyntax> GetAssemblyDefaultMappers(IncrementalGeneratorInitializationContext context)
    {
        return context.SyntaxProvider.ForAttributeWithMetadataName(
            "Riok.Mapperly.Abstractions.DefaultMapperAttribute",
            static (s, _) => s is CompilationUnitSyntax,
            static (ctx, _) => ctx.TargetNode as CompilationUnitSyntax);
}
```

latonz (changes requested):

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
ni507: Already pushed after the release.

latonz (changes requested):

src/Riok.Mapperly.Abstractions/MapperAttribute.cs
```diff
@@ -6,7 +6,7 @@ namespace Riok.Mapperly.Abstractions;
 /// Marks an abstract class or an interface as a mapper.
 /// </summary>
 [AttributeUsage(AttributeTargets.Interface | AttributeTargets.Class)]
-public sealed class MapperAttribute : Attribute
+public class MapperAttribute : Attribute
```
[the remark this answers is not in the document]
CommonGuy: I think unsealing this is ok if it makes it easier for us

docs/docs/configuration/mapper.mdx
ni507: Added an example.

src/Riok.Mapperly/Configuration/MapperConfiguration.cs
ni507: Added a comment.

src/Riok.Mapperly/Configuration/MergedMapperConfiguration.cs
ni507: Changed to `MapperConfigurationReader`.

src/Riok.Mapperly/Configuration/MergedMapperConfiguration.cs
ni507: Used a different way. Accessed the assembly attributes with the `Microsoft.CodeAnalysis.Compilation` which is already provided through the context.

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
ni507: Changed the text of the diagnostic. It is required, since no diagnostic is created from the csharp compiler.

docs/docs/configuration/mapper.mdx
```diff
@@ -185,3 +185,11 @@ dotnet_diagnostic.RMG020.severity = error # Unmapped source member
...
+
+```csharp
+[assembly: Riok.Mapperly.Abstractions.DefaultMapperAttribute(EnumMappingIgnoreCase = true)]
```
latonz, line 194: IMO using the short name should be enough here, and makes it also simpler to read `[assembly: MapperDefaults(EnumMappingIgnoreCase = true)]`

src/Riok.Mapperly/MapperGenerator.cs
```diff
@@ -72,6 +74,25 @@ CancellationToken cancellationToken
...
+            diagnostics.Add(
+                Diagnostic.Create(DiagnosticDescriptors.CannotUseMultipleAssemblyAttributes, null, typeof(DefaultMapperAttribute))
+            );
+        }
+
+        var defaultMapperAttribute = defaultMapperAttributes.FirstOrDefault();
+        var defaultConfiguration =
+            defaultMapperAttribute != null
+                ? AttributeDataAccessor.Access<DefaultMapperAttribute, MapperConfiguration>(defaultMapperAttribute)
+                : null;
```
latonz, line 94: To reduce this method's complexity, I'd extract this into a separate method.
Couln't this be extracted in another incremental step? I think this could improve the overall performance.

test/Riok.Mapperly.Tests/Helpers/MapperConfigurationBuilderTest.cs
```diff
@@ -0,0 +1,90 @@
+using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.Configuration;
+using Riok.Mapperly.Helpers;
+
+namespace Riok.Mapperly.Tests.Helpers;
+
+public class MapperConfigurationBuilderTest
+{
+    [Fact]
+    public void ShouldMergeMapperConfigurations()
```
latonz, line 10: Couldn't we use Mapperly itself to ensure `MapperConfiguration` and `MapperAttribute` share all attributes by adding a map method in the integration tests and enforcing strict mappings?

latonz (changes requested):

docs/docs/configuration/mapper.mdx
ni507: Fixed.

test/Riok.Mapperly.Tests/Helpers/MapperConfigurationBuilderTest.cs
ni507: Mapperly itself is not possible. Added a comment.

latonz (approved):

PR #681, opened 2023-08-24
Generate static methods in nonstatic classes

### LINKED ISSUE #673: Use interface with a static abstract member on class managed by Mapperly
Use interface with a static abstract member on class managed by Mapperly

I would like that this code will generate mapping:
```cs
public interface IMyInterface
{
  static abstract void Update(SourceDto source, DestinationDto destination);
}

[Mapper]
public partial class MyMapping : IMyInterface
{
  public static partial void Update(SourceDto source, DestinationDto destination);
}
```

A form of a hack is this solution:
```cs
public interface IMyInterface
{
  static abstract void Update(SourceDto source, DestinationDto destination);
}

public class MyMapping : IMyInterface
{
  public static void Update(SourceDto source, DestinationDto destination) => MyMappingInternal.Update(source, destination);
}

[Mapper]
public static partial class MyMappingInternal
{
  public static partial void Update(SourceDto source, DestinationDto destination);
}
```

**Additional context**
For some reason class with the attribute `[Mapper]` does not generate static members when the class itself is not static. The issue with static interface members is that interface cannot be defined on class which is static at the same time it's members are static

trejjam:
# Generate static methods in nonstatic classes

## Description

This PR adds the Mapper configuration option `GenerateStaticMethods` which is a switch that generates static methods inside a non-static class.

Fixes #673

## Checklist


trejjam:
Hi, @latonz, @TimothyMakkison,

can I ask you for your opinion about this functionality? And if so, can I ask for feedback on implementation?

Thank you,
Jan

trejjam:
The goal is to support this sode:

```cs
public interface IMyInterface
{
    static abstract CarDto ToDto(Car car);
}

[Mapper(GenerateStaticMethods = true)]
public partial class CarMapper : IMyInterface
{
    public static partial CarDto ToDto(Car car);
}
```

latonz:
Sorry for my late feedback, I had the pleasure to enjoy some vacation days 🏝️
Thank you for opening the corresponding issue and submitting this PR, I understand your use case.
On a first glance the current implementation has several not yet addressed problems:
* Generated nested mapping methods are not static
(to reproduce simply add `public static partial Riok.Mapperly.IntegrationTests.Dto.TestObjectDto Map(Riok.Mapperly.IntegrationTests.Models.TestObject value);` to the `TestMapperWithStaticMethods` in your integration tests)
* Non-Static user defined mappings are called from generated static mapping methods
* Non-Static user implemented mappings are called from generated static mapping methods
* Non-Static object factories are called from generated static mapping methods
* Non-Static external mappers are called from generated static mapping methods (probably due to rebasing, only landed recently in main)

I will comment on the issue https://github.com/riok/mapperly/issues/673 with the concerns that I have for this feature.

trejjam:
Thanks, I will take a look at the missing parts

latonz:
@trejjam before you are going to continue working on this PR we should probably agree on the design. I added my bits [here](https://github.com/riok/mapperly/issues/673#issuecomment-1697382302).

trejjam:
Yes, for sure :)

trejjam:
Hi @latonz,
I finally found some time to finish this PR. I will be happy if you find some time for a review.

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -94,6 +96,47 @@ private void ConfigureMemberVisibility()
 #endif
     }
 
+    private bool IsInstanceWithStaticPartialMethod()
```
latonz, line 99: This extracts the user method mappings an additional time. Is this really required? I think this could be simplified a lot with the following approach:
* Revert changes in `UserMethodMappingExtractor`
* In `Riok.Mapperly.Descriptors.DescriptorBuilder.ExtractUserMappings`:
    * Use a boolean to track whether all mappings should be static, initialize it to true if the mapper class itself is static
    * Store the first non-static user mapping method symbol
    * Simplified, it could look like this:
    ```c#
        var allShouldBeStatic = _mapperDescriptor.Symbol.IsStatic;
        IMethodSymbol? firstNonStaticUserMapping = null;

        foreach (...)
        {
            if (userMapping.Method.IsStatic)
            {
                allShouldBeStatic = true;
            }
            else if (firstNonStaticUserMapping == null)
            {
                firstNonStaticUserMapping = userMapping.Method;
            }

            ...
        }
        ...
     ```
     * after the extraction of all user mappings is done report the diagnostic on `firstNonStaticUserMapping` if `firstNonStaticUserMapping` is not null and `allShouldBeStatic` is true.

test/Riok.Mapperly.Tests/Mapping/InstantiableMapperWithStaticMethodsTest.cs
```diff
@@ -0,0 +1,80 @@
...
+public class InstantiableMapperWithStaticMethodsTest
+{
+    [Fact]
+    public Task StaticPartialMethod()
+    {
+        var source = TestSourceBuilder.CSharp(
+            """
+            using Riok.Mapperly.Abstractions;
+
+            record A(int Value);
```
latonz, line 15: Can you add a test with a nested object type?

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
```diff
@@ -33,6 +33,7 @@ public abstract class MethodMapping : NewInstanceMapping
     private readonly IMethodSymbol? _partialMethodDefinition;
 
     private string? _methodName;
+    private bool? _methodIsStatic;
```
latonz, line 36: I don't think this approach does fully work. Scenario where I think it wont work (probably add a unit test for this):
* Non-static mapper class
* Source object tree `A>B>C>int`
* Target object property tree `D>E>F>int`.
* User defined mapping method from A to D
* User implemented mapping method from C to F

Mapperly generates the B to E method as an instance method. From this method it then calls the user implemented static method to convert from C to F. I think whether the mapper is a static/pseudo-static mapper needs to be extracted into the descriptor and passed to the `SourceEmitterContext` in `Riok.Mapperly.Emit.SourceEmitter.Build`.

latonz (changes requested):
Thank you for the updates, I added my feedback 😊

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
trejjam: I changed it to SourceEmitterContext version, and I added some integration tests:
image
I hope it tests what you mentiond

trejjam:
Hi, thank you for the review! I processed all recommendations.

I left changes as separate commits, so it's easier to navigate through changes. Please let me know for the final rebase/squash

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
latonz: IMO this should rather be a unit test then an integration test. It's behaviour doesn't really need to be tested on different .net versions (that's mainly what integration tests are for).

src/Riok.Mapperly/Emit/SourceEmitter.cs
```diff
@@ -11,10 +11,14 @@ public static class SourceEmitter
...
-    public static CompilationUnitSyntax Build(MapperDescriptor descriptor, CancellationToken cancellationToken)
+    public static CompilationUnitSyntax Build(
+        MapperDescriptor descriptor,
+        bool instanceWithStaticPartialMethod,
+        CancellationToken cancellationToken
+    )
     {
         var ctx = new SourceEmitterContext(
-            descriptor.Symbol.IsStatic,
+            descriptor.Symbol.IsStatic || instanceWithStaticPartialMethod,
```
latonz, line 21: Instead of using this additional boolean parameter, why not just store it inside the descriptor as a new property (eg. `public bool Static { get; set; }`)?

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -111,9 +113,18 @@ private void ExtractObjectFactories()
...
     }
 
-    private void ExtractUserMappings()
+    private ExtractUserMappingReport ExtractUserMappings()
     {
-        foreach (var userMapping in UserMethodMappingExtractor.ExtractUserMappings(_builderContext, _mapperDescriptor.Symbol))
+        var extractUserMappingReport = new ExtractUserMappingReport();
+
+        foreach (
+            var userMapping in UserMethodMappingExtractor.ExtractUserMappings(
```
latonz, line 121: IMO the code could be simplified a lot if the checks happen here instead of inside of `ExtractUserMappings`. Also I think logically it isn't part of a single user mapping method extraction but more of a validation over all user mapping methods which would belong to `DescriptorBuilder` instead of `UserMethodMappingExtractor`.

Together with a boolean property on the descriptor (see other comment) this would allow to keep this method void, remove the added `ExtractUserMappingReport` class and remove the additional return value / parameter passed through `MapperGenerator`.

test/Riok.Mapperly.Tests/Mapping/InstantiableMapperWithStaticMethodsTest.cs
latonz: I think you misunderstood me here, my bad for not being enough specific...
What I actually meant is a test with a sth. like this:
Source: `record A(B Value); record B(int Value);`
Target: `record C(D Value); record D(int Value);`

Related to what I was trying to describe [here](https://github.com/riok/mapperly/pull/681#discussion_r1384755733).
A second test case could be like this (to address https://github.com/riok/mapperly/pull/681#discussion_r1384755733): 
Source: `record A(B Value); record B(C Value); record C(int Value)`
Target: `record D(E Value); record E(F Value); record F(int Value)`
With a user implemented mapping from C to F.

latonz (changes requested):

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
trejjam: I use it because it compiles generated code -> discover instance method calls from static methods. But I can move it to unit test also

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
```diff
@@ -51,7 +82,7 @@ bool isStatic
         return BuildUserImplementedMappings(ctx, methods, receiver, isStatic);
     }
 
-    private static IEnumerable<IMethodSymbol> ExtractMethods(ITypeSymbol mapperSymbol) => mapperSymbol.GetMembers().OfType<IMethodSymbol>();
+    public static IEnumerable<IMethodSymbol> ExtractMethods(ITypeSymbol mapperSymbol) => mapperSymbol.GetMembers().OfType<IMethodSymbol>();
```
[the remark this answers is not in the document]
trejjam: Not anymore, I was using it in dropped `IsInstanceWithStaticPartialMethod`

src/Riok.Mapperly/Emit/SourceEmitter.cs
trejjam: Make sense, I do not know why I chose `instanceWithStaticPartialMethod`

test/Riok.Mapperly.Tests/Mapping/InstantiableMapperWithStaticMethodsTest.cs
trejjam: Thank you for explaining. Both test cases added

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
trejjam: Do you want me to drop this part?
Unit tests were updated

src/Riok.Mapperly/Emit/SourceEmitter.cs
```diff
@@ -14,7 +14,7 @@ public static class SourceEmitter
     public static CompilationUnitSyntax Build(MapperDescriptor descriptor, CancellationToken cancellationToken)
     {
         var ctx = new SourceEmitterContext(
-            descriptor.Symbol.IsStatic,
+            descriptor.Symbol.IsStatic || descriptor.Static,
```
latonz, line 17: The first part is not needed, is it? Could be simplified to `descriptor.Static`.

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -123,9 +127,29 @@ private void ExtractUserMappings()
                 userMapping.TargetType
             );
 
+            if (userMapping.Method is { IsStatic: true, IsPartialDefinition: true })
```
latonz, line 130: Is the partial definition part really required? IMO also user implemented mapping methods need to be considered, don't they?

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -123,9 +127,29 @@ private void ExtractUserMappings()
...
+            {
+                _mapperDescriptor.Static = true;
+            }
+            else if (firstNonStaticUserMapping is null && userMapping.Method is { IsStatic: false, MethodKind: not MethodKind.Constructor })
```
latonz, line 134: Is the second part of the condition really required? This can't ever be a constructor, can it? `ExtractUserMappings` should never return a ctor. Also if the if condition is simplified to `userMapping.Method.IsStatic` the else if could be simplified to `firstNonStaticUserMapping == null` since the static part is already ensured to be false by the if condition.

test/Riok.Mapperly.IntegrationTests/Mapper/TestMapperWithStaticMethods.cs
```diff
@@ -0,0 +1,39 @@
...
+        private static partial TestObjectDto MapToDtoInternal(TestObject testObject);
+
+        public static partial IEnumerable<TestObjectDto> MapAllDtos(IEnumerable<TestObject> objects);
+
+        public static TestObjectDto MapToDto(TestObject src)
+        {
+            var target = MapToDtoInternal(src);
+            target.StringValue += "+after-map";
+            return target;
+        }
```
latonz, line 37: I think this test case can be removed as it is unit tested and it also covered in other integration tests. Sorry for the misunderstanding.

latonz (changes requested):
Thank you for the updates, only a few comments left and this can get merged 👍

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
latonz: Now this is unit tested quite well and Mapperly runs the same code for all Roslyn versions, therefore I don't see a need to keep this additional integration test: I'd drop it.

test/Riok.Mapperly.IntegrationTests/Mapper/TestMapperWithStaticMethods.cs
latonz: Just saw the older, still open, discussion regarding this.

latonz:
I just checked my earlier comment https://github.com/riok/mapperly/pull/681#issuecomment-1697329056 and noticed that there is still a problem with object factories and external mappers. In pseudo-static mappers, instance object factories and external mappers are still beeing used without a diagnostic. These should be ignored and a diagnostic should be emitted. `ExtractObjectFactories` can probably be moved after `ExtractUserMappings`. Then `_mapperDescriptor.Static` can be used in `ExtractUserMappings` and `ExtractObjectFactories`.

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
trejjam: You were right with ctor, the condition was left there from moment where I used `ExtractMethods` directly

trejjam: I think the condition cannot be simplified due to `InstanceMapperShouldSupportUserDefinedStaticMethods` mention above

trejjam:
I will take a look at the factories and external mappers

trejjam:
I dropped one test which I would say did not test what it was saying so: https://github.com/riok/mapperly/pull/681/commits/2b6db69caa8064ae9ad45da98566cf8e891f1707

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
trejjam: I dropped this integration test

trejjam:
I added tests for [external mappers](https://github.com/riok/mapperly/pull/681/commits/3f907b4c3421152d503a0ab13715826fffd12c3b) and [object factories](https://github.com/riok/mapperly/pull/681/commits/44ac9ff022ab9546aaced6fde4bf49a2cb6a7316)

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
latonz: Ah you are correct I didn't thought about that one.

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -63,8 +64,8 @@ MapperConfiguration defaultMapperConfiguration
     {
         ConfigureMemberVisibility();
         ReserveMethodNames();
-        ExtractObjectFactories();
         ExtractUserMappings();
+        ExtractObjectFactories();
```
latonz, line 68: As the order has changed here, the documentation in `architecture.md` needs to be adjusted too.

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -123,9 +127,29 @@ private void ExtractUserMappings()
                 userMapping.TargetType
             );
 
+            if (userMapping.Method is { IsStatic: true, IsPartialDefinition: true })
+            {
+                _mapperDescriptor.Static = true;
+            }
+            else if (firstNonStaticUserMapping is null && userMapping.Method is { IsStatic: false })
```
latonz, line 134: Nit: For single property conditions and null checks this code base usually doesn't use pattern matching: `firstNonStaticUserMapping == null && !userMapping.Method.IsStatic`

src/Riok.Mapperly/Descriptors/ObjectFactories/ObjectFactoryBuilder.cs
```diff
@@ -28,6 +28,7 @@ public static ObjectFactoryCollection ExtractObjectFactories(SimpleMappingBuilde
             || methodSymbol.IsPartialDefinition
             || methodSymbol.MethodKind != MethodKind.Ordinary
             || methodSymbol.ReturnsVoid
+            || (methodSymbol.IsStatic is false && ctx.Static)
```
latonz, line 31: Nit: For simple conditions this code base usually doesn't use pattern matching: `(!methodSymbol.IsStatic && ctx.Static)`

test/Riok.Mapperly.Tests/_snapshots/InstantiableMapperWithStaticMethodsTest.ShouldUseSimpleObjectFactory#CarMapper.g.verified.cs
```diff
@@ -0,0 +1,12 @@
+﻿//HintName: CarMapper.g.cs
+// <auto-generated />
+#nullable enable
+public partial class CarMapper
+{
+    static partial global::B Map(global::A a)
+    {
+        var target = new global::B();
```
latonz, line 8: This doesn't look like what is expected, does it? The name of the test is `ShouldUseSimpleObjectFactory`. IMO this should be `var target = CreateB();`

test/Riok.Mapperly.Tests/_snapshots/ReferenceHandlingTest.CustomHandlerWithObjectFactoryShouldWork#Mapper.g.verified.cs
```diff
@@ -7,7 +7,7 @@ public partial class Mapper
     {
         if (refHandler.TryGetReference<global::A, global::B>(a, out var existingTargetReference))
             return existingTargetReference;
-        var target = CreateB();
+        var target = new global::B();
```
latonz, line 10: IMO this shouldn't change. That's also why a lot of unit tests fail.
The problem is that when the context (`MappingBuilderContext`) for a user mapping is created the object factories are still empty and not yet extracted. This could probably be fixed by making `ObjectFactoryCollection` mutable and removing `ObjectFactoryCollection.Empty` as well as the constructor with the collection and instead adding a public method to add entries. I'm not super happy making the object factory collection mutable but can't currently think of a better solution. Do you have a better idea?

latonz (changes requested):

test/Riok.Mapperly.Tests/_snapshots/ReferenceHandlingTest.CustomHandlerWithObjectFactoryShouldWork#Mapper.g.verified.cs
trejjam: Thanks for spotting. My eyes saw a different change `new B -> new global::B`

trejjam: I solved it by dropping global `_objectFactories`. My attempt to turn it to mutable affected the test state and some tests started to fail randomly.

image

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -106,24 +108,59 @@ private void ReserveMethodNames()
         }
     }
 
-    private void ExtractObjectFactories()
+    private IEnumerable<IUserMapping> ExtractUserMappings()
```
latonz, line 111: IMO it would be easier to just use `_mappings.UserMappings` in `EnqueueToBuildBody` instead yield return + ToList and passing as variable.

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -106,24 +108,59 @@ private void ReserveMethodNames()
...
     }
 
-    private void ExtractUserMappings()
+    private ObjectFactoryCollection ExtractObjectFactories()
     {
-        foreach (var userMapping in UserMethodMappingExtractor.ExtractUserMappings(_builderContext, _mapperDescriptor.Symbol))
+        return ObjectFactoryBuilder.ExtractObjectFactories(_builderContext, _mapperDescriptor.Symbol);
+    }
+
+    private void EnqueueToBuildBody(IReadOnlyCollection<IUserMapping> userMappings, ObjectFactoryCollection objectFactoryCollection)
```
latonz, line 152: I'd rename to `EnqueueUserMappings` (IMO that reflects better what happens here). Probably also simplify `objectFactoryCollection` to `objectFactories`.

test/Riok.Mapperly.Tests/_snapshots/ReferenceHandlingTest.CustomHandlerWithObjectFactoryShouldWork#Mapper.g.verified.cs
latonz: I really like this approach and it can even be simplified further (instead of returning / passing the user mappings `_mappings.UserMappings` could be used).

latonz (changes requested):

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
trejjam: I am not sure about this collection, it can be populated also from `ExtractExternalMappings` which should not be passed into `EnqueueUserMappings`.

This collection is populated after the `EnqueueUserMappings` call so it's not an issue right now. If you want to use it I can

trejjam: It's inside https://github.com/riok/mapperly/pull/681/commits/d931e58fb1758e0fb387ba307711037f8685ab40

latonz (approved):

PR #700, opened 2023-08-28
feat: add strict mapping

### LINKED ISSUE #687: MemberList option to enforce one-sided strict mappings
Rel. https://github.com/riok/mapperly/discussions/677

Add an option to specify whether the relevant members are on the source, the target, both or none. This makes it possible to work with strict mappings when only one side of a mapping is relevant, without ignoring lots of properties.
For example when mapping from an entity with 10 properties to a public DTO which only has 2 properties, one want to make sure the two properties of the DTO are mapped and enables strict mappings. However, only the target properties are relevant for the strict mappings.
The idea is to add a configuration property on the mapper attribute and on a method level via an attribute.

## API

```diff
public sealed class MapperAttribute : Attribute
{
+   public MemberList MemberList { get; set; } = MemberList.Both;
}

+[AttributeUsage(AttributeTargets.Method)]
+public sealed class MemberListAttribute : Attriubte
+{
+    public MemberListAttribute(MemberList memberList) {}
+}

+[Flags]
+public enum MemberList
+{
+    None = 0,
+    Both = ~None,
+    Source = 1 << 0,
+    Target = 1 << 1,
+}
```

If the `Source` flag is not set, `RMG020` and `RMG038` should not be emitted.
If the `Target` flag is not set, `RMG012` and `RMG037` should not be emitted.

TimothyMakkison:
# Add strict mapping support

## Description
Disable the unmapped source/target member warning on a mapper or method level. ie `[MapperAllowUnmapped(AllowUnmapped.Source)]` will not warn about unmapped source members and will still automatically map members.

Currently using `AllowUnmapped` as a placeholder. I haven't done integration tests, documentation, xml docs etc.

Fixes #687 / #677

## Checklist

### DIFF
--- a/test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
+++ b/test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
@@ -14,7 +14,8 @@ public record TestSourceBuilderOptions(
     MappingConversionType? EnabledConversions = null,
     EnumMappingStrategy? EnumMappingStrategy = null,
     bool? EnumMappingIgnoreCase = null,
-    IgnoreObsoleteMembersStrategy? IgnoreObsoleteMembersStrategy = null
+    IgnoreObsoleteMembersStrategy? IgnoreObsoleteMembersStrategy = null,
+    RequiredMappingStrategy? RequiredMappingStrategy = null
 )
 {
     public const string DefaultMapperClassName = "Mapper";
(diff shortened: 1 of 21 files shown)


latonz:
See my comments in https://github.com/riok/mapperly/issues/687.
The same also needs to be done for enums, probably in a separate MR.

latonz:
@TimothyMakkison any updates on this? Are you still working on this?  😊 Don't worry but otherwise I'd probably schedule this feature to be done by someone else...

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
```diff
@@ -26,6 +27,9 @@ public record TestSourceBuilderOptions(
     public static TestSourceBuilderOptions WithIgnoreObsolete(IgnoreObsoleteMembersStrategy ignoreObsoleteStrategy) =>
         new(IgnoreObsoleteMembersStrategy: ignoreObsoleteStrategy);
 
+    public static TestSourceBuilderOptions WithMemberList(RequiredMapping requiredMappingStrategy) =>
```
latonz, line 30: I'd use the same name for this method as for the functionality itself `WithRequiredMapping`

latonz:
@TimothyMakkison thanks for the updates 😊 A few smaller feedback points, looks good otherwise, naming works with me 👍

PR #703, opened 2023-08-30
feat: improve incremental generator

latonz:
# Make better use of incremental source generator pipeline

## Checklist


TimothyMakkison:
Nice, haven't read through the changes yet but I did have some similar plans.

### Symbol accessor
When nested private mapping was added, `SymbolAccessor` was changed to be per mapper instead of being shared. This was because it stores the known visible members of types and this could now be different between mappers.

I thought it would be possible to create a shared `SharedSymbolAccessor` to store expensive information that would not vary between mapper context. I did create a branch but presumably messed up as I couldn't see any benefit 😕

### Remove Normalize Whitespace
I have a branch where I stopped using `Expression` and `NormalizeWhitespace` as the latter has dreadful performance (legit most our memory and time usage)

### Make all `Mapping` equitable
With a little work it will be possible to remove all usage of `ITypeSynbol` making it easier to make all types equatable so another step in the pipeline could be added.


Would love to hear what you think of these

ni507 (dismissed):

src/Riok.Mapperly/Configuration/MapperConfiguration.cs
```diff
@@ -9,23 +9,25 @@ namespace Riok.Mapperly.Configuration;
 /// Newly added properties must also be added to <see cref="Riok.Mapperly.Helpers.MapperConfigurationBuilder"/>.
```
ni507, line 9: Rename reference of `MapperConfigurationBuilder`

ni507: It should be `MapperConfigurationMerger`

latonz:
@TimothyMakkison

## SymbolAccessor
You are right, but I think in real world scenarios there are probably not a lot of members which are shared between mappers 🤔 

## Remove Normalize Whitespace
What do you mean by stopped using Expression? Switched it for blocks? I just experimented with a msbuild settings to disable normalize whitespace [here](https://github.com/latonz/mapperly/tree/configurable-normalize-whitespace).

## Make all `Mapping` equitable

I experimented with this in the past... I think this would be a nice improvement, however the last time I experimented with it I came to the conclusion that the added complexity is not worth it, but we may revisit it...

src/Riok.Mapperly/Configuration/MapperConfiguration.cs
latonz: Fixed2 😊

ni507 (approved):

TimothyMakkison:
> What do you mean by stopped using Expression? Do you mean using string based code generation instead of Microsoft.CodeAnalysis.CSharp.Syntax? I just experimented with a msbuild settings to disable normalize whitespace [here](https://github.com/latonz/mapperly/tree/configurable-normalize-whitespace) but it would need a lot more work.

Most other source generators don't use `ExpressionSyntax`, I copied `SourceWriter` from AOT `System.Text.Json`.

```C#
public override ExpressionSyntax Build(TypeMappingBuildContext ctx)

// instead I return a string
public override string Build(TypeMappingBuildContext ctx)

public override IEnumerable<StatementSyntax> BuildBody(TypeMappingBuildContext ctx)

// BuildBody uses ctx.SourceWriter to build the text.
public override void BuildBody(TypeMappingBuildContext ctx)
```

Interesting idea with disablig `NormalizeWhitespace` not sure if people would use it much.

> I experimented with this in the past... I think this would be a nice improvement, however the last time I experimented with it I came to the conclusion that the added complexity is not worth it, but we may revisit it...

Yeah, each mapper would have to implement some equality method and all future mappers couldn't use any `ISymbol` types.

latonz:
@TimothyMakkison when we started the project we thought about using strings directly instead, however, since the generated code can get quite complex we preferred the stronger typed syntax objects.

Issue #642, opened 2023-08-10
`[MapProperty]` on a nested target prevents the root member being resolved from the source.


TimothyMakkison:
**Is your feature request related to a problem? Please describe.**
Using MapProperty to assign a nested value will prevent the root member being auto resolved from the source.

```C#
[Mapper]
public static partial class Mapper
{
    [MapProperty("NestedValue", "Nested.Value")]
    public static partial B Map(A source);
}

public class A { public C Nested { get; } public string NestedValue { get; } }
public class B { public C Nested { get; set; } }
public class C { public string Value { get; set; } }

// here target.Nested is not assigned source.Nested
// removing [MapProperty] fixes this
var target = new global::Riok.Mapperly.Sample.B();
target.Nested.Value = source.NestedValue;
return target;
```

**Describe the solution you'd like**
`ObjectMemberMappingBodyBuilder.BuildMappingBody` should try to resolve the member if its configs don't map to the target. 

I'm not sure how error handling should occur. i.e. If a specified root assignment is invalid, should a resolution be attempted?

latonz:
This is expected with the current implementation. The current implementation marks all target properties as mapped, which are part of a target property mapping path. Not sure if we should change that.

TimothyMakkison:
IMO this is pretty confusing. You might have a functioning mapping, but by adjusting a child it stops working. Does automapper/mapster do this?

latonz:
On a second thought we should probably improve this and map such members unless they are explicitly ignored.

TimothyMakkison:
👍 think I have the fix already made, see if I can find it

PR #705, opened 2023-08-30
fix: remove unneeded null conditional initializer

### LINKED ISSUE #640: `AddNullMemberInitializers` should check if a member is assigned before instantiating with `??= new()`
When mapping paths, `MapProperty` initializes parents of members using `??= new()` to prevent exceptions. This doesn't check if the member has already been set resulting in redundant code.

```C#
[MapProperty("NestedValue", "Nested.Value")]
[MapProperty("Parent", "Nested")] 
partial B Map(A source);

class A { public string NestedValue { get; init; } public C Parent { get; set; } }
class B { public C? Nested { get; set; } }
class C { public string Value { get; set; } }

// Generates
var target = new global::B();
target.Nested ??= new(); // redundant
target.Nested = source.Parent;
target.Nested.Value = source.NestedValue;
return target;

// Should generate
var target = new global::B();
target.Nested = source.Parent;
target.Nested.Value = source.NestedValue;
return target;
```

`AddNullMemberInitializers` should check if the member is already set before creating a `MemberNullAssignmentInitializerMapping`.

TimothyMakkison:
## Description
Add check to only insert a null initializer if the root value is uninitialized or initialized by a nullable type.

tbh I feel like mapperly shouldn't insert a null initializer if the root is mapped. Having a nullable path seems like an error and this obscures it.

Partially fixes #640

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersContainerBuilderContext.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersContainerBuilderContext.cs
@@ -27,7 +27,7 @@ public class MembersContainerBuilderContext<T> : MembersMappingBuilderContext<T>
 
     private void AddMemberAssignmentMapping(IMemberAssignmentMappingContainer container, IMemberAssignmentMapping mapping)
     {
-        SetSourceMemberMapped(mapping.SourcePath);
+        SetMembersMapped(mapping);
         AddNullMemberInitializers(container, mapping.TargetPath);
         container.AddMemberMapping(mapping);
     }
@@ -40,6 +40,10 @@ public class MembersContainerBuilderContext<T> : MembersMappingBuilderContext<T>
             var type = nullablePath.Member.Type;
 
             if (!nullablePath.Member.CanSet)
+                continue;
+
+            // if member is initialised with a non null value then skip
+            if (NotNullInitializedTargetMemberNames.Contains(path.Path.First().Name))
                 continue;
 
             if (!BuilderContext.SymbolAccessor.HasAccessibleParameterlessConstructor(type))
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
@@ -1,6 +1,7 @@
 using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Configuration;
 using Riok.Mapperly.Descriptors.Mappings;
+using Riok.Mapperly.Descriptors.Mappings.MemberMappings;
 using Riok.Mapperly.Diagnostics;
 using Riok.Mapperly.Helpers;
 using Riok.Mapperly.Symbols;
@@ -72,6 +73,8 @@ public abstract class MembersMappingBuilderContext<T> : IMembersBuilderContext<T
 
     public Dictionary<string, List<PropertyMappingConfiguration>> MemberConfigsByRootTargetName { get; }
 
+    public readonly HashSet<string> NotNullInitializedTargetMemberNames = new();
+
     public void AddDiagnostics()
     {
         AddUnmatchedIgnoredTargetMembersDiagnostics();
@@ -82,7 +85,19 @@ public abstract class MembersMappingBuilderContext<T> : IMembersBuilderContext<T
         AddMappedAndIgnoredTargetMembersDiagnostics();
     }
 
-    protected void SetSourceMemberMapped(MemberPath sourcePath) => _unmappedSourceMemberNames.Remove(sourcePath.Path.First().Name);
+    protected void SetMembersMapped(IMemberMapping mapping)
+    {
+        _unmappedSourceMemberNames.Remove(mapping.SourcePath.Path.First().Name);
+    }
+
+    protected void SetMembersMapped(IMemberAssignmentMapping assignmentMapping)
+    {
+        _unmappedSourceMemberNames.Remove(assignmentMapping.SourcePath.Path.First().Name);
+        if (!assignmentMapping.TargetType.IsNullable())
+        {
+            NotNullInitializedTargetMemberNames.Add(assignmentMapping.TargetPath.Path.First().Name);
+        }
+    }
 
     private HashSet<string> InitIgnoredUnmatchedProperties(IEnumerable<string> allProperties, IEnumerable<string> mappedProperties)
     {
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/NewInstanceBuilderContext.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/NewInstanceBuilderContext.cs
@@ -15,14 +15,14 @@ public class NewInstanceBuilderContext<T> : MembersMappingBuilderContext<T>, INe
 
     public void AddInitMemberMapping(MemberAssignmentMapping mapping)
     {
-        SetSourceMemberMapped(mapping.SourcePath);
+        SetMembersMapped(mapping);
         Mapping.AddInitMemberMapping(mapping);
     }
 
     public void AddConstructorParameterMapping(ConstructorParameterMapping mapping)
     {
         MemberConfigsByRootTargetName.Remove(mapping.Parameter.Name);
-        SetSourceMemberMapped(mapping.DelegateMapping.SourcePath);
+        SetMembersMapped(mapping.DelegateMapping);
         Mapping.AddConstructorParameterMapping(mapping);
     }
 }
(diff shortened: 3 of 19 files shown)


src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersContainerBuilderContext.cs
```diff
@@ -42,6 +42,17 @@ private void AddNullMemberInitializers(IMemberAssignmentMappingContainer contain
             if (!nullablePath.Member.CanSet)
                 continue;
 
+            // if member is initialised with a non null value then skip
+            if (
```
latonz, line 46: Wouldn't it be easier to keep a set of mapped source members (in `SetSourceMemberMapped`) and perform a simple lookup here? This way we wouldn't need to make the information public on the mapping.

latonz (changes requested):
Thank you for this PR 😊

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersContainerBuilderContext.cs
TimothyMakkison: > Wouldn't it be easier to keep a set of mapped source members (in SetSourceMemberMapped) and perform a simple lookup here? 

Tried something similar but realised it would cause errors: if the source value is transformed we'd lose the type info ie `target.Value = UserDefinedDoggyMapper(source.Value)`

Yeah, I'm not a fan of exposing the properties either.

latonz: But even in this case, we shouldn't initialize the property afterwards, should we? Wouldn't this be strange? It probably is the case today, but I don't think this makes a lot of sense.

TimothyMakkison: I'm not sure what you mean, could you rephrase this please?

FYI this PR doesn't change the order of member instantiation. It is still possible to null initialise before a member is set. All this PR does is omit a null initializer if the source value is not nullable.

latonz: Sorry for my bad english... I thought Mapperly should not generate a `??= new();` if previously another mapping assigned a value to this property. However, on a second thought I think we should keep the current behaviour. But isn't the described approach possible by simply adding a parameter to a `Set...MemberMapped` method, whether the member is nullable or not?

TimothyMakkison: > I thought Mapperly should not generate a `??= new();` if previously another mapping assigned a value to this property. However, on a second thought I think we should keep the current behaviour. 

What do you think we should do?
- Initialise if source value is nullable
- Initialise if member is never assigned

The latter is "technically" a breaking change although I'd argue hiding null exceptions is more confusing.

Tbh I'm not a fan of the null initializer. I'd argue it causes more errors than it solves, is hardly used as it requires an empty initializer and is useless half the time if the member is assigned to.

This pr and #642 should address some of my complaints and perhaps I'm overlooking an area where it is useful. I wonder if a warning for unitialized sections could be added.

> But isn't the described approach possible by simply adding a parameter to a `Set...MemberMapped` method, whether the member is nullable or not?

I see what you mean, I'll try it out 😃

latonz: I completely understand your point, but for now I'd stick with the current behaviour. We can probably address this with the next major release/breaking change set.

TimothyMakkison: > I'd stick with the current behaviour. 

Okay, I'll only remove the null initializer if the source value cannot be null. 

I'm not sure how `SetSourceMemberMapped` could be used here. Adding a parameter for nullable values to it is weird as there is currently no way of getting that information without exposing the `IMemberMapping`, in which case why not just use my current approach and get the information when needed? 

Not sure if I'm missing something here 🤔  Sorry in advance

latonz: Oh sorry, I just realised SetSourceMemberMapped doesn't help anything here 🙄 It is the target which is relevant...

Quickly scanned through the code, I'm thinking of sth. like this:
* Rename `SetSourceMemberMapped` to `SetMembersMapped`
* Have to overloads of `SetMembersMapped`, one with a `IMemberMapping` parameter, one with a `IMemberAssignmentMapping` parameter
* Remove the source member path from `_unmappedSourceMemberNames` in both overloads
* Move `BodyBuildingPriority` from `IMapping` to `ITypeMapping`
* Make `IMemberMapping` and `IMemberAssignmentMapping` implement `IMapping` by returning the source/target type from their delegate mappings
* In `SetMembersMapped(IMemberAssignmentMapping)` add the target member path to a set of non-null initialized members if the target type of the mapping is non-nullable
* In `AddNullMemberInitializers` check if the member is already in the set of the initialized members.

IMO it is a little more work to be done but I think it could be worth it as the solution seems cleaner to me.
WDYT about such an approach?

latonz:
@TimothyMakkison any updates on this? Are you still working on this?  😊 Don't worry but otherwise I'd probably schedule this feature to be done by someone else...

TimothyMakkison:
> @TimothyMakkison any updates on this? Are you still working on this? 😊 Don't worry but otherwise I'd probably schedule this feature to be done by someone else...

Really sorry, should be able to fix this and the unmapped property warning PR. I'll need to install the preview for Unsafe accessors

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersContainerBuilderContext.cs
TimothyMakkison: > WDYT about such an approach?

I've made this change and unless I made a mistake it doesn't appear to work. By the looks of it `(IMapping)IMemberAssignmentMapping` will return the non nullable source and target types. This is because the internal `NewInstanceMapping` is created at line 239 of `ObjectMemberMappingBodyBuilder` from the non nullable types, the actual nullablility is handled by the wrapping `IMemberMapping`.

latonz: Have you pushed it in a branch so I can quickly look into it? 😊 Thank you!

TimothyMakkison: I swear I pushed the changes last night

TimothyMakkison: > Have you pushed it in a branch so I can quickly look into it? 😊 Thank you!

This branch should have the changes as commit [99afa22](https://github.com/riok/mapperly/pull/705/commits/99afa22385ccabfc8fc1f338147a830e592a848e). Do you

latonz: Thanks for the commit. I tried a little around on the provided commit and came up with the following solution (available [here](https://github.com/riok/mapperly/commit/783587c456a2ab2aae4b2f1a82faa38fabfbbc72)):
* I exposed a `MemberTypeMapping` from the `IMemberAssignmentMapping`
* `SetTargetNonNullInitializedIfNonNullable` added to `MembersContainerBuilderContext` to keep track of the non-nullable initialized target paths

With this the correct nullable initializers were applied. However, under some conditions the initializers were emitted after a nested mapping. I fixed this with:
* Introducing `IMemberAssignmentMappingOrContainer` to have a common interface for `IMemberAssignmentMappingContainer` and `IMemberAssignmentMapping` 
* Introducing an additional `List<IMemberAssignmentMappingOrContainer>` to keep the containers and mappings ordered
* Adjusted the `MemberAssignmentMappingContainer.Build` to use the ordered list as source

I'm not super happy with this solution, but IMO it still exposes less information/complexity than `TryGetMemberMapping`. WDYT?

TimothyMakkison: Sound good 👍 
Does this mean this PR is no longer needed!?

latonz: Since I'm not really happy with the solutions discussed (and you don't seem to be either), I'm thinking about whether we should just remove the initialisation of these nullable values with the next major release/breaking change.

PR #711, opened 2023-09-01
feat: remove the Riok.Mapperly.Abstractions runtime dependency

### LINKED ISSUE #685: Remove `Riok.Mapperly.Abstractions` runtime dependency
Rel. https://github.com/riok/mapperly/discussions/666.

Mapperly provides configuration attributes in `Riok.Mapperly.Abstractions`. For reference handling, a default implementation of `IReferenceHandler` is also provided in `Riok.Mapperly.Abstractions` as `PreserveReferenceHandler`. The `PreserveReferenceHandler` is the only runtime dependency of the code generated by Mapperly (only if reference handling is enabled).
If reference handling is not used `PrivateAssets="all" ExcludeAssets="runtime"` can be set which eliminates the need of `Riok.Mapperly.Abstractions` as a runtime dependency.
We should improve the documentation on this (document `PrivateAssets="all" ExcludeAssets="runtime"` on the installation page) and try to eliminate the runtime dependency. Sadly we cannot set `DevelopmentDependency=true`, since this does not work for source generators.
Ideas on how to eliminate the runtime dependency on `Riok.Mapperly.Abstractions`:
* Emit the `PreserveReferenceHandler` class from the source generator if reference handling is enabled
  * If the class is emitted as an internal class in a Mapperly namespace, this could lead to duplicated symbols when `InternalsVisibleTo` is set
  * The class could be emitted as a file scoped class. Mapperly would need to generate the class for each mapper using reference handling and would require C# language version 11.
  * The class could be emitted in a namespace which is dependent on the name of the assembly.
* Create a new NuGet package `Riok.Mapperly.ReferenceHandling` which a user needs to reference when enabling reference handling.

latonz:
# remove the Riok.Mapperly.Abstractions runtime dependency

## Description

Creates a new Riok.Mapperly.Templates project with templates. Some Mapperly features may require such a template (currently only the reference handling). If such a template is required by a used Mapperly feature, Mapperly emites the template in a namespace which includes the name of the assembly of the target project. This is done to prevent collisions when using `InternalsVisibleTo` (see #685).
Currently the only template is the PreserveReferenceHandler which is emitted if reference handling is enabled for any mapper and the mapper does have mapper method definitions without a reference handling parameter and Mapperly needs to instantiate a IReferenceHandler.

The runtime dependency can be removed by setting `PrivateAssets="all" ExcludeAssets="runtime"` on the package reference. Unluckily we cannot set `developmentDependency=true` on the NuGet, since this would exclude the compile assets which are required for a source generator. See also https://github.com/NuGet/Home/issues/10746.

Fixes #685

## Checklist


samples/Riok.Mapperly.Sample/Riok.Mapperly.Sample.csproj
```diff
@@ -2,8 +2,6 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net7.0</TargetFramework>
```
CommonGuy, line 5: I would leave this in here to make it easier for others to copy/view this project

docs/docs/getting-started/installation.mdx
```diff
@@ -12,26 +12,36 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
...
 Install-Package Riok.Mapperly
 ```
 
+Make sure to set `PrivateAssets="all" ExcludeAssets="runtime"` on the added `PackageReference`.
+
 </TabItem>
 </Tabs>
+
+:::info
+`PrivateAssets="all"` ensures projects referencing this project will not also get a reference to `Riok.Mapperly`.
```
CommonGuy, line 44: Would change this to 
```
`PrivateAssets="all"` ensures that projects referencing this project do not also get a reference to "Riok.Mapperly".
`ExcludeAssets="runtime"` ensures that the Mapperly .dll files are not copied to the build output (they are not needed at runtime).
```

latonz: Adjusted

samples/Riok.Mapperly.Sample/Riok.Mapperly.Sample.csproj
latonz: The csproj is not easy copyable at all due to the Mapperly references beeing `ProjectReference` instead of `PackageReference`. Due to that fact I think we should keep the project file as maintainable and as simple as possible for us.

CommonGuy (approved):

PR #732, opened 2023-09-06
feat: C#12 `UnsafeAccessor` for private properties and fields

### LINKED ISSUE #600: .NET 8 zero overhead private member mapping
Previously, Mapperly hasn't had the option to access or set private members. Although private/hidden members have always been visible to source generators, there hasn't been a suitable way of accessing them. While workarounds such as `Reflection.Emit`, `Linq.Expressions` and reflection are available, upon closer inspection, they are insufficient for our needs due to poor performance and incompatibilities in AOT usage. For these reasons, Mapperly has never added private member mapping.

With the release of .NET 8, the [`UnsafeAccessorAttribute`](https://github.com/dotnet/runtime/issues/81741) will be added. This supports code that accesses internal methods, constructors, fields, and properties with zero overhead while being AOT compatible. By applying it to an extern static method and configuring it, the runtime will attempt to find the corresponding field or method, to which the call will be forwarded.

Using the new attribute, Mapperly could add support for private member mapping, init-only/private setter mapping, and support private constructors.

### Enabling
Private member mapping could be enabled by default, although this would likely lead to nasty unexpected behaviour while breaking backwards compatibility. Instead I suggest that a property `EnablePrivateMapping` be added to `[Mapper]` and a method attribute `[EnablePrivateMapping]` be added. Alternatively private member mapping could always be enabled but only for explicit `MapProperty` mappings.

### Generated code
The private access methods can be added as additional methods to the mapping class. These could be implemented like a `MethodMapping` and used like so: `SetId(target, GetId(source))`. 
Alternatively, to make the code easier to read, the private accessors could be added to a file scoped static accessor class. This class way the methods could be implemented as extensions methods, with the resulting code reading left to right. `target.SetId(source.GetId())`.

`target.Tires.SetSpareWheel(source.Tires.GetSpareWheel())` vs `SetSpareWheel(target.Tires, GetSpareWheel(source.Tires))`

### Example
**Mapper**
```C#
public class Car
{
    private Tire _spareWheel { get; set; }
    // ...
}

public class CarDto
{
    private TireDto _spareWheel { get; set; }
    // ...
}

{
    [MapProperty("_spareWheel", "_spareWheel")]
}
```
**Generated Code**
```C#
static file class 
{
    [UnsafeAccessor(UnsafeAccessorKind.Method, Name="set__spareWheel")]
    public static extern Tire GetSpareWheel(this Car source)

    [UnsafeAccessor(UnsafeAccessorKind.Method, Name="set__spareWheel")]
    public static extern void SetSpareWheel(this CarDto target, TireDto tireDto)
}

{
    public static partial CarDto MapCarToDto(Car car)
    {
        target.SetSpareWheel(MapToTireDto(car.GetSpareWheel()));
        // ...
    }
}

```

Related #599, #531

### LINKED ISSUE #689: .NET 8.0, Roslyn 4.7 and C# 12 Support
An overview over all issues regarding .NET 8.0, Roslyn 4.7 and C# 12 support.

- [x] https://github.com/riok/mapperly/issues/682
- [x] https://github.com/riok/mapperly/issues/683
- [x] https://github.com/riok/mapperly/issues/684
- [x] https://github.com/riok/mapperly/issues/600

TimothyMakkison:
# C#12 `UnsafeAccessor` for private properties and fields

## Description
WIP for `UnsafeMemberAccessor`, logic is mostly sound but not cleaned up.

##### Ramblings/Thoughts
- Used `MethodMember` successfully 👍 
- Generated methods check the target type for conflicting member names.
- Should `MemberVisbility.Accessible` be removed and merged into `Public`, I have no idea why someone would want to only use it or need to disable it.
- `SetterMemberPath` needs the caller to check if `IsMethod` is true before either assigning to or invoking the result. (`target.Set_value(target.Get_value() ?? new())` vs `target.Value ??= new()`) I did consider creating two functions, or overloads. Not sure what is better. Will sleep on it.
- Need to check/decide how `private set;` should be handled.
- I suspect it is now possible to assign to `init` properties after construction. Not sure if this should be explored.
- Struggled to format the generated mapper and accessor classes so I added `SeparateByTrailingLineFeed` not sure if I missed a helper function.
- ~Probably need to add `using System.CompilerService;` somewhere.~
- Should the generated `static class UnsafeAccessor` appear before or after the code. Could the class names/generated extension names be improved?
- `SymbolAccessor` is initialized before `ConfigurationReader` so I pass in the `MemberVisibility` via `SetMemberVisibility`, this is hacky. I should probably create a `AccesibilityContext` class (just realied it would have to injected into `SymbolAccessor`. not sure what to do) 
- Not using the .Net 8 branch because I haven't installed the preview yet.
- Haven't added a .Net 8 constraint, don't know if I can get away with using `#if NET8_OR_GREATER` preprocessors everywhere, or if I should check the compilation.

### Update
- Use qualified namespace for attributes
- Prevented initializer from using a broken private setter function

Fixes #600

## Checklist


latonz (changes requested):
Thank you for this PR and sorry for the late review... Looks pretty good already 😊 Is it still in a draft state? I didn't do a complete review yet.

What do you think about https://github.com/riok/mapperly/issues/600#issuecomment-1706599722, I think such an approach would lead to cleaner and improved readability.

Addressing the questions:
> Need to check/decide how private set; should be handled.
   
   Would treat them the same as private fields
> I suspect it is now possible to assign to init properties after construction. Not sure if this should be explored.
  
I'd focus on other things for now and see if there is demand / a use case.

> Should the generated static class UnsafeAccessor appear before or after the code?
    
I'd put it after the code, as these are only accessors and seem to be less relevant.

> Could the class names/generated extension names be improved?

Probably upper-case the first char of the name of the target property in the accessor name?

> SymbolAccessor is initialized before ConfigurationReader so I pass in the MemberVisibility via SetMemberVisibility, this is hacky. I should probably create a AccesibilityContext class (just realied it would have to injected into SymbolAccessor. not sure what to do)

I'll try some things here and come back to you...

> Haven't added a .Net 8 constraint, don't know if I can get away with using #if NET8_OR_GREATER preprocessors everywhere, or if I should check the compilation.

I think it should be safe to check whether the `UnsafeAccessorAttribute` is available.

TimothyMakkison:
> Thank you for this PR and sorry for the late review... Looks pretty good already 😊 Is it still in a draft state? I didn't do a complete review yet.

 It is <em>nearish</em> to completion 😅 I need to add xml comments, docs, review my logic etc. Not a fan of how I've handled `MemberPath` or `SourceEmitter`. Looking back, I misunderstood what `Accessible` was meant to do, hence why my enum is so different, I'll fix this.

> > Should the generated static class UnsafeAccessor appear before or after the code?
> 
> I'd put it after the code, as these are only accessors and seem to be less relevant.

👍 

> > Could the class names/generated extension names be improved?
> 
> Probably upper-case the first char of the name of the target property in the accessor name?

How should a leading underscore `_` be handled? Debated stripping it or capitalising the next letter.

> > Haven't added a .Net 8 constraint, don't know if I can get away with using #if NET8_OR_GREATER preprocessors everywhere, or if I should check the compilation.
> 
> I think it should be safe to check whether the `UnsafeAccessorAttribute` is available.

I might create a spearate pr to help this. There are a couple of version specific methods that mapperly will try to use like `Enumerable.EnsureCapacity`, A class caching this information would preventing redundant, repeat checks, improving performance.

> What do you think about [#600 (comment)](https://github.com/riok/mapperly/issues/600#issuecomment-1706599722), I think such an approach would lead to cleaner and improved readability.

Thanks, it was really helpful. `MethodAccessorMember` was a massive help and simplified a lot of my logic. `ExpressionSyntax Build(ExpressionSyntax source, bool nullConditional = false)` does feel a little hacky, but I couldn't think of a nicer way to support getters, setters, method invocation and nullable paths. I deliberately created `GetterMemberPath` and `SetterMemberPath` due to the differences in getter/setter method logic and to prevent unwanted `UnsafeAccessors` from being added to the `descriptor`.

latonz:
> How should a leading underscore _ be handled? Debated stripping it or capitalising the next letter.

I'd strip it and capitalise the first letter after _.

TimothyMakkison:
I haven't read through my code yet but this PR is near completion. 
- Added integration test
- Check for .NET 8 if `Accessible` is disabled - emitting a warning and generating normal code by reenabling it.
- `UnsafeAccessor` code hopefully won't be emitted for `IQueryable` but I may have missed something; the accessibility code is pretty messy.

Unfortunately, the preview is still broken for me, so I;ve never been able to run code containing `UnsafeAccessor`. I have no idea if this even works 😆. I should probably address this with the integration tests

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/UnsafeAccess/UnsafeAccessSetProperty.cs
```diff
@@ -0,0 +1,59 @@
...
+        var targetName = nameBuilder.New("target");
+        var valueName = nameBuilder.New("value");
+
+        var target = Parameter(_targetType, targetName, true);
+        var value = Parameter(_valueType, valueName);
+
+        var parameters = ParameterList(CommaSeparatedList(target, value));
+        var attributeList = UnsafeAccessorMethods.UnsafeAccessorAttributeList(ctx, "Method", $"set_{_memberName}");
+
+        return MethodDeclaration(PredefinedType(Token(SyntaxKind.VoidKeyword)).AddTrailingSpace(), Identifier(MethodName))
```
TimothyMakkison, line 47: Could create a `UnsafeAccessorMethods`/`SyntaxFactoryHelper` method for this.

test/Riok.Mapperly.IntegrationTests/_snapshots/MapperTest.SnapshotGeneratedSource_NET8_0.verified.cs
```diff
@@ -464,4 +467,19 @@ private string MapToString1(global::Riok.Mapperly.IntegrationTests.Dto.TestEnumD
             return target;
         }
     }
+        
+    file static class UnsafeAccessor
```
TimothyMakkison, line 471: Ide says this should `static file class` unlike `public static class` 🤔

src/Riok.Mapperly/Symbols/IMappableMember.cs
```diff
@@ -22,7 +23,11 @@ public interface IMappableMember
 
     bool CanSet { get; }
 
+    bool CanAccessiblySet { get; }
+
     bool IsInitOnly { get; }
 
     bool IsRequired { get; }
+
+    ExpressionSyntax Build(ExpressionSyntax source, bool nullConditional = false);
```
latonz, line 32: IMO `BuildAccess` would be a better name here as this doesn't really build the member itself but only access to it.

src/Riok.Mapperly/Symbols/GetterMemberPath.cs
```diff
@@ -0,0 +1,82 @@
...
+public class GetterMemberPath : MemberPath
+{
+    private GetterMemberPath(IReadOnlyList<IMappableMember> path)
+        : base(path) { }
+
+    public static GetterMemberPath Build(MappingBuilderContext ctx, MemberPath memberPath)
+    {
+        var path = memberPath.Path;
+        var array = path.Select(item =>
+            {
```
latonz, line 19: IMO the readability could be improved if this is extracted into an own static method instead of a lambda.

test/Riok.Mapperly.Tests/Mapping/UnsafeAccessorTest.cs
```diff
@@ -0,0 +1,334 @@
+﻿using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+[UsesVerify]
+public class UnsafeAccessorTest
```
latonz, line 7: Add a test with an existing target object

src/Riok.Mapperly/Symbols/IMappableMember.cs
```diff
@@ -22,7 +23,11 @@ public interface IMappableMember
 
     bool CanSet { get; }
 
+    bool CanAccessiblySet { get; }
```
latonz, line 26: Add a comments what differentiates this from `CanSet`.

latonz, line 26: Probably rename to `CanSetDirectly` (see also comment on `SymbolAccessor`).

test/Riok.Mapperly.Tests/Mapping/UnsafeAccessorTest.cs
latonz, line 7: Add a test without unsafe accessors but with the member visibility filter (`IncludedMembers`).

latonz (changes requested):
Solid work 👍 Thank you!
I added my comments, feel free to discuss any provided feedback points 😊 
Please don't forget the documentation updates.

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/UnsafeAccess/UnsafeAccessField.cs
```diff
@@ -0,0 +1,61 @@
...
+        var targetName = nameBuilder.New("target");
+
+        var target = Parameter(_targetType, targetName, true);
+
+        var parameters = ParameterList(CommaSeparatedList(target));
+        var attributeList = UnsafeAccessorMethods.UnsafeAccessorAttributeList(ctx, "Field", _memberName);
+
+        return MethodDeclaration(
+                RefType(IdentifierName(_result).AddTrailingSpace())
+                    .WithRefKeyword(Token(TriviaList(), SyntaxKind.RefKeyword, TriviaList(Space))),
```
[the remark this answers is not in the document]
TimothyMakkison: Fields don't have setters in C#. It's possible to use `UnsafeAccessor` to get the value of a field without the `ref` keyword, however it wouldn't be possible to set it.

src/Riok.Mapperly/Symbols/SetterMemberPath.cs
```diff
@@ -0,0 +1,78 @@
...
+        return new SetterMemberPath(array, isMethod);
+    }
+
+    public ExpressionSyntax BuildAccess(ExpressionSyntax? baseAccess)
+    {
+        IEnumerable<IMappableMember> path = Path;
+
+        if (baseAccess == null)
+        {
+            baseAccess = SyntaxFactory.IdentifierName(path.First().Name);
```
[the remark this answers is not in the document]
TimothyMakkison: Not possible `path` is an `IEnumerable` type. It can't be `IReadOnlyList` because this prevents `Skip` from working. 
There are probably work arounds if you want me to explore them, although the original `MemberPath` code never used them

src/Riok.Mapperly/Descriptors/UnsafeAccessorContext.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    public UnsafeAccessorContext(UniqueNameBuilder nameBuilder, SymbolAccessor symbolAccessor)
+    {
+        _nameBuilder = nameBuilder.NewScope();
+        _symbolAccessor = symbolAccessor;
+    }
+
+    public IReadOnlyCollection<IUnsafeAccessor> UnsafeAccessors => _unsafeAccessors.Values;
+
+    public IUnsafeAccessor BuildGetProperty(ITypeSymbol source, IPropertySymbol symbol)
```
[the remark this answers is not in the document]
TimothyMakkison: WDYT about the following? I can cleanup the `Unsafe*****Accessor` constructors by using `symbol.ContainingType`, the function does feel a little strange due to the heavy usage of switch statemenets.

```C#
 public IUnsafeAccessor GetOrBuildAccessor(UnsafeAccessorType type, ISymbol symbol)
{
    var key = new UnsafeAccessorKey(symbol, type);
    if (_unsafeAccessors.TryGetValue(key, out var value))
    {
        return value;
    }

    var formatted = FormatAccessorName(symbol.Name);

    var defaultMethodName = type switch
    {
        UnsafeAccessorType.GetField => $"Get{formatted}",
        UnsafeAccessorType.GetProperty => $"Get{formatted}",
        UnsafeAccessorType.SetProperty => $"Set{formatted}",
        _ => throw new Exception($"Unknown {nameof(UnsafeAccessorType)}."),
    };
    var methodName = GetValidMethodName(symbol.ContainingType, defaultMethodName);

    IUnsafeAccessor accessor = type switch
    {
        UnsafeAccessorType.GetField => new UnsafeFieldAccessor(symbol.ContainingType, (IFieldSymbol)symbol, methodName),
        UnsafeAccessorType.GetProperty => new UnsafeGetPropertyAccessor(symbol.ContainingType, (IPropertySymbol)symbol, methodName),
        UnsafeAccessorType.SetProperty => new UnsafeSetPropertyAccessor(symbol.ContainingType, (IPropertySymbol)symbol, methodName),
        _ => throw new Exception($"Unknown {nameof(UnsafeAccessorType)}."),
    };

    _unsafeAccessors.Add(key, accessor);
    return accessor;
}
```

src/Riok.Mapperly/Descriptors/SimpleMappingBuilderContext.cs
```diff
@@ -83,5 +89,25 @@ public void ReportDiagnostic(DiagnosticDescriptor descriptor, ISymbol? location,
         _diagnostics.Add(Diagnostic.Create(descriptor, nodeLocation ?? _descriptor.Syntax.GetLocation(), messageArgs));
     }
 
+    public void BuildMemberVisibility()
+    {
+        var includedMembers = _configurationReader.Mapper.IncludedMembers;
```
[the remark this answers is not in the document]
TimothyMakkison: Should I add a new compile time constant for Roslyn 4.7? i'm not 100% sure but would `UnsafeAccessors` be `ROSLYN4_7_OR_GREATER`.

test/Riok.Mapperly.Tests/Mapping/UnsafeAccessorTest.cs
TimothyMakkison: I'm not sure what you mean here. Do you want me to add a test where the .NET version doesn't support `UnsafeAccessor`. I'm not sure how to do this. Should it be an integration test?

src/Riok.Mapperly/Symbols/SetterMemberPath.cs
```diff
@@ -0,0 +1,78 @@
...
+
+public class SetterMemberPath : MemberPath
+{
+    private SetterMemberPath(IReadOnlyList<IMappableMember> path, bool isMethod)
+        : base(path)
+    {
+        IsMethod = isMethod;
+    }
+
+    public bool IsMethod { get; }
```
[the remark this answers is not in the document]
TimothyMakkison: > I had a super hard time understanding what exactly this does. IMO this makes things really complicated. This seems to be related to setting fields via the ref return. This could get a lot simpler without this parameter, by setting fields via the accessor too.

`IsMethod` indicates that the setter must be invoked as it is an `UnsafeAccessor` for a property set. It has nothing to do with setting fields via `ref` return as the `UnsafeAccessor` for fields have the exact same usage and semantics as a normal property or field. `IsMethod` is only true for setters for `UnsafeAccessor`.

>  Is there a special reason for not doing this? (same for MethodMember).

At one point I did try a solution similar to what you suggested. Unfortunately it didn't work for `MemberNullAssignmentInitializerMapping` (`source.Value ??= new()`). The caller has to know if the `SetterMemberPath` is a method or not because you have to provide it with different sources. ie `new()` for a normal field, property or `UnsafeAccessor` field, whereas `source.GetValue() ?? new()` must be provided for an `UnsafeAccessor` for setters. See `MethodMemberNullAssignmentInitializerMapping`. I ended up with my current solution to handle these edge cases.

I'll try your suggestion with the `coalesceAssignment` parameter. Unfortunately I had to keep `IsMethod`, I might be missing something.

TimothyMakkison:
Thanks for the review, must have taken some time 😅
I've made the changes you've suggested, the only missing bits are `GetOrBuildAccessor` and the accessor test without visibility. Also fixed a pretty glaring issue with existing target types that my tests didn't catch 👍 
I still need to add a couple of xml comments and do the documentation.

Thanks again.

test/Riok.Mapperly.Tests/Mapping/UnsafeAccessorTest.cs
latonz: Currently only `MemberVisibility.All` and `MemberVisibility.AllAccessible` are tested, also add tests for at least another value (eg. `Public` and check whether only public members are mapped).

src/Riok.Mapperly/Descriptors/UnsafeAccessorContext.cs
```diff
@@ -0,0 +1,104 @@
...
+            _ => throw new Exception($"Unknown {nameof(UnsafeAccessorType)}."),
+        };
+        var methodName = GetValidMethodName(symbol.ContainingType, defaultMethodName);
+
+        IUnsafeAccessor accessor = type switch
+        {
+            UnsafeAccessorType.GetField => new UnsafeFieldAccessor((IFieldSymbol)symbol, methodName),
+            UnsafeAccessorType.GetProperty => new UnsafeGetPropertyAccessor((IPropertySymbol)symbol, methodName),
+            UnsafeAccessorType.SetProperty => new UnsafeSetPropertyAccessor((IPropertySymbol)symbol, methodName),
+            _ => throw new Exception($"Unknown {nameof(UnsafeAccessorType)}."),
```
latonz, line 44: Include the type, otherwise this might be quite hard to debug and probably use the more appropriate ArgumentOutOfRangeException here: throw new ArgumentOutOfRangeException(nameof(type), type, $"Unknown {nameof(UnsafeAccessorType)}.")

src/Riok.Mapperly/Symbols/SetterMemberPath.cs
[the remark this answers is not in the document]
latonz: Got it now 😊  Looks good to me.

src/Riok.Mapperly/Riok.Mapperly.Roslyn4.8.props
```diff
@@ -0,0 +1,12 @@
+<Project>
+
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);ROSLYN4_4_OR_GREATER</DefineConstants>
+    <DefineConstants>$(DefineConstants);ROSLYN4_8_OR_GREATER</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.8.0-2.final" PrivateAssets="all" />
```
latonz, line 9: If 4.8 is really needed can you also update the default version in `Riok.Mapperly.csproj > ROSLYN_VERSION`.
Can you explain to me why 4.8 is required instead of 4.7?

test/Riok.Mapperly.IntegrationTests/Mapper/TestMapper.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Riok.Mapperly.IntegrationTests.Mapper
 {
-    [Mapper]
+    [Mapper(IncludedMembers = MemberVisibility.All)]
```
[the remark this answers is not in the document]
latonz: Just saw that you are now guarding this with but wouldn't `NET8_0_OR_GREATER` be better here?

src/Riok.Mapperly/Symbols/GetterMemberPath.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+        else
+        {
+            var unsafeAccessor = ctx.UnsafeAccessorContext.GetOrBuildAccessor(
+                UnsafeAccessorContext.UnsafeAccessorType.GetProperty,
+                (IPropertySymbol)item.MemberSymbol
+            );
+
+            return new MethodAccessorMember(item, unsafeAccessor.MethodName);
+        }
```
latonz, line 50: Since L40 returns, the else block can be unwrapped.

src/Riok.Mapperly/Symbols/MethodAccessorMember.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Riok.Mapperly.Emit.Syntax.SyntaxFactoryHelper;
+
+namespace Riok.Mapperly.Symbols;
+
+public class MethodAccessorMember : IMappableMember
+{
+    private readonly IMappableMember _mappableMember;
+    private readonly string _methodName;
+    private readonly bool _isMethod;
```
latonz, line 11: I think the isMethod name here is still confusing as this is inside a `MethodAccessorMember`. Probably `targetValueIsMethodParameter` or similar works better here?

latonz (changes requested):
Thank you for the updates. Sorry for the slow reviews but this MR is huge 🙈  We should try to split future MR's into smaller ones to be better manageable.

src/Riok.Mapperly/Symbols/MethodAccessorMember.cs
```diff
@@ -0,0 +1,46 @@
...
+    public bool CanSet => _mappableMember.CanSet;
+    public bool CanSetDirectly => _mappableMember.CanSetDirectly;
+    public bool IsInitOnly => _mappableMember.IsInitOnly;
+    public bool IsRequired => _mappableMember.IsRequired;
+
+    public ExpressionSyntax BuildAccess(ExpressionSyntax source, bool nullConditional = false)
+    {
+        if (_isMethod)
+        {
+            return nullConditional ? ConditionalAccess(source, _methodName) : MemberAccess(source, _methodName);
```
latonz, line 35: Add a comment explaining that the caller will do the invocation.

test/Riok.Mapperly.IntegrationTests/Mapper/TestMapper.cs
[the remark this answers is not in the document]
TimothyMakkison: > Just saw that you are now guarding this with but wouldn't NET8_0_OR_GREATER be better here?

I copied the other pre processor directives, do you want me to update them as well?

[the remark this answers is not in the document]
latonz: Oh yep, that doesn't make sense, my bad. You can update them, thank you 😊

src/Riok.Mapperly/Riok.Mapperly.Roslyn4.8.props
TimothyMakkison: I was literally just about to ask for help with props/preprocessors, before you added your comments. I'm struggling
😆 

> If 4.8 is really needed 

Not 100% sure but here's my thinking: `DescriptorBuilder` should emit a diagnostic for versions of .NET lower than 8, to this I used `#if ROSLYN4_4_OR_GREATER`. This worked until I realised that the integration tests would fail because Roslyn 4.4 applied to older versions of .NET.

I tried to fix this by adding `ROSLYN4_8_OR_GREATER` which requires its corresponding props. As I understand it, Roslyn 4.8 corresponds to .NET 8. I could set the version for `Microsoft.CodeAnalysis.CSharp` to 4.7.0 but I thought it made more sense to use 4.8.0.

I'd have preferred not to make any of these changes, but I can't think of how else to add a new constant. I'd appreciate guidance here, I've never really understood custom constants and props.

src/Riok.Mapperly/Symbols/SetterMemberPath.cs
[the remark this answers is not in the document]
TimothyMakkison: The variable path is reassigned to path.Skip(1) if baseAccess is null. It has to be a local variable and must be IEnumerable<IMappableMember> for the skip to work.

I'm pretty sure I copied this code straight from the original memberPath

docs/docs/configuration/private-member-mapping.md
```diff
@@ -0,0 +1,62 @@
...
+description: Private member mapping
+---
+
+# Private member mapping
+
+As of .NET 8.0, Mapperly supports mapping members that are normally inaccessible like `private` or `protected` properties. This is made possible by using the [UnsafeAccessorAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute) which lets Mapperly access normally inaccessible members with zero overhead while being completely AOT safe.
+
+By default `IncludedMembers` is set to `MemberVisibility.AllAccessible` which will configure Mapperly to map members of all accessibility levels as long as they are ordinarily accessible. To enable unsafe accessor usage, set `IncludedMembers` to `MemberVisibility.All`. Mapperly will then try to map members of all accessibilities, including ones that are not usually visible to external types.
+
+```csharp
```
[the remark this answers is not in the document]
TimothyMakkison: Should this be a separate page or a sub paragraph?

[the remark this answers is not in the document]
latonz: I'd do it on the same page as a separate paragraph.

src/Riok.Mapperly/Riok.Mapperly.Roslyn4.8.props
latonz: Since this PR is already huge, just use `4.7` for the condition, this should work for now. We can then add `4.8` in another PR (the steps needed are more or less explained [here](https://mapperly.riok.app/docs/contributing/common-tasks/#add-support-for-a-new-roslyn-version)). Let me know if you need any help on how to do this.

docs/docs/configuration/private-member-mapping.md
```diff
@@ -0,0 +1,115 @@
...
+    [global::System.Runtime.CompilerServices.UnsafeAccessor(global::System.Runtime.CompilerServices.UnsafeAccessorKind.Method, Name = "set_Sweetness")]
+    public static extern void SetSweetness(this global::FruitDto target, int value);
+}
+```
+
+Here Mapperly generates a file scoped class containing extension method for each internal member for both the source and target. Mapperly then uses the extension methods to get and set the members. Note that this uses zero reflection and is as performant as using an ordinary property or field.
+
+## Controlling member accessibility
+
+In addition to mapping inaccessible members, `MemberVisbility` can be used to control which members are mapped, depending on their accessibility modifier. For instance `MemberVisibility.Private & MemberVisibility.Protected` will cause mapperly to only map private and protected members, generating an unsafe accessor if needed.
```
latonz, line 66: It should be `MemberVisibility.Private | MemberVisibility.Protected` to combine the flags...

src/Riok.Mapperly/Symbols/SetterMemberPath.cs
```diff
@@ -0,0 +1,92 @@
...
+        if (member.MemberSymbol.Kind == SymbolKind.Field)
+        {
+            var unsafeAccessor = ctx.UnsafeAccessorContext.GetOrBuildAccessor(
+                UnsafeAccessorContext.UnsafeAccessorType.GetField,
+                (IFieldSymbol)member.MemberSymbol
+            );
+
+            return (new MethodAccessorMember(member, unsafeAccessor.MethodName), false);
+        }
+        else
```
latonz, line 51: The else is not needed here since the if block returns already.

src/Riok.Mapperly/Riok.Mapperly.csproj
```diff
@@ -1,68 +1,68 @@
...
-  <ItemGroup>
-    <None Include="../../README.md" Pack="true" PackagePath="\" />
-    <None Include="../../docs/static/img/logo.png" Pack="true" PackagePath="\" />
-    <None Include="$(OutputPath)\$(AssemblyName).dll" Pack="true" PackagePath="analyzers/roslyn$(ROSLYN_VERSION)/dotnet/cs" Visible="false" />
-    <None Include="$(OutputPath)\$(AssemblyName).Abstractions.dll" Pack="true" PackagePath="analyzers/roslyn$(ROSLYN_VERSION)/dotnet/cs" Visible="false" />
-    <None Include="$(OutputPath)\$(AssemblyName).Abstractions.dll" Pack="true" PackagePath="lib\netstandard2.0" Visible="true" />
-    <None Include="$(OutputPath)\$(AssemblyName).Abstractions.xml" Pack="true" PackagePath="lib\netstandard2.0" Visible="true" />
-  </ItemGroup>
-</Project>
+<Project Sdk="Microsoft.NET.Sdk">
```
latonz, line 1: A lot of diffs here... Not sure what changed, but probably revert all except for the expected changes? Probably a change in the file encoding or similar 🤔

latonz (changes requested):

TimothyMakkison:
That should be the last of that changes 😅 no idea why `csproj` got into a funk.

Massive thanks @latonz, this PR is massive. I'll definitely have to break up #622 if I ever return to it.

latonz (approved):
Looks good to me now 🥳  Thank you!
Can you rebase onto main?
For future PR's we should definitely try to keep them smaller, it makes reviewing and merging easier for everyone involved.

PR #746, opened 2023-09-12
Fix/issue 744 small

### LINKED ISSUE #744: Nullable Enums not mapped correctly when passed to constructor
**Describe the bug**
When mapping nullable enums or struct values as a constructor parameter, the null values will be mapped to default instances of the enum or struct.

**To Reproduce**
Use this code snippit to gennerate a Mapper:
```csharp

enum SourceEnum { Value1, Value2 }
enum TargetEnum { Value1, Value2 }

struct SourceStruct { public int Value { get; set; } }
struct TargetStruct { public int Value { get; set; } }

record SourceRecord(SourceEnum? EnumValue, SourceStruct? StructValue);
record TargetRecord(TargetEnum? EnumValue, TargetStruct? StructValue);

static partial class Mapper
{
    public static partial TargetRecord Map(this SourceRecord record);
}
``` 

This will generate the following code for the Mapper:
```csharp
static partial class Mapper
{
    public static partial global::TargetRecord Map(this global::SourceRecord record)
    {
        var target = new global::TargetRecord(
            record.EnumValue != null ? MapToTargetEnum(record.EnumValue.Value) : default, 
            record.StructValue != null ? MapToTargetStruct(record.StructValue.Value) : default);
        return target;
    }

    //...
}
```

**Problem**
The issue is with the use of `default` which will infer the type from the `MapToTargetEnum()` and the `MapToTargetStruct()` functions, and create defaults of the non-nullable enum and structs.

**Expected behavior**
Use the type specific `default` to ensure nullability is preserved:
```csharp
static partial class Mapper
{
    public static partial global::TargetRecord Map(this global::SourceRecord record)
    {
        var target = new global::TargetRecord(
            record.EnumValue != null ? MapToTargetEnum(record.EnumValue.Value) : default(global::TargetEnum?),
            record.StructValue != null ? MapToTargetStruct(record.StructValue.Value) : default(global::TargetStruct?));
        return target;
    }

    //...
}
```

- Mapperly Version: 3.1.0
- .NET Version: 7.0.306
- Target Framework: net7.0
- IDE: VisualStudio Enterprise 2022 v17.6.5
- OS: Windows 11

rickykaare:
# Add correct type for `default` statements when using nullable value types

## Description
* Change `NullSubstitute` to use `DefaultExpression` instead of `DefaultLiteral` for nullable value types
* Change callers to pass correct type to `NullSubstitute` 
* Update snapshots and unit tests

Fixes #744

## Checklist


src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.Null.cs
```diff
@@ -33,9 +33,10 @@ public static ExpressionSyntax NullSubstitute(ITypeSymbol t, ExpressionSyntax ar
     {
         return nullFallbackValue switch
         {
+            NullFallbackValue.Default when t.IsNullableValueType() => DefaultExpression(FullyQualifiedIdentifier(t)),
             NullFallbackValue.Default => DefaultLiteral(),
             NullFallbackValue.EmptyString => StringLiteral(string.Empty),
-            NullFallbackValue.CreateInstance => CreateInstance(t),
+            NullFallbackValue.CreateInstance => CreateInstance(t.NonNullable()),
```
latonz, line 39: `CreateInstance` internally always uses a non-nullable identifier, so `NonNullable` is not needed here.

test/Riok.Mapperly.Tests/Mapping/EnumTest.cs
```diff
@@ -213,7 +213,10 @@ public void NullableEnumToOtherNullableEnumShouldCast()
     {
         var source = TestSourceBuilder.Mapping("E1?", "E2?", "enum E1 {A, B, C}", "enum E2 {A, B, C}");
 
-        TestHelper.GenerateMapper(source).Should().HaveSingleMethodBody("return source == null ? default : (global::E2)source.Value;");
+        TestHelper
+            .GenerateMapper(source)
+            .Should()
+            .HaveSingleMethodBody("return source == null ? default(global::E2? ) : (global::E2)source.Value;");
```
latonz, line 219: Do you have an idea why this whitespace in `default(global::E2? )` is added?

latonz (changes requested):
Thank you for this contribution 🚀 
Looks almost ready to merge 😊 I added one feedback and one question.
Could you also add a test for the case you described in #744 (probably in `ObjectPropertyConstructorResolverTest`)?

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.Null.cs
rickykaare: This fixed.

test/Riok.Mapperly.Tests/Mapping/EnumTest.cs
rickykaare: I was annoyed by this myself, but it turns out the `NormalizeWhitespace()` call in [`GeneratedMethod.ExtractBody()`](https://github.com/riok/mapperly/blob/e073f74390d6ad37b0f1ff8d3d00bf7bbf7f0c72/test/Riok.Mapperly.Tests/GeneratedMethod.cs#L32), is the one adding the extra space:

```csharp
    private static string ExtractBody(MethodDeclarationSyntax declarationSyntax)
    {
        var body =
            declarationSyntax.Body
                ?.NormalizeWhitespace()
                .ToFullString()
                .Trim(' ', '\r', '\n')
                .Trim('{', '}')
                .Trim(' ', '\r', '\n')
                .ReplaceLineEndings() ?? string.Empty;
        var lines = body.Split(Environment.NewLine);
        return lines.Length == 0
            ? string.Empty
            : string.Join(Environment.NewLine, lines.Select(l => l.StartsWith(MethodIndentation) ? l[MethodIndentation.Length..] : l));
    }
```

Not sure if there is a way to control the formatting, or if it matters - as it seems this only affects the tests.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyConstructorResolverTest.cs
```diff
@@ -483,4 +483,70 @@ public void CanResolveToClassConstructorWithMapPropertyAttribute()
...
+            {
+                public static partial TargetRecord Map(this SourceRecord record);
+            }
+            """
+        );
+
+        TestHelper
+            .GenerateMapper(source)
+            .Should()
+            .HaveMethodBody(
```
latonz, line 544: You should be able to use the simplified `HaveMapMethodBody` here.

rickykaare:
> Thank you for this contribution 🚀 Looks almost ready to merge 😊 I added one feedback and one question. Could you also add a test for the case you described in #744 (probably in `ObjectPropertyConstructorResolverTest`)?

I've added tests using the sample code in the issue #744

test/Riok.Mapperly.Tests/Mapping/EnumTest.cs
latonz: Thank you for investigating this. We should probably have a look (in another PR as it is unrelated) if we can remove this `NormalizeWhitespace()` here completely.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyConstructorResolverTest.cs
rickykaare: Fixed.

latonz:
@rickykaare thank you for your first contribution to Mapperly 😻

rickykaare:
@latonz you're welcome 😊 Thanks for all the help - this was a really nice experience.

And thank you for a great library! 🚀

PR #756, opened 2023-09-16
fix: check for arguments with restricted keywords

### LINKED ISSUE #402: `@` in variable names is ignored
**Describe the bug**
If a map is declared with a parameter name being a keyword escaped with `@` (e.g., `Dto Map(MyObject @object)`), the generator fails because `@` gets ignored.

**To Reproduce**
```C#
[Mapper]
public partial class Mapper
{
    public partial string DomainToDto(object @object);
}
```
Which gives this code-gen
```C#
#nullable enable
public partial class Mapper
{
    public partial string DomainToDto(object object) // Notice: this is invalid C# and causes the error
    {
        return object.ToString();
    }
}
```

**Expected behavior**
The generated code is syntactically sound.

- Mapperly Version: 2.8.0
- .NET Version:  .NET 6.0.408
- Target Framework: .net6.0
- IDE: Rider 2022.3.3
- OS: Windows 11

TimothyMakkison:
## Description
- Added `RestrictedNameHelper` which is called by the `MethodParameter` constructor
   - `ToSafeName` checks the input for every restricted keyword. This should be really fast in .Net 8 but I doubt it will make any impact in prior versions
- Added tests

Fixes #402

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Symbols/MethodParameter.cs
+++ b/src/Riok.Mapperly/Symbols/MethodParameter.cs
@@ -6,8 +6,14 @@ namespace Riok.Mapperly.Symbols;
 
 public readonly record struct MethodParameter(int Ordinal, string Name, ITypeSymbol Type)
 {
+    private static readonly SymbolDisplayFormat _parameterNameFormat =
+        new(
+            parameterOptions: SymbolDisplayParameterOptions.IncludeName,
+            miscellaneousOptions: SymbolDisplayMiscellaneousOptions.EscapeKeywordIdentifiers
+        );
+
     public MethodParameter(IParameterSymbol symbol)
-        : this(symbol.Ordinal, symbol.Name, symbol.Type.UpgradeNullable()) { }
+        : this(symbol.Ordinal, symbol.ToDisplayString(_parameterNameFormat), symbol.Type.UpgradeNullable()) { }
 
     public MethodArgument WithArgument(ExpressionSyntax? argument) =>
         new(this, argument ?? throw new ArgumentNullException(nameof(argument)));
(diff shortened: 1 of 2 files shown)


src/Riok.Mapperly/Symbols/MethodParameter.cs
```diff
@@ -4,11 +4,24 @@
 
 namespace Riok.Mapperly.Symbols;
 
-public readonly record struct MethodParameter(int Ordinal, string Name, ITypeSymbol Type)
+public readonly record struct MethodParameter
 {
     public MethodParameter(IParameterSymbol symbol)
         : this(symbol.Ordinal, symbol.Name, symbol.Type.UpgradeNullable()) { }
```
latonz, line 10: Couldn't this fix be implement as simple as instead of using `symbol.Name` here using `symbol.ToDisplayString(_parameterNameFormat)` with 
```c#
private static readonly SymbolDisplayFormat _parameterNameFormat = new SymbolDisplayFormat(
        parameterOptions: SymbolDisplayParameterOptions.IncludeName,
        miscellaneousOptions: SymbolDisplayMiscellaneousOptions.EscapeKeywordIdentifiers);
```

test/Riok.Mapperly.Tests/Mapping/RestrictedKeywordTest.cs
```diff
@@ -0,0 +1,63 @@
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class RestrictedKeywordTest
+{
+    [Fact]
+    public void NewInstance()
```
latonz, line 6: If the proposed approach works, I think a simple test in `Riok.Mapperly.Tests.Mapping.MapperTest` should be sufficient.

latonz (changes requested):

latonz (approved):

latonz (approved):

PR #864, opened 2023-11-03
Use pattern matching for null guards

### LINKED ISSUE #843: Use of match expression in null guards
The nullable check guard can be seen i.e. [here](https://github.com/riok/mapperly/blob/main/test/Riok.Mapperly.Tests/_snapshots/UnsafeAccessorTest.PrivateNestedNullableProperty%23Mapper.g.verified.cs) causes duplicated call for source value.
```cs
if (source.GetNested()?.GetValue() != null)
{
    target.SetValue(source.GetNested().GetValue().Value);
}
```

I think use of match expression is better:
```cs
if (source.GetNested()?.GetValue() is {} sourceGetNestedGetValue)
{
    target.SetValue(sourceGetNestedGetValue);
}
```

trejjam:
# Use pattern matching for null guards

## Description

Generate null guards using pattern matching:
```cs
if (source.MayBeNull is {} sourceMayBeNull)
{
    target.NonNullableValue = sourceMayBeNull;
}
```

Fixes #843

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberNullDelegateAssignmentMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberNullDelegateAssignmentMapping.cs
@@ -34,11 +34,16 @@ public class MemberNullDelegateAssignmentMapping : MemberAssignmentMappingContai
         //   target.Value = Map(sourceValue);
         // else
         //   throw ...
-        var sourceNullConditionalAccess = _nullConditionalSourcePath.BuildAccess(ctx.Source, false, _needsNullSafeAccess, true);
-        var nameofSourceAccess = _nullConditionalSourcePath.BuildAccess(ctx.Source, false, false, true);
+        if (!GetterMemberPath.TryBuild(ctx, _nullConditionalSourcePath, out var nullConditionalSourcePath))
+        {
+            nullConditionalSourcePath = _nullConditionalSourcePath;
+        }
+
+        var sourceNullConditionalAccess = nullConditionalSourcePath.BuildAccess(ctx.Source, false, _needsNullSafeAccess, true);
+        var nameofSourceAccess = nullConditionalSourcePath.BuildAccess(ctx.Source, false, false, true);
 
         var (conditionCtx, nullGuardedValue) = ctx.AddIndentation()
-            .WithTrimSourcePath(_nullConditionalSourcePath.Path)
+            .WithTrimSourcePath(nullConditionalSourcePath.Path)
             .WithNewSource(sourceNullConditionalAccess.ToFullString().Replace(".", ""));
 
         var condition = IsNotNullWithPattern(sourceNullConditionalAccess, nullGuardedValue);
--- a/src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberNullDelegateAssignmentMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberNullDelegateAssignmentMapping.cs
@@ -30,14 +30,19 @@ public class MemberNullDelegateAssignmentMapping : MemberAssignmentMappingContai
 
     public override IEnumerable<StatementSyntax> Build(TypeMappingBuildContext ctx, ExpressionSyntax targetAccess)
     {
-        // if (source.Value != null)
-        //   target.Value = Map(Source.Name);
+        // if (source.Value is {} sourceValue)
+        //   target.Value = Map(sourceValue);
         // else
         //   throw ...
         var sourceNullConditionalAccess = _nullConditionalSourcePath.BuildAccess(ctx.Source, false, _needsNullSafeAccess, true);
         var nameofSourceAccess = _nullConditionalSourcePath.BuildAccess(ctx.Source, false, false, true);
-        var condition = IsNotNull(sourceNullConditionalAccess);
-        var conditionCtx = ctx.AddIndentation();
+
+        var (conditionCtx, nullGuardedValue) = ctx.AddIndentation()
+            .WithTrimSourcePath(_nullConditionalSourcePath.Path)
+            .WithNewSource(sourceNullConditionalAccess.ToFullString().Replace(".", ""));
+
+        var condition = IsNotNullWithPattern(sourceNullConditionalAccess, nullGuardedValue);
+
         var trueClause = base.Build(conditionCtx, targetAccess);
         var elseClause = _throwInsteadOfConditionalNullMapping
             ? new[] { conditionCtx.SyntaxFactory.ExpressionStatement(ThrowArgumentNullException(nameofSourceAccess)) }
--- a/src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberNullDelegateAssignmentMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberNullDelegateAssignmentMapping.cs
@@ -44,7 +44,14 @@ public class MemberNullDelegateAssignmentMapping : MemberAssignmentMappingContai
 
         var (conditionCtx, nullGuardedValue) = ctx.AddIndentation()
             .WithTrimSourcePath(nullConditionalSourcePath.Path)
-            .WithNewSource(sourceNullConditionalAccess.ToFullString().Replace(".", ""));
+            .WithNewSource(
+                sourceNullConditionalAccess
+                    .ToFullString()
+                    .Replace(".", string.Empty, StringComparison.InvariantCultureIgnoreCase)
+                    .Replace("?", string.Empty, StringComparison.InvariantCultureIgnoreCase)
+                    .Replace("(", string.Empty, StringComparison.InvariantCultureIgnoreCase)
+                    .Replace(")", string.Empty, StringComparison.InvariantCultureIgnoreCase)
+            );
 
         var condition = IsNotNullWithPattern(sourceNullConditionalAccess, nullGuardedValue);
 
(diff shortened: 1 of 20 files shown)


trejjam:
I will squash it to the required commit message after the review

benchmarks/Riok.Mapperly.Benchmarks/MappingBenchmarks.cs
```diff
@@ -1,4 +1,4 @@
-﻿using BenchmarkDotNet.Attributes;
```
TimothyMakkison, line 1: Why is the first import for each file modified? 😁

TimothyMakkison (changes requested):

benchmarks/Riok.Mapperly.Benchmarks/MappingBenchmarks.cs
trejjam: I applied `dotnet format` as requested, and these files were UTF-8-BOM. `dotnet format` changed them to UTF-8. Since BOM is several bytes at the beginning of each file it seems like a line change.

latonz: Could you undo these changes (adding the UTF-8-BOM)? They make it super hard to review, if we want to add it, we should do such mass changes in a separate PR. Also if I'm running `dotnet format` on the current main branch I don't get any changes 🤔  Which version of `dotnet format` are you using (`dotnet format --version`)?

latonz: However, the current main definitely has mixed UTF-8 (with BOM and without BOM). We should unify this, would be happy to accept a separate PR for this. Probably we should also invest why the editorconfig / linting didn't work properly here.

trejjam: Hi,
my dotnet format:
```sh
> dotnet format --version
8.0.447701+0cf0f51b02d759e89f2acc09663493ab422548b0
```
I dropped the code style commit, I will create a separate PR.

trejjam: My git is configured to `autocrlf = input`, if configured differently it may hide issues with platform line endings/file encoding

trejjam: PR: #869

latonz (commented):
Thank you for your efforts and this PR. I first looked at it some time ago but wasn't really happy with the design of the solution. In the meantime I've tried to come up with alternative designs, but haven't come up with a better solution. Tracking the member prefix in the context looks like a "hack" to me and may cause problems later on. I'll keep this PR open for now as I continue to think about alternative designs. Please feel free to comment any ideas you might have.

src/Riok.Mapperly/Descriptors/Mappings/TypeMappingBuildContext.cs
```diff
@@ -30,15 +32,17 @@ SyntaxFactoryHelper syntaxFactory
...
     }
 
-    public UniqueNameBuilder NameBuilder { get; }
+    public UniqueNameBuilder NameBuilder { get; private init; }
 
-    public ExpressionSyntax Source { get; }
+    public ExpressionSyntax Source { get; private init; }
 
-    public ExpressionSyntax? ReferenceHandler { get; }
+    public IReadOnlyList<IMappableMember> TrimSourcePath { get; private init; } = ImmutableList<IMappableMember>.Empty;
```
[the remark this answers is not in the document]
trejjam: I can work a bit on the documentation side.

This property should contain `A, B` when a full path is `A.B?.C` so it can replace `A.B` with the newly introduced guard variable.

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberNullDelegateAssignmentMapping.cs
```diff
@@ -30,14 +30,31 @@ bool needsNullSafeAccess
...
+
+        var (conditionCtx, nullGuardedValue) = ctx.AddIndentation()
+            .WithTrimSourcePath(nullConditionalSourcePath.Path)
+            .WithNewSource(
+                sourceNullConditionalAccess
+                    .ToFullString()
+                    .Replace(".", string.Empty, StringComparison.InvariantCultureIgnoreCase)
+                    .Replace("?", string.Empty, StringComparison.InvariantCultureIgnoreCase)
+                    .Replace("(", string.Empty, StringComparison.InvariantCultureIgnoreCase)
+                    .Replace(")", string.Empty, StringComparison.InvariantCultureIgnoreCase)
```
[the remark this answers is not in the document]
trejjam: I can simplify the name, my goal was to prevent name duplicities. But they are handled in the UniqueNameBuilder

trejjam:
No worries, I understand it as a one-at-the-time. I was ok to wait until the static PR made it :)

TimothyMakkison:
This is a little late but are we sure this is faster? I wouldn't be shocked if the jit compiler does this.

trejjam:
Hello @TimothyMakkison 

I run a small benchmark:
```cs
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Jobs;

namespace Sandbox;

public record A(int Value);

public record D(int Value);

public record struct As(int Value);

public record struct Ds(int Value);

public record B(A? Value);

public record E(D? Value);

public record struct Bs(As? Value);

public record struct Es(Ds? Value);

[SimpleJob(RuntimeMoniker.Net70)]
[SimpleJob(RuntimeMoniker.Net80)]
[MemoryDiagnoser]
public class Bench
{
    private readonly B[] _source =
    {
        new(new A(1)),
        new(null),
    };

    private readonly Bs[] _sourceStruct =
    {
        new(new As(1)),
        new(null),
    };

    [Benchmark]
    public List<E> Map()
    {
        var result = new List<E>(_source.Length);
        foreach (var x in _source)
        {
            if (x.Value != null)
            {
                result.Add(new E(new D(x.Value.Value)));
            }
            else
            {
                result.Add(new E(null));
            }
        }

        return result;
    }

    [Benchmark]
    public List<E> Map_Guard()
    {
        var result = new List<E>(_source.Length);
        foreach (var x in _source)
        {
            if (x.Value is { } val)
            {
                result.Add(new E(new D(val.Value)));
            }
            else
            {
                result.Add(new E(null));
            }
        }

        return result;
    }

    [Benchmark]
    public List<Es> Map_Struct()
    {
        var result = new List<Es>(_sourceStruct.Length);
        foreach (var x in _sourceStruct)
        {
            if (x.Value != null)
            {
                result.Add(new Es(new Ds(x.Value.Value.Value)));
            }
            else
            {
                result.Add(new Es(null));
            }
        }

        return result;
    }

    [Benchmark]
    public List<Es> Map_Struct_Guard()
    {
        var result = new List<Es>(_sourceStruct.Length);
        foreach (var x in _sourceStruct)
        {
            if (x.Value is { } val)
            {
                result.Add(new Es(new Ds(val.Value)));
            }
            else
            {
                result.Add(new Es(null));
            }
        }

        return result;
    }
}
```

with the following result:
```md
BenchmarkDotNet v0.13.10, Windows 11 (10.0.22621.2715/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1260P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.100
  [Host]   : .NET 8.0.0 (8.0.23.53103), X64 RyuJIT AVX2
  .NET 7.0 : .NET 7.0.14 (7.0.1423.51910), X64 RyuJIT AVX2
  .NET 8.0 : .NET 8.0.0 (8.0.23.53103), X64 RyuJIT AVX2


| Method           | Job      | Runtime  | Mean     | Error    | StdDev   | Median   | Gen0   | Allocated |
|----------------- |--------- |--------- |---------:|---------:|---------:|---------:|-------:|----------:|
| Map              | .NET 7.0 | .NET 7.0 | 26.95 ns | 0.593 ns | 0.990 ns | 26.75 ns | 0.0153 |     144 B |
| Map_Guard        | .NET 7.0 | .NET 7.0 | 26.59 ns | 0.764 ns | 2.078 ns | 26.35 ns | 0.0153 |     144 B |
| Map_Struct       | .NET 7.0 | .NET 7.0 | 19.52 ns | 1.598 ns | 4.660 ns | 17.45 ns | 0.0076 |      72 B |
| Map_Struct_Guard | .NET 7.0 | .NET 7.0 | 19.32 ns | 0.549 ns | 1.603 ns | 19.01 ns | 0.0076 |      72 B |
| Map              | .NET 8.0 | .NET 8.0 | 21.74 ns | 0.739 ns | 2.048 ns | 20.79 ns | 0.0153 |     144 B |
| Map_Guard        | .NET 8.0 | .NET 8.0 | 29.29 ns | 1.817 ns | 5.212 ns | 27.74 ns | 0.0153 |     144 B |
| Map_Struct       | .NET 8.0 | .NET 8.0 | 11.12 ns | 0.427 ns | 1.218 ns | 10.74 ns | 0.0076 |      72 B |
| Map_Struct_Guard | .NET 8.0 | .NET 8.0 | 13.05 ns | 0.805 ns | 2.310 ns | 12.51 ns | 0.0076 |      72 B |
```

It seems that on .NET7 the guard variable version was a little bit faster. NET8 is faster in general and the naive implementation is also faster. So the PR may be not useful...

TimothyMakkison:
Hey @trejjam, thanks for writing the benchmarks! For what it's worth the number vary a lot when performing microbenchmarks, how many times have you run this? The change between runs can be quit significant and hard to track.
The drop in time for `Map` & `Map_Struct` seems suspiciously large. I wonder if the relevant change is mentioned in Stpehen Toub's [.NET 8 optimization blog](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/).

I'll try running your benchmarks later, I'll try looking at IL code but I can't promise anything 😄

TimothyMakkison:
> I have run this benchmark 3times (first without struct and memory info) and the numbers were always in this way.

Thanks, my numbers were all over the place 😅

>  I accept it as a usual performance improvement :D 6->7->8 I will let you know if I find something in Stephen's blog post

Suffering from success 😁. I only mentioned his blog because that's where I usually read about improvement. Please don't read the whole thing on behalf of me.

PR #878, opened 2023-11-09
Feat: add recursive depth support to projection mappings

### LINKED ISSUE #785: parent child IQueryable mapping: recursion depth support
**Describe the bug**
when i want map parent class with prop of list of child and child class have prop of parent class (simple parent child relationship) and use IQueryable mapping mappings not generated and  "Stack overflow" error occurre on build

**To Reproduce**
```csharp
public class CarDto
{
    public string Name { get; set; }
    public List<SeatDto> Seats { get; set; }
}
public class SeatDto { 
    public int Position { get; set; }
    public CarDto Car { get; set; }
}

public class Car
{
    public string Name { get; set; }
    public List<Seat> Seats { get; set; }
}
public class Seat
{
    public int Position { get; set; }
    public Car Car { get; set; }
}

```

```csharp
[Mapper]
{
    public static partial IQueryable<CarDto> ProjectToDto(this IQueryable<Car> q);
}
```

**Expected behavior**
i thing its stuck in loop of mapping but i think would be great if ignoring `Car `property on `Seat` class mapping or atlest it could be ignord by MapperIgnoreSourceAttribute (`MapperIgnoreSourceAttribute` & `MapperIgnoreTargetAttribute` not working on IQueryable mappings )

- Mapperly Version: 3.2.0
- .NET Version: .NET 6.0.102
- Target Framework: .net6.0
- IDE: Rider 2023
- OS: windows 10
e following information):**
- Mapperly Version: 3.2.0
- .NET Version: .NET 6.0.102
- Target Framework: .net6.0
- IDE: Rider 2023
- OS: windows 10

**Additional context**
 `MapperIgnoreSourceAttribute` & `MapperIgnoreTargetAttribute` not working on IQueryable mappings

TimothyMakkison:
# Added max recursive depth support for projection mappings.

## Description
Added attributes and properties for `MaxDepth`. `InlineExpressionMappingBuilderContext` uses an `ImmutableDictionary<(ISymbol, ISymbol), int>` to track types mapped by its parents, returning default when the limit is reached. The loop is prevented from occuring by disabling `FindMapping` for previously seen types.

Undecided/ not done
- What should the names be.
- How should negative/0 numbers be handled ie errors or revert back to current behaviour.
- What should the default depth be set to.
- This could be used in `MappingBuilderContext` for improved loop detection, perhaps this could be explored in a separate PR.
- Need to add tests and documentation 😞 

WIP
Fixes #785, #820, #99

## Checklist


latonz:
Does this really resolve #849?

TimothyMakkison:
Mb 😅, knew this had a load of issues and saw "circular reference" in the title. I'll fix it

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
```diff
@@ -154,12 +168,18 @@ public override bool IsConversionEnabled(MappingConversionType conversionType)
         ITypeSymbol sourceType,
         ITypeSymbol targetType,
         MappingBuildingOptions options
-    ) =>
-        new InlineExpressionMappingBuilderContext(
+    )
+    {
+        var newParentTypes = _parentTypes.TryGetValue((sourceType, targetType), out var count)
+            ? _parentTypes.SetItem((sourceType, targetType), count + 1)
+            : _parentTypes.SetItem((sourceType, targetType), 1);
```
latonz, line 175: Couldn't this be simplified to:
```c#
var mappingRecursionCount = _parentTypes.GetValueOrDefault((sourceType, targetType));
var newParentTypes = _parentTypes.SetItem((sourceType, targetType), mappingRecursionCount + 1);        
```

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
```diff
@@ -15,6 +16,7 @@ public class InlineExpressionMappingBuilderContext : MappingBuilderContext
 {
     private readonly MappingCollection _inlineExpressionMappings;
     private readonly MappingBuilderContext _parentContext;
+    private readonly ImmutableDictionary<(ISymbol, ISymbol), int> _parentTypes;
```
latonz, line 19: Wouldn't this be much better suited inside the `MappingCollection`? Probably named as `mappingRecursionCount`. The mapping collection could then expose a method to get the recursion count of a mapping by its source and target type.

latonz (changes requested):
Thank you for this PR, nice work!
Is there sth. which hinders us to implement this for all kind of mappings instead of just IQueryable projections? I think it should almost never apply to non-IQueryable mappings anyway (as the mappings are re-used). WDYT?

TimothyMakkison:
> Is there sth. which hinders us to implement this for all kind of mappings instead of just IQueryable projections?

Don't think so, just didn't see a massive need to.

> I think it should almost never apply to non-IQueryable mappings anyway (as the mappings are re-used). WDYT?

It could be used to enhance loop detection. The current implementation relies on the parent contexts type bing the same as the current mapped type, this fails to detect mapping like: `A->B->A->B`. Perhaps there is a use case for a nested mapping that eventually terminates without using `IReferenceHandler` :shrug:.

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
TimothyMakkison: IMO this makes more sense to be in `InlineExpressionMappingBuilderContext`.

`mappingRecursionCount` will either be unused by `MappingBuilderContext` or a new `MappingCollection` would have to be created for every context created to avoid passing state to other contexts.

latonz:
@TimothyMakkison I think I was confused, `IReferenceHandler` does only work at runtime, but what we are discussing here is the compile time 🙈  That all sounds good 😊 Since nothing is hindering us from implementing this for all mappings, I'd do so to improve the loop detection handling and eliminate possible source generator crashes due to stack overflow.

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
latonz: That makes sense, thank you for explaining.

test/Riok.Mapperly.Tests/Mapping/QueryableProjectionTest.cs
```diff
@@ -18,6 +18,21 @@ public Task ClassToClass()
         return TestHelper.VerifyGenerator(source);
     }
 
+    [Fact]
+    public Task ClassToClass2()
```
latonz, line 22: Use a more meaningful name, eg. `ClassToClassNested`

latonz (changes requested):
Thank you for the updates.

I think extracting `ImmutableDictionary<TypeMappingKey, int>` and the calls to it into a new struct/class (eg. `MappingRecoursingDepthTracker`) with a simple public API would probably make sense to simplify the logic inside the `InlineExpressionMappingBuilderContext`. WDYT?

Is `MaxRecursiveDepth` the correct term? Wouldn't a noun fit better eg. `MaxRecursionDepth`?

TimothyMakkison:
- Check for OBOE
- Logic is faulty

TODO:
- Use logic for normal mappings, not sure where
- Look at changing `ReferenceLoopInInitOnlyMapping` maybe
- Update integration tests
- What do `MaxRecursionDepth` less than zero, could diagnostic, not sure where

latonz:
@TimothyMakkison what is the status on this one? Should I review again?

TimothyMakkison:
> 
@TimothyMakkison what is the status on this one? Should I review again?

Super sorry about this. Kind of abadoned it.

Iirc this PR was nearish to completion

- Haven't added documentation
- Doesn't handle negative recursive values.
- Didn't add improved loop detection as I thought it would be better of as a different PR.

I'll try to finish it but I'm happy for someone else to complete it.

qutayba:
Hi @TimothyMakkison , is there any update on this feature?
Thank you very much for your PR. It would be awesome if we could get it finished.

TimothyMakkison:
Hey @qutayba IIRC this PR was mostly finished but wasn't merged due to changes to `MappingBuilderContext`. I don't think there is much more work to do, aside from resolving the merge conflcits with the tests etc. You are welcome to take this branch and make your own pull request 👍

PR #923, opened 2023-11-20
fix: extension method parameters for existing target mappings

trejjam:
# Extension methods may generate invalid code

## Description

Source
```cs
[Mapper]
class Mapper
{
  static partial void MapToB(this A value, B value);
}
class A {
}
class B {
}
```

Generates:
```cs
public partial class Mapper
{
    static partial void MapToB(this global::A source, this global::B target)
    {
        target.Value = source.Value;
    }
}
```

Notice `this` before the second parameter

## Checklist


test/Riok.Mapperly.Tests/Mapping/ExtensionMethodTest.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Riok.Mapperly.Tests.Mapping;
+
+[UsesVerify]
+public class ExtensionMethodTest
+{
+    [Fact]
+    public Task UpdateMethod()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """static partial void MapToB(this A source, B target);""",
```
TimothyMakkison, line 10: Don't think raw strings are needed here. I'd change this to `"partial void MapToB(this A source, B target);"` and `"class A { public int Value { get; set; } }"`, etc to match the other tests.

TimothyMakkison (changes requested):

test/Riok.Mapperly.Tests/Mapping/ExtensionMethodTest.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Riok.Mapperly.Tests.Mapping;
+
+[UsesVerify]
+public class ExtensionMethodTest
+{
+    [Fact]
+    public Task UpdateMethod()
```
latonz, line 7: `UpdateMethod` is not really a descriptive name for a test...
Probably `ExtensionMethodShouldWork` or similar should fit better.

latonz (changes requested):
Thank you for this PR!

latonz (approved):

PR #938, opened 2023-11-22
fix: when resolving constructors allow a non-nullable value type to be assigned to a nullable value type

### LINKED ISSUE #805: Nullable constructor parameters do not accept non nullable arguments.
**Describe the bug**
Mapperly does will not use a constructor if the source is a non nullable value type `T` and the constructor parameter is `Nullable<T>`.

Note that C# supports using `T` as arguments for `Nullable<T>`
```C#
var a = 10;
Doer(a);
// valid C# code
void Doer(int? v) { }
```

**To Reproduce**
```C#
public class A { public A(int? v) } }

public static partial A Map(int src);

// generates
public partial global::A Map(int src)
{
    // Could not generate mapping
    throw new System.NotImplementedException();
}
```

**Expected behavior**
Mapperly should use the nullable parameters constructor with a non nullable value.

latonz:
Fixes #805

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilders/CtorMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilders/CtorMappingBuilder.cs
@@ -22,7 +22,7 @@ public static class CtorMappingBuilder
             .FirstOrDefault(
                 m =>
                     m.Parameters.Length == 1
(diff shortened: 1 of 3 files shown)


TimothyMakkison:
Nice! I'll see how you did this later. I remember struggling with #809. I'd fix one issue and something else would break 😅

PR #988, opened 2023-12-08
Add parameter to annotation to allow mapping only primitive types

ziongh:
# Add parameter to annotation to allow mapping only primitive types

## Description

Sometimes there is the need to map only primitive types of a class. For instance if we want to clone an object to insert into the database, etc. So I've added a new property into the anotation configuration to allow that.

## Checklist


latonz:
Thank you for this contribution, this seems to be a very specific use case. Can you elaborate on why you need this?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
```diff
@@ -201,3 +216,44 @@ private void AddMappedAndIgnoredSourceMembersDiagnostics()
...
+                            && namedType.AllInterfaces.Any(s => string.Equals(s.Name, "IEnumerable", StringComparison.OrdinalIgnoreCase))
+                            && namedType.TypeArguments.Count() > 0
+                            && namedType.TypeArguments[0].IsComplexType(knowTypes)
+                        )
+                        || !(
+                            type is INamedTypeSymbol namedType2
+                            && namedType2.AllInterfaces.Any(s => string.Equals(s.Name, "IEnumerable", StringComparison.OrdinalIgnoreCase))
+                        )
+                    ),
+            TypeKind.Struct => false,
```
TimothyMakkison, line 241: It looks like this would accept any struct type, not just primitives. Is this intentional?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
```diff
@@ -201,3 +216,44 @@ private void AddMappedAndIgnoredSourceMembersDiagnostics()
         }
     }
 }
+
+public static class TypeExtensions
+{
+    public static bool IsComplexType(this ITypeSymbol type, WellKnownTypes knowTypes)
```
TimothyMakkison, line 222: Consider moving this method to `SymbolExtensions`

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
```diff
@@ -201,3 +216,44 @@ private void AddMappedAndIgnoredSourceMembersDiagnostics()
...
+            TypeKind.Delegate => true,
+            TypeKind.Interface => true,
+            // check if it's an array of primitives
+            TypeKind.Array
+                => type is IArrayTypeSymbol arrayType && arrayType.ElementType.IsComplexType(knowTypes)
+                    || type is not IArrayTypeSymbol arrayType2,
+            TypeKind.Dynamic => true,
+            TypeKind.Error => true,
+            TypeKind.Pointer => true,
+            TypeKind.Submission => false,
```
TimothyMakkison, line 252: Redundant case arm?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
```diff
@@ -201,3 +216,44 @@ private void AddMappedAndIgnoredSourceMembersDiagnostics()
...
+        return type.TypeKind switch
+        {
+            // check if it's an IEnumerable of primitives
+            TypeKind.Class
+                => type.Name != "string"
+                    && (
+                        (
+                            type is INamedTypeSymbol namedType
+                            && namedType.AllInterfaces.Any(s => string.Equals(s.Name, "IEnumerable", StringComparison.OrdinalIgnoreCase))
+                            && namedType.TypeArguments.Count() > 0
```
TimothyMakkison, line 233: Use `namedType.TypeArguments.Length`

ziongh:
@latonz, sorry, indeed I haven't provided enough background about this feature.

The main ideia is that normally the mapper will map every single property, complex type, and all single property of the related complex type unless we specify the ```MapperIgnoreSource``` or the ```MapperIgnoreTarget```.

However, when the object is a large object with many relations and the relations are also large object with many sub relations, and we want to map only the depth 0 (only the main object and not all the relations), currently, we have to ignore manually every single relation.

Another possible implementation would be to create a **MaxDepth** Parameter to allow the user to specify the maximum Depth to map. However, this option to map only primitive is also interesting, even in combination with the depth one.

@TimothyMakkison thank you for the review, indeed the code needed some refactoring, I've rewriten to be clearer. I've added one cenario sample in ```Riok.Mapperly.Sample``` project

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingBuilderContext.cs
ziongh: I actually did this, because in C# some system types are structs (decimal, DateOnly, etc.). However, I've rewritten this logic to be clearer.

latonz:
Thank you for the explanation.
After some consideration, we came to the conclusion that this feature isn't a good fit for Mapperly. It would generate many edge cases, for example required properties. The amount of possible configurations in Mapperly is already quite high, making it difficult to grasp for newer users. We want to keep the API surface as small as possible.
In the future, please open an issue for features and non-trivial work to discuss it before implementing it. This prevents situations such as this, where you spent a considerable amount of time on a feature that we sadly do not want to merge into Mapperly.

PR #1000, opened 2023-12-12
feat: generate generic mapping only with LSP compatible methods

trejjam:
# Generate generic mapping only with LSP compatible methods

## Description

Generic mapping is selecting target methods only by type argument constraint. This PR filters them also by LSP type compatibility.

Right now, the mapping does not produce valid code when e.g. source is a specific abstract type (not the `object`). And your mapper has e.g. enum mapping. The compiler then complains: `Error CS8121 : An expression of type 'TheClass' cannot be handled by a pattern of type 'TheEnum'.` on generated code that looks like this:
```cs
partial T Map<T>(TheClass source)
{
    return source switch
            {
                global::TheEnum x when typeof(T).IsAssignableFrom(typeof(global::TheEnum)) => (T)(object)MapToTheEnum(x),
            };
}
```

## Checklist

### DIFF
--- a/docs/docs/configuration/generic-mapping.md
+++ b/docs/docs/configuration/generic-mapping.md
@@ -17,10 +17,6 @@ using all mappings the user defined in the mapper that:
 public static partial class ModelMapper
 {
     // highlight-start
-    public static partial TTarget Map<TTarget>(object source);
-    // highlight-end
-
-    // highlight-start
     public static partial TTarget MapFruit<TTarget>(Fruit source);
     // highlight-end
 
--- a/docs/docs/configuration/generic-mapping.md
+++ b/docs/docs/configuration/generic-mapping.md
@@ -7,10 +7,7 @@ description: Create a generic mapping method
 
 Mapperly supports generic user defined mapping methods.
 Mapperly implements this mapping method
-using all mappings the user defined in the mapper that:
-
-- satisfy generic constraints,
-- and can be substituted instead of generic mapper
+using all mappings the user defined in the mapper that satisfy the source and target type constraints.
 
 ```csharp
 [Mapper]
--- a/docs/docs/configuration/generic-mapping.md
+++ b/docs/docs/configuration/generic-mapping.md
@@ -7,7 +7,10 @@ description: Create a generic mapping method
 
 Mapperly supports generic user defined mapping methods.
 Mapperly implements this mapping method
-using all mappings the user defined in the mapper.
+using all mappings the user defined in the mapper that:
+
+- satisfy generic constraints,
+- and can be substituted instead of generic mapper
 
 ```csharp
 [Mapper]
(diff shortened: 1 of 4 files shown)


docs/docs/configuration/generic-mapping.md
```diff
@@ -17,12 +20,17 @@ public static partial class ModelMapper
     public static partial TTarget Map<TTarget>(object source);
     // highlight-end
 
+    // highlight-start
+    public static partial TTarget MapFruit<TTarget>(Fruit source);
+    // highlight-end
+
```
latonz, line 26: I don't think this documentation changes add any additional value. I'd either remove the old sample and keep the new one or only keep the old one.

docs/docs/configuration/generic-mapping.md
```diff
@@ -7,7 +7,10 @@ description: Create a generic mapping method
 
 Mapperly supports generic user defined mapping methods.
 Mapperly implements this mapping method
-using all mappings the user defined in the mapper.
+using all mappings the user defined in the mapper that:
+
+- satisfy generic constraints,
+- and can be substituted instead of generic mapper
```
latonz, line 13: I don't really understand the second part `and can be substituted instead of generic mapper` (probably due to my bad english 🤔 🙈 )... Couldn't this be generalized by writing `... using all mappings the user defined in the mapper that satisfy the source and target type constraints.`

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserDefinedNewInstanceGenericTypeMapping.cs
```diff
@@ -32,6 +32,10 @@ ITypeSymbol objectType
 {
     public GenericMappingTypeParameters TypeParameters { get; } = typeParameters;
 
+    public bool DoesTypesSatisfySubstitutionPrinciples(SymbolAccessor symbolAccessor, ITypeSymbol sourceType, ITypeSymbol targetType) =>
```
latonz, line 35: IMO this method should not belong to the mapping itself but more to the mapping builder.

test/Riok.Mapperly.Tests/Mapping/GenericTest.cs
```diff
@@ -365,6 +365,41 @@ public void WithGenericTargetTypeConstraints()
             );
     }
 
+    [Fact]
+    public void WithGenericTargetSpecificSource()
```
latonz, line 369: Can you add the same the other way round (generic source, specific target)?

latonz (changes requested):
Thank you for this contribution 😊

docs/docs/configuration/generic-mapping.md
trejjam: I had an association with a math term substitution. Replace one with another without making it invalid. But I am not a native speaker either, so it can be a rubbish idea 😄

latonz (approved):

PR #1029, opened 2023-12-20
fix: handle nullable correctly in nullable disabled contexts for generics and arrays

latonz:
Fixes https://github.com/riok/mapperly/discussions/1025#discussion-5976133


ni507 (approved):

test/Riok.Mapperly.Tests/Mapping/UseMapperTest.cs
```diff
@@ -306,7 +306,7 @@ public void DisabledNullableShouldWork()
                 """
                 if (source == null)
                     return default;
-                var target = new global::B(_otherMapper.ToBExternal(source.Value));
+                var target = new global::B(source.Value != null ? MapToBExternal(source.Value) : default);
```
trejjam, line 309: I think this change is a bug. `MapToBExternal` should be called on the `_otherMapper`

latonz: Thanks for pointing out. How could we oversee this😕 I‘ll look into it.

latonz: Fixed in https://github.com/riok/mapperly/pull/1057

Issue #1050, opened 2024-01-06
Bug in nullability and external mappers


trejjam:
**Describe the bug**
I was looking for changes in new release, and I spotted this change:
https://github.com/riok/mapperly/pull/1029/files/0a2b80480a5fe55e06218fe5d0d5d7b8ba014f39#diff-3826207f8b2de11cdae56b06301dd51774e2dd3b94fb86e2aedc8e0b799344afR309

It changes `DisabledNullableShouldWork` https://github.com/riok/mapperly/blob/main/test/Riok.Mapperly.Tests/Mapping/UseMapperTest.cs#L309:
```cs
// from
var target = new global::B(_otherMapper.ToBExternal(source.Value));

// into
var target = new global::B(source.Value != null ? MapToBExternal(source.Value) : default);
```

**Expected behavior**
Which I believe should be

```cs
var target = new global::B(source.Value != null ? _otherMapper.MapToBExternal(source.Value) : default);
```

latonz:
@trejjam I'd appreciate your review on https://github.com/riok/mapperly/pull/1057 😊 (I can't request your review there...)

trejjam:
Happy to help :)

PR #1057, opened 2024-01-10
fix: upgrade parameter nullability correctly

latonz:
Fixes https://github.com/riok/mapperly/pull/1029/files/0a2b80480a5fe55e06218fe5d0d5d7b8ba014f39#r1443806308
Fixes https://github.com/riok/mapperly/issues/1050


latonz:
I think
`var target = new global::B(_otherMapper.ToBExternal(source.Value));`
(as it was before #1029) is correct, as `_otherMapper` is also in a nullable-disabled context. Therefore Mapperly considers all types as nullable (input parameter type as well as the return type). That again means that the null check of the expected code (`var target = new global::B(source.Value != null ? _otherMapper.MapToBExternal(source.Value) : default);`) in #1050 should not be needed.

ni507 (approved):

PR #1061, opened 2024-01-13
feat : add mapper attribute PreferParameterlessConstructors

### LINKED ISSUE #885: Configurable Constructor Priority
When trying to map properties to types that I don't own, it can be frustrating that Mapperly will always use the parameterless constructor. 

Here is an example that I've run into trying to map two nearly identical classes.

```c#
public class MyClass
{
    public string? Value1 { get; }
    public string? Value2 { get; }
    public string? Value3 { get; }
}

public class NotMyClass
{
    public NotMyClass() { }

    public NotMyClass(string? value1, string? value2, string? value3)
    {
        this.Value1 = value1;
        this.Value2 = value2;
        this.Value3 = value3;
    }

    public string? Value1 { get; }
    public string? Value2 { get; }
    public string? Value3 { get; }
}

[Mapper]
public static partial class MyClassMapper
{
    public static partial NotMyClass Map(MyClass myClass);
}
```

By default, the mapping of the above conversion from MyClass to NotMyClass will look like this:
```c#
public static partial class MyClassMapper
{
    public static partial global::Example.NotMyClass Map(global::Example.MyClass myClass)
    {
        var target = new global::Example.NotMyClass();
        return target;
    }
}
```
You can see that this conversation isn't really mapping anything, it's just creating an instance via the parameterless constructor. However, if I just comment out the default constructor, then all values are mapped perfectly.

```c#
public static partial class MyClassMapper
{
    public static partial global::Example.NotMyClass Map(global::Example.MyClass myClass)
    {
        var target = new global::Example.NotMyClass(myClass.Value1, myClass.Value2, myClass.Value3);
        return target;
    }
}
``` 

This behavior is well defined in the documentation:
```
Mapperly supports using parameterized constructors of mapping target types. Mapperly resolves the constructor to be used by the following priorities:
- accessible constructors annotated with MapperConstructor
- accessible parameterless constructors
- accessible constructors ordered in descending by their parameter count
- constructors with a System.ObsoleteAttribute attribute, unless they have a MapperConstructor attribute
```
The solution of either using the `MapperConstructor` or `Obsolete` attribute in order to use the desired constructor doesn't work in my case because I don't own the type that I'm mapping to.

Would it be possible to provide a mapper configuration that flips the second and the third bullet point so that I can prefer the constructor with the max number of parameters descending all the way down to zero parameters?

I'm happy to help work on this issue if that functionality sounds desirable and if you'd be able to point me to the correct area of the codebase.

R-abodyak:
# add mapper attribute PreferParameterlessConstructors
## Controls the priority of constructors used in mapping

 When enabled, a parameterless constructor is prioritized over constructors with parameters. When disabled, accessible constructors are ordered in descending order by their parameter count.

Fixes #885

## Checklist


src/Riok.Mapperly.Abstractions/MapperAttribute.cs
```diff
@@ -99,4 +99,11 @@ public class MapperAttribute : Attribute
     /// Determines the access level of members that Mapperly will map.
     /// </summary>
     public MemberVisibility IncludedMembers { get; set; } = MemberVisibility.AllAccessible;
+
+    /// <summary>
+    /// Controls the priority of constructors used in mapping.
+    ///  When enabled, a parameterless constructor is prioritized over constructors with parameters.
+    ///  When disabled, accessible constructors are ordered in descending order by their parameter count.
```
latonz, line 106: Nit: I'd use `<c>true</c>` and `<c>false</c>` instead of "enabled"/"disabled" as it directly links to the value.
Is there a reason for the space at the beginning of the line?
The same applies to the comment on `MapperConfiguration`.

src/Riok.Mapperly.Abstractions/MapperDefaultsAttribute.cs
```diff
@@ -7,4 +7,12 @@ namespace Riok.Mapperly.Abstractions;
...
 [Conditional("MAPPERLY_ABSTRACTIONS_SCOPE_RUNTIME")]
-public sealed class MapperDefaultsAttribute : MapperAttribute;
+public sealed class MapperDefaultsAttribute : MapperAttribute
+{
+    /// <summary>
+    /// Controls the priority of constructors used in mapping.
+    ///  When enabled, a parameterless constructor is prioritized over constructors with parameters.
+    ///  When disabled, accessible constructors are ordered in descending order by their parameter count.
+    /// </summary>
+    public bool PreferParameterlessConstructors { get; set; } = true;
```
latonz, line 17: As this extends the `MapperAttribute` this change can be reverted (the property is already inherited).

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
```diff
@@ -202,15 +202,23 @@ private static void BuildConstructorMapping(INewInstanceBuilderContext<IMapping>
...
         // ctors annotated with [Obsolete] are considered last unless they have a MapperConstructor attribute set
         var ctorCandidates = namedTargetType
             .InstanceConstructors.Where(ctor => ctx.BuilderContext.SymbolAccessor.IsDirectlyAccessible(ctor))
             .OrderByDescending(x => ctx.BuilderContext.SymbolAccessor.HasAttribute<MapperConstructorAttribute>(x))
-            .ThenBy(x => ctx.BuilderContext.SymbolAccessor.HasAttribute<ObsoleteAttribute>(x))
-            .ThenByDescending(x => x.Parameters.Length == 0)
-            .ThenByDescending(x => x.Parameters.Length);
+            .ThenBy(x => ctx.BuilderContext.SymbolAccessor.HasAttribute<ObsoleteAttribute>(x));
+
+        if (preferParameterlessConstructors)
```
latonz, line 213: IMO the parameter can be removed and replaced here directly with `ctx.BuilderContext.MapperConfiguration.PreferParameterlessConstructors`.

test/Riok.Mapperly.Tests/Helpers/MapperConfigurationBuilderTest.cs
```diff
@@ -36,6 +37,7 @@ public void ShouldMergeMapperConfigurations()
         mapper.UseReferenceHandling.Should().BeTrue();
         mapper.IgnoreObsoleteMembersStrategy.Should().Be(IgnoreObsoleteMembersStrategy.Source);
         mapper.RequiredMappingStrategy.Should().Be(RequiredMappingStrategy.Source);
+        mapper.PreferParameterlessConstructors.Should().BeTrue();
```
latonz, line 40: This assertion is failing isn't it?
This test ensures that the configuration is merged correctly. To test this set different values in the `defaultMapperConfiguration` as in the `NewMapperConfiguration` and assert whether the resulting value is the one from the `NewMapperConfiguration`.

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
```diff
@@ -27,6 +28,8 @@ public record TestSourceBuilderOptions(
     public static readonly TestSourceBuilderOptions WithDeepCloning = new(UseDeepCloning: true);
     public static readonly TestSourceBuilderOptions WithReferenceHandling = new(UseReferenceHandling: true);
 
+    public static readonly TestSourceBuilderOptions WithDescParametersConstructorPriority = new(PreferParameterlessConstructors: false);
```
latonz, line 31: Nitpick: IMO an easier to read and understandable name would be `PreferParameterizedConstructors` (same with the added test method name)

latonz (changes requested):
Thank you for your first contribution to Mapperly! I added my feedback, feel free to discuss any points or ask questions.

latonz:
The new public API can be added to the `PublicAPI.Unshipped.txt` file with the code-fix and then be moved to the `PublicAPI.Shipped.txt` file right away.

src/Riok.Mapperly.Abstractions/MapperDefaultsAttribute.cs
R-abodyak: I will remove it .
 but please, could you give more explanation about  (the property is already inherited) 
why the class MapperDefaultsAttribut is there while it contains nothing ?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/NewInstanceObjectMemberMappingBodyBuilder.cs
R-abodyak: yes , nice

test/Riok.Mapperly.Tests/Helpers/MapperConfigurationBuilderTest.cs
R-abodyak: okay I got that , so in this case `defaultMapperConfiguration  `has the value as `null `
and `NewMapperConfiguration  `has it as `true `, so the result is `true `

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
R-abodyak: done

src/Riok.Mapperly.Abstractions/MapperAttribute.cs
R-abodyak: done

src/Riok.Mapperly.Abstractions/MapperDefaultsAttribute.cs
latonz: The idea behind the separate `MapperDefaultsAttribute`  class is that we want to keep the option to introduce properties exclusive to `MapperDefaultsAttribute` or diverging properties without introducing a breaking change.

src/Riok.Mapperly.Abstractions/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,3 @@
 #nullable enable
+Riok.Mapperly.Abstractions.MapperAttribute.PreferParameterlessConstructors.get -> bool
```
latonz, line 2: Can you move these directly to the shipped file? Thank you!

latonz (changes requested):

latonz (changes requested):
A few linter checks are still failing, otherwise this looks good to me 😊 

https://github.com/riok/mapperly/actions/runs/7533542937/job/20521692711?pr=1061#step:4:140
https://github.com/riok/mapperly/actions/runs/7533542928/job/20521691733?pr=1061#step:8:11

R-abodyak:
Hi @latonz , can you check them now ? 
btw , is that something I can check locally or not ?

latonz:
You can run the linters locally. Find the commands in the workflow files:
https://github.com/riok/mapperly/blob/cdee78b4e9c003a44e45541a94987f1d365e90c3/.github/workflows/lint.yml#L18-L22
https://github.com/riok/mapperly/blob/cdee78b4e9c003a44e45541a94987f1d365e90c3/.github/workflows/docs.yml#L54-L59

latonz (approved):

PR #1069, opened 2024-01-18
feat: annotate all emitted methods and classes with GeneratedCodeAttribute.

### LINKED ISSUE #1038: Add the GeneratedCodeAttribute to generated code
Coverlet allows for exclusion of code having specific attributes, such as 

```
dotnet test "My.csproj" -p:ExcludeByAttribute=\"GeneratedCodeAttribute,CompilerGeneratedAttribute\"
```

But currently this attribute doesn't exist on the generated code, like other source generators such as the .NET Regex source generator.

Add GeneratedCodeAttribute to generated classes.

I've tried using file exclusions like:

```
dotnet test "My.csproj" -p:ExcludeByAttribute=\"GeneratedCodeAttribute,CompilerGeneratedAttribute\" -p:ExcludeByFile=\"**/*.g.cs\"
```

But mapperly generated code still shows up in coverage reports.

image

Along with messages the the related .g.cs file no longer exists.

skwasjer:
# Add `GeneratedCodeAttribute` to all generated methods and classes (except root/partial class).

## Description

- The attribute is added to generated methods, as well as all classes (except root/partial class).
- Refactored some related helper methods.
- Updated all snapshot tests. I did not deem it necessary to include unit tests because the coverage via the snapshots is extensive, but do let me know if you feel otherwise.
- ~Note: a lot of snapshot files had an UTF-8 BOM, but `.editorconfig` specifies UTF-8 (without BOM). Hence, for many files, line 1 is also part of the diff.~

Happy to adjust if needed.

Fixes #1038

## Checklist


skwasjer:
Now that I think about it, perhaps it is better to decorate individual methods instead, since the root mapper type is declared in user code?

docs/docs/configuration/generated-source.mdx
```diff
@@ -63,3 +63,8 @@ You won't see the updated mapper code or mapper diagnostics until you perform a
 This is done for performance reasons,
 otherwise the IDE could become laggy.
 :::
+
+## Excluding from code coverage
+
+Mapperly decorates each generated mapper type with the `GeneratedCodeAttribute`. This attribute can be used
```
latonz, line 69: It is probably also worth noting the `.g.cs` and the `/// <auto-generated />` comment.

latonz (changes requested):
Thank you for this contribution! The code looks good to me 😊
* Unit tests are not needed IMO.
* BOM in snapshots: Doesn't VerifyTests generate these files always with a BOM? Shouldn't we then adjust the .editorconfig to expect a BOM for snapshot files?

Regarding whether to annotate methods or classes: I'm not sure which is the right way to do it either... How do other source generators (especially the built in ones of .NET (logging, json, ...)) handle this?
Originally the issue was raised to be able to ignore the generated code with coverlet. How does it behave if the attribute is on the class and how if it is on the generated methods? (e.g. does user-implemented mapping coverage still work?)

skwasjer:
> How does it behave if the attribute is on the class and how if it is on the generated methods? (e.g. does user-implemented mapping coverage still work?)

This was exactly my concern. Annotating the class, due to it being partial will affect all type members. I was a bit quick in just slapping it on the classes because I had done so in the past for other generators, however in those cases the classes were non-partial.

> How do other source generators (especially the built in ones of .NET (logging, json, ...)) handle this?

Thanks for the hint. Looking at logger generated code, MS seems to indeed annotate type members individually. I am not sure why the static ctor is not decorated though. I will update the PR asap.

```cs
// source
public static partial class Log
{
    [LoggerMessage(
        EventId = 0,
        Level = LogLevel.Critical,
        Message = "Hello world!")]
    public static partial void HelloWorld(ILogger logger);
}

// generated
public static class Log
{
	[GeneratedCode("Microsoft.Extensions.Logging.Generators", "8.0.9.3103")]
	private static readonly Action<ILogger, Exception?> __HelloWorldCallback;

	[LoggerMessage(EventId = 0, Level = LogLevel.Critical, Message = "Hello world!")]
	[GeneratedCode("Microsoft.Extensions.Logging.Generators", "8.0.9.3103")]
	public static void HelloWorld(ILogger logger)
	{
		if (logger.IsEnabled(LogLevel.Critical))
		{
			__HelloWorldCallback(logger, null);
		}
	}

	static Log()
	{
		//IL_0011: Unknown result type (might be due to invalid IL or missing references)
		//IL_0016: Unknown result type (might be due to invalid IL or missing references)
		//IL_0023: Expected O, but got Unknown
		EventId eventId = new EventId(0, "HelloWorld");
		LogDefineOptions val = new LogDefineOptions();
		val.set_SkipEnabledCheck(true);
		__HelloWorldCallback = LoggerMessage.Define(LogLevel.Critical, eventId, "Hello world!", val);
	}
}
```

skwasjer:
> Shouldn't we then adjust the .editorconfig to expect a BOM for snapshot files?

Probably that's a good idea yea. Do you want me to include that in this PR or should this be a separate `style:` PR?

docs/docs/configuration/generated-source.mdx
skwasjer: I will add that and make it a bit more generic. It could indeed be used for any static code/security analysis tooling.

latonz:
A separate `style:` pr sounds good to me, thank you!

skwasjer:
@latonz as discussed, methods are now annotated. I removed the annotation for public/internal/protected classes but kept it for private/file classes. This is important for STA tools to not start complaining about unused/uncovered types (as potentially their method members are not analyzed).

Some extra considerations on this change:
- With future improvements/changes to Mapperly, there is a responsibility for the contributor and code reviewer to ensure the `CodeGeneratedAttribute` is applied to any generated type not explicitly declared by user code, and any of its generated type members (not just methods).
- If users of Mapperly have committed the generated code to an SCM, each subsequent version update of Mapperly that they pull will obviously now cause a diff due to the version number that has changed in the `GeneratedCodeAttribute`. Personally I don't see a problem with it, since it was a conscious choice in the first place to commit generated code. Not sure if it warrants a note in the docs though.

docs/docs/configuration/generated-source.mdx
skwasjer: Updated the documentation as per your suggestions, please check if it is satisfactory.

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.GeneratedCode.cs
```diff
@@ -0,0 +1,26 @@
...
+
+public partial struct SyntaxFactoryHelper
+{
+    private const string GeneratedCodeAttributeName = "global::System.CodeDom.Compiler.GeneratedCode";
+    private static readonly AssemblyName _generatorAssemblyName = typeof(SyntaxFactoryHelper).Assembly.GetName();
+
+    public SyntaxList<AttributeListSyntax> GeneratedCodeAttributeList()
+    {
+        var arguments = CommaSeparatedList(
+            AttributeArgument(LiteralExpression(SyntaxKind.StringLiteralExpression, Literal(_generatorAssemblyName.Name))),
```
latonz, line 17: Nitpick: There is a helper for string literals: `StringLiteral(_generatorAssemblyName.Name)`

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.GeneratedCode.cs
```diff
@@ -0,0 +1,26 @@
...
+            AttributeArgument(LiteralExpression(SyntaxKind.StringLiteralExpression, Literal(_generatorAssemblyName.Version.ToString())))
+        );
+
+        var attribute = Attribute(IdentifierName(GeneratedCodeAttributeName))
+            .WithArgumentList(AttributeArgumentList(CommaSeparatedList(arguments)));
+
+        return SingletonList(AttributeList(SingletonSeparatedList(attribute)).AddTrailingLineFeed(Indentation));
```
latonz, line 24: I think it would make sense to create a new helper method in `SyntaxFactoryHelper.Attribute.cs` which creates the attribute list (e.g. 
```c#
SyntaxList<AttributeListSyntax> AttributeList(string name, params Expression[] arguments)
{
    var arguments = CommaSeparatedList(arguments.Select(x => new AttributeArgument(x));

    var attribute = Attribute(IdentifierName(name)).WithArgumentList(AttributeArgumentList(CommaSeparatedList(arguments)));

    return SingletonList(AttributeList(SingletonSeparatedList(attribute)).AddTrailingLineFeed(Indentation));
}
```

This could also be used for `SyntaxFactoryHelper.UnsafeAccessorAttributeList`.

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.SymbolDeclaration.cs
```diff
@@ -17,13 +17,19 @@ public NamespaceDeclarationSyntax Namespace(string ns)
 
     public ClassDeclarationSyntax Class(string name, SyntaxTokenList modifiers, SyntaxList<MemberDeclarationSyntax> members)
     {
+        var isVisible = modifiers.Any(
```
latonz, line 20: Instead of checking whether it is public, wouldn't it make more sense to check whether it is not partial and add the generated code attribute to all non-partial classes? A partial class usually contains content at other locations which we can't be sure of that it is generated code (it usually is the declaration to the generated code...). All non partial classes are completely generated by Mapperly.

latonz (changes requested):
Thank you for the updates, I added a few small feedback points on the code.

Regarding the points you mentioned:
* Since the class generating syntax is in a central helper method, I think the risk of forgetting this is acceptable and should also be quite low. If it ever happens in the future we can consider implementing arch tests for the generated code (e.g. on the generated code of the integration tests) to ensure it can't happen (similar as we do for the attributes in `Riok.Mapperly.Abstractions` already)
* IMO the code changes on the attributes with each new version of Mapperly are acceptable. This attribute was designed to include the verison number for some reason. Also I don't know a lot of repositories which check in generated code. If this should rise a lot of feedback / issues we can think of an MSBuild option to disable the attribute.

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.SymbolDeclaration.cs
skwasjer: > wouldn't it make more sense to check whether it is not partial
Of course it would... :) 

Also addressed the other suggestions.

latonz:
Thank you for your contribution! 🥳

Issue #1033, opened 2023-12-24
Ignoring a property based on the attribute on the property instead of the mapper attribute


moxplod:
I am using MapDerivedType extensively. It is much cleaner to specify an attribute on the property that this property is read-only or ignore this property for mapping. 

Potentially use System.Components.ReadOnlyAttribute. 

This should ignore the property on the entity and not map from DTO to entity.

```

    [MapDerivedType<CompanyDto, Company>]
    [MapDerivedType<ContactDto, Contact>]
    [MapDerivedType<UserDto, User>]
    private static partial void MapDtoToEntity(EntityDto dto, IEntity entity);

```
```

[ReadOnly(true)]
    public string ReadOnlyText { get; set; }

```

moxplod:
My use case is for system/calculated fields. That I do not want the users to modify. But I do want to show them those fields.

latonz:
Relates to https://github.com/riok/mapperly/issues/585. I'm not sure about configuring mappings on the entities themselves. With the current approach all configuration of the mapper is visible within its class.

moxplod:
Yea - it is related. But in my scenario, using the current approach leads to compiler warnings as I am using MapDerivedType. 

The ignored fields are only going to be present in one of the child classes. 

For example, I want to ignore Risk property in the Company class. But that does not exist in Contact, that will throw a warning. 

![image](https://github.com/riok/mapperly/assets/4554395/2315e913-bf4c-40c9-84de-7be42de0729f)

latonz:
Currently configuration is not supported on methods using `MapDerivedType` (see also https://github.com/riok/mapperly/issues/910). You would need to configure the ignores on a mapping method for each mapping individually.

latonz:
Also related to #585.

PR #1070, opened 2024-01-19
feat: Add option to disable automatic discovery of user implemented mapping method and to include/exclude specific user implemented mapping methods

latonz:
Fixes https://github.com/riok/mapperly/issues/1033


docs/docs/configuration/user-implemented-methods.mdx
```diff
@@ -22,6 +22,46 @@ public partial class CarMapper
 
 Whenever Mapperly needs a mapping from `TimeSpan` to `int` inside the `CarMapper` implementation, it will use the provided implementation.
 
+## Automatic user implemented mapping method discovery
+
+By default, user implemented mapping methods are discovered automatically.
```
CommonGuy, line 27: This is the new behavior, correct?

- If `AutoUserMappings = true` (default), then all non-partial methods are considered as user mappings
  - Methods can be manually excluded by setting `[UserMapping(Ignore = true)]`
- If `AutoUserMappings = false`, then users must manually mark methods with `[UserMapping]` for Mapperly to pick them up

correct? I think this is not immediately obvious from the docs.

IMO this paragraph just needs slight adjustments to explain this a little better. What threw me off was the sentence
> To mark methods as user implemented mapping methods the `UserMappingAttribute` has to be applied.

which is only relevant when `AutoUserMappings = false`

latonz: Adjusted

docs/docs/configuration/user-implemented-methods.mdx
```diff
@@ -22,6 +22,46 @@ public partial class CarMapper
...
+This can be disabled by setting `AutoUserMappings` to `false`.
+With disabled `AutoUserMappings`, only methods marked by the `UserMappingAttribute` hare discovered by Mapperly.
```
CommonGuy, line 29: hare -> are

Maybe also change
> With disabled `AutoUserMappings`

 to
 
 >With `AutoUserMappings` disabled
 
 Not quite sure which one is more correct

latonz: Adjusted 🙈

CommonGuy (approved):

Issue #1008, opened 2023-12-15
PrimaryMappingAttribute to avoid accidentally using the wrong method for mapping an entity.


PaulVrugt:
**Is your feature request related to a problem? Please describe.**
See https://github.com/riok/mapperly/discussions/1007 . The current implementation of mapping priority (taking the first matching mapping in the file) is unclear and error prone

**Describe the solution you'd like**
I'd like to have an attribute PrimaryMappingAttribute added to the library, that is required to be used (or at least emits a diagnostic) when more than 1 method matches a mapping.

**Describe alternatives you've considered**
I don't see any alternatives. Having it being optional doesn't make sense because you can break your mapping without knowing a different mapper is used, simply because of the position in the file you placed the method.

**Additional context**
https://github.com/riok/mapperly/discussions/1007

latonz:
This is released on the next channel as version `3.5.0-next.2`. It can be used by applying `[UserMapping(Default = true)]`. Any feedback would be appreciated 😊

PaulVrugt:
Let me start by thanking you for taking the effort to fix this issue.
I haven't used the new version in practice, but if I understand correctly how it has now been implemented, I don't think this is very user friendly.

To be able to warnings/errors about ambiguous mappers, I now need to disable the `AutoUserMappings` property on ALL my mappers, and manually mark all user mappers. This has a few issues for me:

- The issue with ambiguous mapping methods is very rare, so I have to write extra code for each mapper a case that rarely happens
- The solution is still very error prone. Because the issue is rare, developers have no indication that they did something wrong if they do not disable the `AutoUserMappings` property. If they later add user mapper method, there is still now indication that it might be using the wrong mapper method.

I think, what we need, is being able to simply leave `AutoUserMappings` enabled, but have the `RMG060` emitted whenever it encounters an ambiguous user mapping. This can then be resolved by marking a user mapping as default.

Please let me know if I misunderstood something, because this doesn't seem like the most robust solution

latonz:
Thank you a lot for your feedback 😊 
You can actually use the `MapperDefaultsAttribute` on the assembly to set `AutoUserMappings` for all mappers in an assembly. See [docs](https://mapperly.riok.app/docs/configuration/mapper/#default-mapper-configuration).
[...]

PR #1071, opened 2024-01-19
feat: allow user implemented mappings to be marked as default

latonz:
Fixes https://github.com/riok/mapperly/issues/1008

* Preparation for https://github.com/riok/mapperly/pull/1072
* `RMG060` default severity is info for now, since otherwise this would result in lots of warnings for before-/aftermap. This should be changed with the next breaking change (see https://github.com/riok/mapperly/issues/1131).


latonz:
I'm not sure whether or not we should consider the first user implemented mapping as default if `AutoUserMappings = false`, wdyt @CommonGuy?

CommonGuy:
Do we really need this? Let's take this as an example:

```
[UserMapping]
public static int Method1(int s) => s;

[UserMapping(Default = true)]
public static int Method2(int s) => s;

// some partial mapping method would follow here
```

Is there any scenario where `Method1` will ever be used as a user mapping by Mapperly? As far as I can see, `Method2` will always be picked. We could use `[UserMapping(Ignore = true)]` on `Method1` here, which would have the same effect.

But maybe there are some scenarios (generic mapping method?) where multiple user methods could apply? If that is the case, then this new feature makes sense, but those scenarios should be used in the tests instead of mapping from `int` to `int`.

Instead of using a `Default` property, maybe an `Order` or `Priority` property (default 0, higher priority is picked first) would make more sense? Let's imagine three user implemented methods:
- Method1 with Priority = 1
- Method2 with Priority = 2
- Method3 with Priority = 3

For a partial method "MyMethod1", all three methods are suitable, so Method 3 is picked. For another partial method "MyMethod2", only Method1 and Method2 are suitable, so Method2 is picked. This wouldn't work if only `Default = true` could be defined on Method3.

----

Regardless of implementation, IMO the behavior should be:

- One matching method -> everything is fine
- Multiple matching methods, one has a default/best priority -> everything is fine
- Multiple matching methods with default=false/priority=0 -> emit a warning, even though this is a slightly breaking change if someone has "Report Warnings as Errors" enabled. Relying on method ordering is dangerous and this is fixed easily 
- Multiple matching methods with multiple `Default = true`/same priority >0 -> emit an error

I don't think there is an advantage to making `Default` a nullable boolean, could default it to false.

CommonGuy (approved):

PR #1090, opened 2024-01-28
fix: do not resolve existing one sided loose nullable property mappings

### LINKED ISSUE #1089: Mapping a collection property of nullable type elements affects other non-related mapping methods
**Bug description:**

In case a source class has a collection property of nullable type elements, while in the destination class the type of the collection property elements is non-nullable, and a mapper method is created to map those classes, then it also changes all the other generated mapping methods, where the source and destination classes have a property of the same nullable type.
Note that this issue does not occur, when the other mapping methods accept the source class as a non-nullable parameter and also have a non-nullable return type.

**Declaration code:**

The source classes:
```csharp
public sealed class SourceA
{
    public int? PropertyA { get; set; }
}
```

```csharp
public sealed class SourceB
{
    public List<int?> PropertyB { get; set; } = new();
}
```

The destination classes:

```csharp
public sealed class DestinationA
{
    public int? PropertyA { get; set; }
}
```

```csharp
public sealed class DestinationB
{
    public List<int> PropertyB { get; set; } = new();
}
```

The mapper methods:

```csharp
[Mapper]
public static partial class MapperClass
{
    public static partial DestinationA? MapToDestinationA(this SourceA? model);

    public static partial DestinationB MapToDestinationB(this SourceB model);
}
```

**Actual generated code:**

```csharp
// <auto-generated />
#nullable enable
namespace MapperlyTestProject
{
    public static partial class MapperClass
    {
        public static partial global::MapperlyTestProject.Destinations.DestinationA? MapToDestinationA(this global::MapperlyTestProject.Sources.SourceA? model)
        {
            if (model == null)
                return default;
            var target = new global::MapperlyTestProject.Destinations.DestinationA();
            target.PropertyA = model.PropertyA == null ? throw new System.ArgumentNullException(nameof(model.PropertyA)) : model.PropertyA.Value;
            return target;
        }

        public static partial global::MapperlyTestProject.Destinations.DestinationB MapToDestinationB(this global::MapperlyTestProject.Sources.SourceB model)
        {
            var target = new global::MapperlyTestProject.Destinations.DestinationB();
            target.PropertyB = MapToList(model.PropertyB);
            return target;
        }

        private static global::System.Collections.Generic.List<int> MapToList(global::System.Collections.Generic.List<int?> source)
        {
            var target = new global::System.Collections.Generic.List<int>(source.Count);
            foreach (var item in source)
            {
                target.Add(item == null ? throw new System.ArgumentNullException(nameof(item)) : item.Value);
            }
            return target;
        }
    }
}
```

**Expected generated code:**

```csharp
// <auto-generated />
#nullable enable
namespace MapperlyTestProject
{
    public static partial class MapperClass
    {
        public static partial global::MapperlyTestProject.Destinations.DestinationA? MapToDestinationA(this global::MapperlyTestProject.Sources.SourceA? model)
        {
            if (model == null)
                return default;
            var target = new global::MapperlyTestProject.Destinations.DestinationA();
            target.PropertyA = model.PropertyA;
            return target;
        }

        public static partial global::MapperlyTestProject.Destinations.DestinationB MapToDestinationB(this global::MapperlyTestProject.Sources.SourceB model)
        {
            var target = new global::MapperlyTestProject.Destinations.DestinationB();
            target.PropertyB = MapToList(model.PropertyB);
            return target;
        }

        private static global::System.Collections.Generic.List<int> MapToList(global::System.Collections.Generic.List<int?> source)
        {
            var target = new global::System.Collections.Generic.List<int>(source.Count);
            foreach (var item in source)
            {
                target.Add(item == null ? throw new System.ArgumentNullException(nameof(item)) : item.Value);
            }
            return target;
        }
    }
}
```

**Environment:**

- Mapperly Version: 3.3.0
- .NET Version: 8.0.101
- Target Framework: .NET 8.0
- Compiler Version: 4.8.0-7.23572.1 (7b75981c)
- C# Language Version: 12.0
- IDE: Visual Studio Community 2022, v17.8.5
- OS: Windows 10

**Additional context:**

As you can see, there is an unnecessary null check while mapping the `PropertyA` in the generated `MapToDestinationA` method, and this change was introduced in the generated `MapToDestinationA` method by creating the `MapToDestinationB` method. Making a new mapper method should not affect my previously created mapper methods, which have no relation at all with the new mapper method.

It is worth to mention that the order of the mapping methods in the `MapperClass` also affects the generated code.
In case I introduce a new `MapToDestinationC` method **after** the `MapToDestinationB` method, then it 'fixes' the source generated part of the `MapToDestinationA` method as well:

```csharp
public sealed class SourceC
{
    public int? PropertyC { get; set; }
}
```

```csharp
public sealed class DestinationC
{
    public int? PropertyC { get; set; }
}
```

```csharp
[Mapper]
public static partial class MapperClass
{
    public static partial DestinationA? MapToDestinationA(this SourceA? model);

    public static partial DestinationB MapToDestinationB(this SourceB model);

    public static partial DestinationC? MapToDestinationC(this SourceC? model);
}
```

```csharp
// <auto-generated />
#nullable enable
namespace MapperlyTestProject
{
    public static partial class MapperClass
    {
        public static partial global::MapperlyTestProject.Destinations.DestinationA? MapToDestinationA(this global::MapperlyTestProject.Sources.SourceA? model)
        {
            if (model == null)
                return default;
            var target = new global::MapperlyTestProject.Destinations.DestinationA();
            target.PropertyA = model.PropertyA;
            return target;
        }

        public static partial global::MapperlyTestProject.Destinations.DestinationB MapToDestinationB(this global::MapperlyTestProject.Sources.SourceB model)
        {
            var target = new global::MapperlyTestProject.Destinations.DestinationB();
            target.PropertyB = MapToList(model.PropertyB);
            return target;
        }

        public static partial global::MapperlyTestProject.Destinations.DestinationC? MapToDestinationC(this global::MapperlyTestProject.Sources.SourceC? model)
        {
            if (model == null)
                return default;
            var target = new global::MapperlyTestProject.Destinations.DestinationC();
            target.PropertyC = model.PropertyC;
            return target;
        }

        private static global::System.Collections.Generic.List<int> MapToList(global::System.Collections.Generic.List<int?> source)
        {
            var target = new global::System.Collections.Generic.List<int>(source.Count);
            foreach (var item in source)
            {
                target.Add(item == null ? throw new System.ArgumentNullException(nameof(item)) : item.Value);
            }
            return target;
        }
    }
}
```

latonz:
Removes resolving an existing one sided nullable mapping for a property mapping. 
E.g. when resolving `int?` => `int?` existing mappings for
1. `int?` => `int?`
2. `int` => `int?`
3. `int?` => `int`
4. `int` => `int`
were tried to be resolved. However, the one sided non-nullable version does introduce #1089 (uses a mapping built for other properties which can throw) and I can't think of a real use-case for this. Now only existing mappings for `int?` => `int?` and `int` => `int`  are resolved.

This is a breaking change, however, I can't think of a real use-case for this and it was only introduced in the last release. Not sure if we should release this as a major release.

Fixes #1089


latonz:
@TimothyMakkison can you think of any use-cases which this one-sided mapping resolving addresses?
Do you see additional problems where the same problem as described in #1089 could arise (e.g. with the `null` handling configurations on the `MapperAttribute`)?

ni507 (approved):

PR #1129, opened 2024-02-23
feat: add diagnostic if no members are mapped

### LINKED ISSUE #660: Emit a diagnostic if an object mapping maps no member
Report a diagnostic if an object mapping maps no member. The default level should be warning.

Rel. https://github.com/riok/mapperly/discussions/659

latonz:
Add ExpectNoMemberMappingsAttribute to expect no member mappings and reports a diagnostic if no members are mapped but mappings are expected.

Fixes #660


src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -572,4 +572,36 @@ public static class DiagnosticDescriptors
...
+    );
+
+    private static string BuildHelpUri(string id)
+    {
+#if ENV_NEXT
+        var host = "next.mapperly.riok.app";
+#elif ENV_LOCAL
+        var host = "localhost:3000";
+#else
+        var host = "next.mapperly.riok.app";
```
CommonGuy, line 602: Shouldn't this be `mapperly.riok.app`?

docs/docs/configuration/analyzer-diagnostics/RMG066.mdx
```diff
@@ -0,0 +1,29 @@
...
+```csharp
+public partial Target Map(Source source)
+{
+    var target = new Target();
+    return target;
+}
+```
+
+Usually this is not expected.
+If this is expected it can be suppressed for a mapping by applying `[ExpectMemberMappings(false)]`
```
CommonGuy, line 21: Does it make sense to create a new attribute for this, increasing the API surface?

Users could simply ignore this warning by suppressing it, couldn't they?

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
latonz: Yep fixed.

CommonGuy (approved):

Issue #585, opened 2023-07-21
[MapperIgnore] on properties of classes


mutzl:
A common usecase is, that entities have properties that never need to be mapped to any DTO.
Some of these metadata properties might even be defined in a base class.
For example:

``` c#
public abstract class BaseEntity
{
    public DateTime CreatedDate { get; set; }
    public string CreatedBy { get; set; } = default!;
    public DateTime? UpdatedDate { get; set; }
    public string? UpdatedBy { get; set; }
}

public class MyEntity : BaseEntity 
{
    public string Foo { get; set; }
}

public class MyDto 
{
    public string Foo { get; set; }
}
```

So the mapper configuration is always pretty verbose:
``` c#
    [MapperIgnoreTarget(nameof(BaseEntity.CreatedDate))]
    [MapperIgnoreTarget(nameof(BaseEntity.CreatedBy))]
    [MapperIgnoreTarget(nameof(BaseEntity.UpdatedDate))]
    [MapperIgnoreTarget(nameof(BaseEntity.UpdatedBy))]
    public static partial MyEntity ToEntity(this MyDto dto);

    [MapperIgnoreSource(nameof(BaseEntity.CreatedDate))]
    [MapperIgnoreSource(nameof(BaseEntity.CreatedBy))]
    [MapperIgnoreSource(nameof(BaseEntity.UpdatedDate))]
    [MapperIgnoreSource(nameof(BaseEntity.UpdatedBy))]
    public static partial MyDto ToDto(this MyEntity entity);
```

A MapperIgnoreAttribute - similar to the well known `[JsonIgnore]` - directly on the property would make code way cleaner.

``` c#
public abstract class BaseEntity
{
    [MapperIgnore] public DateTime CreatedDate { get; set; }
    [MapperIgnore] public string CreatedBy { get; set; } = default!;
    [MapperIgnore] public DateTime? UpdatedDate { get; set; }
    [MapperIgnore] public string? UpdatedBy { get; set; }
}
```

TimothyMakkison:
Would #513 and #201 help?

mutzl:
> Would #513 and #201 help?

#201 would help, I suppose.

The `[MapperIgnore]` looks way more intuitive to me, though. 
(at least for the use case I described above - for this kind of metadata, I know upfront, that it's never gonna be mapped).


BTW: Und vielen Dank für diese großartige Idee und Umsetzung von mapperly!

TimothyMakkison:
[...]

PR #1143, opened 2024-03-01
feat: Added MapperIgnoreMemberAttribute

jjr2000:
# Added MapperIgnoreMemberAttribute
## Description

Added  MapperIgnoreMemberAttribute to allow mapping to be ignored from within a source/target class. 
This is a small feature as it is meerly a rehashing of the ObsoleteAttribute implementation when it is set to Both, without the configuration due to being explicitly specified.

This will reduce require attribution in bi-directional mappings where you want to explicitly exclude a member from all mappings.

I believe I have updated all required code and tests. I however have not updated to documentation as I am not quite sure where the best place would be for it.

Closes https://github.com/riok/mapperly/issues/585

## Checklist


src/Riok.Mapperly.Abstractions/MapperIgnoreMemberAttribute.cs
```diff
@@ -0,0 +1,10 @@
+using System.Diagnostics;
+
+namespace Riok.Mapperly.Abstractions;
+
+/// <summary>
+/// Ignores a member from the mapping.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field, AllowMultiple = true)]
+[Conditional("MAPPERLY_ABSTRACTIONS_SCOPE_RUNTIME")]
+public sealed class MapperIgnoreMemberAttribute : Attribute;
```
latonz, line 10: Naming to be discussed.
IMO `MapperIgnoreAttribute` could be a name which would better match expectations as other serialization frameworks (e.g. SystemTextJson, Newtonsoft Json) use `JsonIgnoreAttribute` too.

`MapperIgnoreAttribute` was previously used by Mapperly (deprecated with v3.0 and removed with v3.4). I think it should be safe to reuse the same attribute name as due to the `AttributeUsage` is different (only Method was allowed before).

test/Riok.Mapperly.Tests/Mapping/IgnoreMember.cs
```diff
@@ -0,0 +1,82 @@
...
+        """
+    );
+
+    [Fact]
+    public void ClassAttributeIgnoreMember()
+    {
+        var source = TestSourceBuilder.Mapping("A", "B", TestSourceBuilderOptions.Default, _classA, _classB);
+
+        TestHelper
+            .GenerateMapper(source, TestHelperOptions.AllowInfoDiagnostics)
```
latonz, line 38: Are there actually any info diagnostics reported? If so, these should be asserted...

latonz (changes requested):
Thank you for contributing to Mapperly! Looks almost perfect already 😊 

Since this is your first contribution to a riok project I need to approve each GH action run, sorry for the delays...

I added a few feedback points, feel free to comment or update the PR as needed 😊 

Regarding the documentation: I think adding a new H4 in `mapper.mdx` under `Ignore properties / fields` just before `Ignore obsolete members` should do the job.

jjr2000:
Requested changes have now been incorporated and the docs have been updates.

docs/docs/configuration/mapper.mdx
```diff
@@ -68,6 +68,37 @@ public partial class CarMapper
...
+}
+
+public class CarDto
+{
+    // highlight-start
+    [MapperIgnore]
+    // highlight-end
+    public int MakeId { get; set; }
+
+    public string modelName { get; set; }
```
latonz, line 98: To be consistent probably use the same casing as in `Car`: `ModelName` (with a capital M).

docs/docs/configuration/mapper.mdx
```diff
@@ -68,6 +68,37 @@ public partial class CarMapper
 }
 ```
 
+#### IgnoreAttribute
```
latonz, line 71: The documentation of Mapperly usually uses describing titles instead of the names of the specific API. Probably `Ignore a member at definition`

test/Riok.Mapperly.Tests/Mapping/IgnoreMember.cs
latonz: This is not yet fixed, is it? Either remove the `TestHelperOptions.AllowInfoDiagnostics` parameter or assert the reported diagnostics...

docs/docs/configuration/mapper.mdx
```diff
@@ -68,6 +68,37 @@ public partial class CarMapper
...
+
+    public string ModelName { get; set; }
+}
+
+public class CarDto
+{
+    // highlight-start
+    [MapperIgnore]
+    // highlight-end
+    public int MakeId { get; set; }
```
latonz, line 96: I think the example would be easier to understand if this property is removed. Otherwise one could be thinking the `MapperIgnoreAttribute` is required on both sides, the source and the target class.

latonz (changes requested):
Thank you very much for the updates. I added my feedback on the documentation. Once this is improved it should be ready to get merged 😊

PR #1158, opened 2024-03-08
docs: Updated link to localhost with relative path

stbychkov:
# Update Localhost Link to Relative Path

## Description

* Navigate to the [Existing target](https://mapperly.riok.app/docs/configuration/existing-target/) page;
* Locate `null value handling` link on the bottom with the localhost address.

image

## Checklist

- [x] The documentation is updated

### DIFF
--- a/docs/docs/configuration/existing-target.md
+++ b/docs/docs/configuration/existing-target.md
@@ -73,4 +73,4 @@ static partial class FruitMapper
   </TabItem>
 </Tabs>
 
-See also [null value handling](./mapper/#null-values).
+See also [null value handling](./mapper.mdx/#null-values).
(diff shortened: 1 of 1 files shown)


docs/docs/configuration/existing-target.md
```diff
@@ -73,4 +73,4 @@ static partial class FruitMapper
   </TabItem>
 </Tabs>
 
-See also [null value handling](http://localhost:3000/docs/configuration/mapper/#null-values).
+See also [null value handling](./mapper/#null-values).
```
latonz, line 76: I think the correct link would be `[null value handling](./mapper.mdx#null-values)`.

latonz (changes requested):

docs/docs/configuration/existing-target.md
stbychkov: Done

latonz (approved):

Issue #953, opened 2023-11-27
Attribute controlling IQueryable mapping


trejjam:
**Is your feature request related to a problem? Please describe.**
It is valid to write the following:
```cs
dbContext.MyEntity
  .Select(x=> new MyDto
  {
    MyEnum = ConvertEnum(x.DbEnum),
  })
  .ToListAsync()

MyEnum ConvertEnum(DbEnum dbEnum) => (MyEnum) dbEnum;
```
as long as there is no database operation after (e.g. `.Select(...).Where(...)`)

**Describe the solution you'd like**
A new attribute, that will control whether other mapping functions may be called or not. Something like:
`[QueryableMapping(EQueryableMapping.AllowHydratationBeforeMapping)]`
and
`[QueryableMapping(EQueryableMapping.DatabaseServerOnly)]` (default)

**Additional context**
Right now, the only functionality that is blocked is ByName enum mapping. But for example, custom mappings or other mapping are allowed for other types.

The idea is with `AllowHydratationBeforeMapping` to allow ByName enum mapping with a mapping function.
With `DatabaseServerOnly` disable the use of custom mappers and other mappings.

The issue with AllowHydratationBeforeMapping is, that the whole entity is loaded and the mapping is called after. One of the advantages is that EF custom converters are executed.
With `DatabaseServerOnly` big entities with mapping to `Dto { public in Id { get; set; } }` are effective (no need to transfer throw-away properties).

latonz:
I don't really understand whats to impact of this attribute. Can you probably come up with some generated code samples for each of the settings?

trejjam:
Sure:

DTOs:
```cs
class A
{
    public B Value { get; set; }
}

class B
{
    public int Value { get; set; }
}

class C
{
    public D Value { get; set; }
}

class D
{
    public float Value { get; set; }
}
```

DatabaseServerOnly:
```cs
//mapping:
partial class Mapping
{
    [QueryableMapping(EQueryableMapping.DatabaseServerOnly)]
    public partial IQueryable<C> MapToC(this IQueryable<A> queryable);

    private D MatToD(B value)
    {
        return new D(value.Value);
    }
}

//generated:
partial class Mapping
{
    public partial IQueryable<C> MapToC(this IQueryable<A> queryable)
    {
        return queryable.Select(
            x => new C 
            {
                Value = new D
                {
                    Value = (float)x.Value.Value,
                }
            }
        );
    }
}
//diagnostics:
[...]

Issue #863, opened 2023-11-02
Queryable projections: Inline user-implemented mapping expressions


chrisevans9629:
**Is your feature request related to a problem? Please describe.**
I am trying to map a complex property in my projection, which can't be modified before/after the mapping due to the nature of how IQueryables work.

**Describe the solution you'd like**
Example:
if I had a car types in the database, something like below. I know you can add logic before and after the mapping, but that is more complicated with projections, as you can't modify with a select with essentially mapping twice. This is what I'm essentially trying to do:

```csharp
public class Car
{
    public double Weight { get; set; }
    public ModelT ModelT {get;set;}
    public ModelB ModelB {get;set;}
}

public class CarDto
{
    public double Mass { get; set; }
    public string CarType {get;set;}
}

[Mapper]
public partial class CarMapper
{
    public partial IQueryable<CarDto> Project(IQueryable<Car> query);
    
    [MapProperty(nameof(@Car.Weight), nameof(@CarDto.Mass))]
    [MapProperty("MethodNameToComputeCarType", nameof(@CarDto.CarType))]
    public partial CarDto Dto(Car dto);
  
   public string MethodNameToComputeCarType(Car c) => c.ModelT != null ? "ModelT" : "ModelB";
}


//Desired Result...
        public partial global::System.Linq.IQueryable<global::Moonshot.Agency.User.Mappers.CarDto> Project(global::System.Linq.IQueryable<global::Moonshot.Agency.User.Mappers.Car> query)
        {
#nullable disable
            return System.Linq.Queryable.Select(query, x => new global::Moonshot.Agency.User.Mappers.CarDto()
            {
                Mass = x.Weight,
                CarType = x.ModelT != null ? "ModelT" : "ModelB",
            });
#nullable enable
        }

```
**Describe alternatives you've considered**
An alternative solution could be adding a computed property to the EFCore entity (Car), which should work, but that has the disadvantage of bloating the entity when it's not needed for every query.

Another solution is to simply do this mapping in memory, which in some cases would reduce performance unnecessarily.

ysmoradi:
Hi
Happy new year!
Any updates on this?

latonz:
No updates. Feel free to contribute 😉 

This is dependent on https://github.com/riok/mapperly/issues/783.

latonz:
https://github.com/riok/mapperly/issues/783 handles the ability to map specific properties with a specific user-implemented mapping. This should soon land on main.

[...]

PR #1169, opened 2024-03-13
feat: Inline user implemented queryable mapping expressions

latonz:
Tries to inline user-implemented methods with an expression body into queryable expression mappings. This leads to better query mappings by EF Core and should result in less data loaded.
If the inlining fails for some reason (e.g. not a valid c# expression tree), RMG068 is reported.

This works with a syntax rewriter (`InlineExpressionRewriter`) which
* extends type names to their fully qualified names (since the using statements are not present in the generated code)
* expands extension method invocations (since the using statements are not present in the generated code)
* extracts invocations of other user-defined mappings, to be inlined by the mapping itself
* tries to assume whether the given code can be successfully inlined in a c# expression tree by checking the presence of given syntax declarations

Additionally a new mapping is added (`UserImplementedInlinedExpressionMapping`) which represents an inlined version of a user implemented mapping. When building the mapping, it rewrites the source parameter of  the user-method to the actual source of the mapping. Additionally mapping invocations extracted by the `InlineExpressionRewriter` are inlined.

Fixes https://github.com/riok/mapperly/issues/953.
Fixes https://github.com/riok/mapperly/issues/863.


src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedInlinedExpressionMapping.cs
```diff
@@ -0,0 +1,74 @@
...
+/// <param name="mappingInvocations">Mapping invocations to be inlined.</param>
+/// <param name="mappingBody">The prepared user written mapping body code (rewritten by <see cref="InlineExpressionRewriter"/>.</param>
+public class UserImplementedInlinedExpressionMapping(
+    UserImplementedMethodMapping userMapping,
+    ParameterSyntax sourceParameter,
+    IReadOnlyDictionary<SyntaxAnnotation, INewInstanceMapping> mappingInvocations,
+    ExpressionSyntax mappingBody
+) : NewInstanceMapping(userMapping.SourceType, userMapping.TargetType), INewInstanceUserMapping
+{
+    public IMethodSymbol Method => userMapping.Method;
```
CommonGuy, line 22: This class contains lots of code that isn't covered by tests. Probably a good idea to add more test coverage

docs/docs/configuration/queryable-projections.mdx
```diff
@@ -76,3 +74,48 @@ public static partial class CarMapper
     private static partial CarDto Map(Car car);
 }
 ```
+
+## User-implemented mapping methods
+
+Mapperly tries to inline user-implemented mapping methods.
+For this to work, user-implemented mapping methods need satisfy certain limitations:
```
CommonGuy, line 81: "need to satisfy"

docs/docs/configuration/queryable-projections.mdx
```diff
@@ -76,3 +74,48 @@ public static partial class CarMapper
...
+
+- Only expression-bodied methods can be inlined.
+- The body needs to follow the expression tree [limitations](https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/expression-trees/#limitations).
```
CommonGuy, line 84: Nitpick: Would add the link over the whole "expression tree limitations" text

docs/docs/configuration/queryable-projections.mdx
```diff
@@ -76,3 +74,48 @@ public static partial class CarMapper
...
+
+    // highlight-start
+    private static string MapCarBrandName(CarBrand brand)
+      => band.Name;
+    // highlight-end
+}
+```
+
+If Mapperly is unable to inline a user-implemented method, `RMG068` is reported.
+Non-inlined method invocations can cause the query to be materialized and more data than necessary could be loaded.
```
CommonGuy, line 101: Not sure if this sentence sounds correct? Maybe change to

> Non-inlined method invocations can lead to more data being loaded than necessary.

docs/docs/configuration/queryable-projections.mdx
```diff
@@ -76,3 +74,48 @@ public static partial class CarMapper
...
+[Mapper]
+public static partial class CarMapper
+{
+    public static partial IQueryable<CarDto> ProjectToDto(this IQueryable<Car> q);
+
+    private static partial CarModelDto MapCarModel(CarModel model);
+
+    // highlight-start
+    private static ICollection<CarModelDto> MapOrderedCarBrandName(ICollection<CarModel> models)
+      => models.OrderBy(x => x.Name).Select(x => MapCarModel(x)).ToList(); // note: do not use the method group
```
CommonGuy, line 115: Maybe change the comment to

```
note: do not use the method group '.Select(MapCarModel)', as that cannot be inlined
```

docs/docs/configuration/queryable-projections.mdx
```diff
@@ -76,3 +74,48 @@ public static partial class CarMapper
...
+    // highlight-end
+  }
+```
+
+It is important that the types in the user-implemented mapping method match the types of the objects to be mapped exactly.
+Otherwise, Mapperly may not resolve the user-implemented mapping methods.
```
CommonGuy, line 121: Nitpick: I think "cannot" is more appropriate here instead of "may not"

latonz: Improved

docs/docs/configuration/queryable-projections.mdx
latonz: Improved

docs/docs/configuration/queryable-projections.mdx
latonz: Improved

docs/docs/configuration/queryable-projections.mdx
latonz: Improved

docs/docs/configuration/queryable-projections.mdx
latonz: Improved

CommonGuy (approved):

Issue #884, opened 2023-11-09
Generic mapping not working for function with 2 parameters


mo-ah-dawood:
```c#
        private partial TTarget CreateEntity<TSource, TTarget>(TSource source);  /// working
        private partial void ModifyEntity<TSource, TTarget>(TSource source, TTarget target); /// not working
```

**Environment (please complete the following information):**
- Mapperly Version: 3.2.0
- .NET Version: 7.0.310
- Target Framework: net7.0
- IDE: vs code & vs 2022
- OS: macos 14

latonz:
Currently existing targets are not supported for generic mappings. Would be happy to accept a PR.

Ixonal:
Just came across this issue report after trying to set this up locally and realizing it doesn't work. Found it kind of odd, since both mapping to an existing implementation and generic mapping were both, themselves, implemented. Is the plan to only have this implemented via pull request or are y'all looking into it internally?

latonz:
@Ixonal no one of us is currently working on this, feel free to contribute 😉

Ixonal:
I'll see if i have time, but can't make any promises. Doesn't look like it'd be particularly complex at least.

smitty3268:
We need this feature too.

latonz:
Please use the 👍 reaction on an issue to show that you're interested in a feature. This way, we can sort the issues based on popularity without cluttering the comments.

Issue #625, opened 2023-08-05
Generic factory does not work if the generic member is generic it self


mo-ah-dawood:
### Not working
```c#
    public abstract class BaseDto<TEntity>  // this is generic
    {
        internal abstract TEntity CreateEntity();
    }

    [Mapper()]
    public partial class DtoMapper
    {

        [ObjectFactory]
        private TTarget Create<TSource, TTarget>(TSource source) where TSource : BaseDto<TTarget> //this is generic
        {
            return source.CreateEntity();
        }

    } 
```
    
    
   ### Working
   
   ```c#
    public abstract class BaseDto  //this is not generic
    {
        internal abstract object CreateEntity();
    }

    [Mapper()]
    public partial class DtoMapper
    {

        [ObjectFactory]
        private TTarget Create<TSource, TTarget>(TSource source) where TSource : BaseDto //this is not generic
        {
            ....
        }

    }
   ```

PR #1181, opened 2024-03-17
Implement generic queryable projection

### LINKED ISSUE #968: Support generic `IQueryable<T>` mapping
Support for generic `IQueryable<T>` mapping methods:
```c#
public static partial System.Linq.IQueryable<TTarget> Map<TSource, TTarget>(System.Linq.IQueryable<TSource> source);

private B Map(A source);
private D Map(C source);

// generates
public static partial System.Linq.IQueryable<TTarget> Map<TSource, TTarget>(System.Linq.IQueryable<TSource> source)
{
    return source switch
    {
        IQueryable<global::A> x when typeof(TTarget).IsAssignableFrom(typeof(global::B)) => source.Select(...),
        IQueryable<global::C> x when typeof(TTarget).IsAssignableFrom(typeof(global::D)) => source.Select(...),
        _ => throw new System.ArgumentException($"Cannot map {source.GetType()} to {typeof(IQueryable<TTarget>)} as there is no known type mapping", nameof(source)),
    };
}
```

Similar to the existing generic mapping methods ([docs](https://mapperly.riok.app/docs/configuration/generic-mapping/)), type constraints should also be supported.

Brought up by @19bartek92 in https://github.com/riok/mapperly/discussions/960#discussion-5910036

mrblumi:
# Add support for generic IQueryable<T> projections

## Description

Right now this is only supposed to be a draft - since this is my first PR here at github ever, please provide initial feedback. Adjustment of documentation, comments, more tests and code style are yet to come. Any hints on how to better reuse existing code are welcome as well, cause I am not that familiar with code generators in common and this code base in detail yet :)

When merged this will include the functionality for generic projection mappings, i.e. `IQueryable<TTarget> ProjectTo<TTarget>(IQueryable querable)` and similar.

Fixes #968, #960


latonz:
Thank you very much for these efforts and this contribution!
The code looks pretty good overall 😊 👏 

But I'm not sure if this is the approach I'd take. This copies more or less the approach of `UserDefinedNewInstanceRuntimeTargetTypeMapping` and extracts the type argument of the queryable instead of using the entire type. When I created this issue based on the GH discussion I didn't invest too much time as I thought we need to rewrite the handling of generics anyway. 
All these issues are related:
https://github.com/riok/mapperly/issues/884
https://github.com/riok/mapperly/issues/625
https://github.com/riok/mapperly/issues/884
https://github.com/riok/mapperly/issues/451

IMO the way to go would be to implement generic type matching/binding similar to how the compiler does it. When this is implemented all the mentioned issues should be a lot simpler to implement. I experimented with it such an approach [here](https://github.com/riok/mapperly/compare/main...latonz:mapperly:625-nested-generic-object-factories) but never finished it... WDYT?

mrblumi:
Yes, as mentioned, this is right now more ore less a draft/poc for the functionality without any code deduplication and I am not very happy with my approach as well, because it is more ore less copy of the existing `UserDefinedNewInstanceRuntimeTargetTypeMapping` as you  already mentioned.

Clearly it could be cleaned up a lot to consolidate all the code duplicated by me right now without that much effort. But it would have needed some minor adjustments here and there and I was not sure if there would be a more straight forward way or if this was even the way you intended it to be...

To be honest, ATM I have no clue how the compiler does the generic type matching/binding internally (a lot of stuff I have to learn), but I will have a look at your approach and the internal compiler magic the next weekend.

Should I invest more time and enhance and consolidate the code from my approach from above then at all? I haven't tested every detail with the gneric constraints yet, but so far it seems to be working as intended. Is there a chance to have this as a "solution" until the more mature approach is implemented (I'd be happy to help here as well, if welcome...)

latonz:
Asking for feedback is always a good idea 😊 

Currently, I don't think this PR adds sustainable value to be worth merging (and it would add quite a bit of complexity to the Mapperly code). The use-case adressed can usually also be solved by the already implemented simple generic support with a mapping method without any type constraints or type constraints to the `IQueryable` without any type parameters (edit: I just noted another bug which prevents this from working: https://github.com/riok/mapperly/issues/1189). But again thank you a lot for your efforts!

I don't have a clue either how exactly Roslyn handles this 🙈 There is also an issue in the Roslyn repository to expose these APIs: https://github.com/dotnet/roslyn/issues/46998.

You can walk through my approach of [this](https://github.com/latonz/mapperly/tree/625-nested-generic-object-factories) attempt by starting with `TypeBinderFactory.TryCreate`. The idea is that the *Binder represent the generic type arguments and can check and bind actual types to them. A lot of work is still missing though.

PS: Help is always welcome on all the [issues](https://github.com/riok/mapperly/issues) 😊 (I just looked again at all issues and labeled some with less complexity as `good first issue`s). For big changes it is a good idea to quickly discuss the ideas for the solution in the issue to prevent unnecessary work.

PR #1191, opened 2024-03-20
fix: correctly assign null if target is nullable instead of throwing

### LINKED ISSUE #1182: ThrowOnPropertyMappingNullMismatch throws an exception for nullable properties
**Describe the bug**
The XML comment of `ThrowOnPropertyMappingNullMismatch` says:
```xml
/// <summary>
/// Specifies the behaviour in the case when the mapper tries to set a non-nullable property to a <c>null</c> value.
/// If set to <c>true</c> an <see cref="ArgumentNullException"/> is thrown.
/// If set to <c>false</c> the property assignment is ignored.
/// This is ignored for required init properties and <see cref="IQueryable{T}"/> projection mappings.
/// </summary>
```

If I understood that right, if I have a non-nullable property in my `target` and the mapper tries to set it to `null`, that will result in an exception. However, that implies that if the target property is nullable, its value should be simply set to null. In my hypothetical scenario, I have the case where I have a nullable property in my target, but the mapper still throws an exception when it tries to set its value to `null`.

In the scenario below, the `UserDTO.Parent` property is a nullable (`UserDTO?`) property, but the generated `else` statement by the mapper throws an exception if my source (`User.Parent`) is `null`. From what it looks to me, that should not be the case, or did I misunderstand the `ThrowOnPropertyMappingNullMismatch` XML comment?

I am aware of `AllowNullPropertyAssignment` existence, but using it in a combination with `ThrowOnPropertyMappingNullMismatch` still throws the exception in this case.

**Declaration code**
```csharp
public class UserDTO
{
    public long? ParentId { get; set; }
    public UserDTO? Parent { get; set; }
}

public class User
{
    public long? ParentId { get; set; }
    public User? Parent { get; set; }
}

[Mapper(ThrowOnPropertyMappingNullMismatch = true)]
public partial class Mapper
{
    public partial UserDTO ToUserDTO(User user);
}
```

```csharp
[global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "3.4.0.0")]
public partial global::UserDTO ToUserDTO(global::User user)
{
    var target = new global::UserDTO();
    if (user.Parent != null)
    {
        target.Parent = ToUserDTO(user.Parent);
    }
    else
    {
        throw new System.ArgumentNullException(nameof(user.Parent));
    }
    target.ParentId = user.ParentId;
    return target;
}
```

```csharp
[global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "3.4.0.0")]
public partial global::UserDTO ToUserDTO(global::User user)
{
    var target = new global::UserDTO();
    if (user.Parent != null)
    {
        target.Parent = ToUserDTO(user.Parent);
    }
    else
    {
        target.Parent = null;
    }
    target.ParentId = user.ParentId;
    return target;
}
```

- Mapperly Version: 3.4.0
- .NET Version: .NET 8.0.202
- Target Framework: net8.0
- Compiler Version: 4.9.0-3.24121.1 (a98c90d5)
- C# Language Version: 12.0
- IDE: Visual Studio v17.9.3
- OS: Windows 11

latonz:
Closes #1182

### DIFF
--- a/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyNullableTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/ObjectPropertyNullableTest.cs
@@ -1,3 +1,5 @@
+using Riok.Mapperly.Diagnostics;
+
 namespace Riok.Mapperly.Tests.Mapping;
 
 public class ObjectPropertyNullableTest
@@ -145,8 +147,8 @@ public class ObjectPropertyNullableTest
             "B",
             "class A { public C? Value { get; set; } }",
             "class B { public D Value { get; set; } }",
-            "class C { public string V {get; set; } }",
-            "class D { public string V {get; set; } }"
+            "class C { public string V { get; set; } }",
+            "class D { public string V { get; set; } }"
         );
 
         TestHelper
@@ -280,8 +282,8 @@ public class ObjectPropertyNullableTest
             "B",
             "class A { public C Value { get; set; } }",
             "class B { public D? Value { get; set; } }",
-            "class C { public string V {get; set; } }",
-            "class D { public string V {get; set; } }"
+            "class C { public string V { get; set; } }",
+            "class D { public string V { get; set; } }"
         );
 
         TestHelper
@@ -304,8 +306,8 @@ public class ObjectPropertyNullableTest
             "B",
             "class A { public C? Value { get; set; } }",
             "class B { public D? Value { get; set; } }",
-            "class C { public string V {get; set; } }",
-            "class D { public string V {get; set; } }"
+            "class C { public string V { get; set; } }",
+            "class D { public string V { get; set; } }"
         );
 
         TestHelper
@@ -339,8 +341,8 @@ public class ObjectPropertyNullableTest
             },
             "class A { public C? Value { get; set; } }",
             "class B { public D? Value { get; set; } }",
-            "class C { public string V {get; set; } }",
-            "class D { public string V {get; set; } }"
+            "class C { public string V { get; set; } }",
+            "class D { public string V { get; set; } }"
         );
 
         TestHelper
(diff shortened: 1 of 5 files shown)


test/Riok.Mapperly.Tests/Mapping/ObjectPropertyNullableTest.cs
```diff
@@ -516,6 +518,89 @@ TestSourceBuilderOptions.Default with
...
+        var source = TestSourceBuilder.Mapping(
+            "A",
+            "B",
+            TestSourceBuilderOptions.Default with
+            {
+                ThrowOnPropertyMappingNullMismatch = true
+            },
+            "class A { public C? Value { get; set; } }",
+            "class B { public D? Value { get; set; } }",
+            "class C { public string V {get; set; } }",
```
ni507, line 533: missing space `{get; set; }`

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyNullableTest.cs
```diff
@@ -516,6 +518,89 @@ TestSourceBuilderOptions.Default with
...
+            [MapProperty("Value.Flattened", "ValueFlattened")]
+            partial B Map(A source);
+            """,
+            TestSourceBuilderOptions.Default with
+            {
+                ThrowOnPropertyMappingNullMismatch = true
+            },
+            "class A { public C? Value { get; set; } }",
+            "class B { public D? Value { get; set; } public string ValueFlattened { get; set; } }",
+            "class C { public string V {get; set; } public string Flattened { get; set; } }",
```
ni507, line 575: missing space `{get; set; }`

ni507 (changes requested):

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyNullableTest.cs
latonz: fixed all of them 😊

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyNullableTest.cs
latonz: fixed all of them 😊

ni507 (approved):

PR #1230, opened 2024-04-09
Fix selecting user methods in generic methods

trejjam:
# Fix selecting user methods in generic methods

## Description

The following mapping:
```cs
partial TTarget Map<TTarget>(object source);

partial B MapToBInternal(A source);

B MapToB(A source) => MapToBInternal(source);
```

Generates the `Map` method using `MapToBInternal`.

## Note
This should be an issue, but I already wrote failing tests...
I am happy to finish it. I would like a hint where the solution should take place. I guess it should be in `UserMethodMappingExtractor.ExtractUserMappings`. I would use some kind of weighted list to select the correct method, but maybe there is another simpler solution.

## Checklist


trejjam:
ObjectFactory moved to #1234

latonz (approved):

latonz:
I just noticed the .net framework integration test is not yet updated, could you update that one? Thank you!

trejjam:
It should be ready now

latonz (approved):

PR #1233, opened 2024-04-11
fix: Inlined argument with full namespace

trejjam:
# The inlined method does not resolve arguments full namespace

## Description

The following mapping produces an invalid code:
```cs
private partial System.Linq.IQueryable<B> Map(System.Linq.IQueryable<A> source);

private static DateTimeOffset MapToDateTimeOffset(
    DateTime dateTime
) => new(dateTime, TimeSpan.Zero);
```
The argument was kept as-is, therefore, without namespace or using statements.

Before:
```cs
[global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "0.0.1.0")]
private partial global::System.Linq.IQueryable<global::B> Map(global::System.Linq.IQueryable<global::A> source)
{
#nullable disable
    return System.Linq.Queryable.Select(source, x => new global::B()
    {
        Value = new(x.Value, TimeSpan.Zero),
    });
#nullable enable
}
```

Expected:
```cs
[global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "0.0.1.0")]
private partial global::System.Linq.IQueryable<global::B> Map(global::System.Linq.IQueryable<global::A> source)
{
#nullable disable
    return System.Linq.Queryable.Select(source, x => new global::B()
    {
        Value = new(x.Value, global::System.TimeSpan.Zero),
    });
#nullable enable
}
```

## Checklist


src/Riok.Mapperly/Descriptors/InlineExpressionRewriter.cs
```diff
@@ -55,6 +55,17 @@ public override SyntaxNode VisitTypeArgumentList(TypeArgumentListSyntax node)
         return node.WithArguments(SeparatedList(args));
     }
 
+    public override SyntaxNode? VisitMemberAccessExpression(MemberAccessExpressionSyntax node)
+    {
+        if (semanticModel.GetSymbolInfo(node.Expression).Symbol is ITypeSymbol namedTypeSymbol)
+        {
+            var expression = FullyQualifiedIdentifier(namedTypeSymbol).WithTriviaFrom(node.Expression);
+            node = node.WithExpression(expression);
```
latonz, line 63: Wouldn't it safer to return the new node here directly instead of visiting it on L66? Otherwise this could result in an exception that the new node is not part of the syntax tree, couldn't it?

latonz (changes requested):
Thank you for this Fix 😊 

While looking at your example, I came up with the idea that we should also extend Target-Type new. I'm not sure if it really causes problems, but since the compilation context changes quite a bit by inlining these expressions it should be safer to just expand them... WDYT? (If we decide to fix this we should do it in another PR 😊)

src/Riok.Mapperly/Descriptors/InlineExpressionRewriter.cs
trejjam: I do not have an experience with visitor pattern implementation. I changed it to return directly

latonz (approved):

latonz:
Created a fix for the implicit new in https://github.com/riok/mapperly/pull/1238

PR #587, opened 2023-07-22
feat: add support for nested properties

zeidoo:
# Adding support for nested properties

## Description

(cont of https://github.com/riok/mapperly/pull/534)

If you don’t want explicitly name all properties from nested source bean, you can use MapperNestedPropertiesAttribute. This will tell Mapperly to map every property from source proerty to the target object.

Currently a WIP. Please let me know what to add for this to be merge eventually.
Checklist

- [x] The existing code style is followed
- [x] The commit message follows our guidelines
- [x] Performed a self-review of my code
- [ ] Hard-to-understand areas of my code are commented
- [ ] The documentation is updated (as applicable)
- [x] Unit tests are added/updated
- [ ] Integration tests are added/updated (as applicable, especially if feature/bug depends on roslyn or framework version in use)

PR #1245, opened 2024-04-22
feat: Include all members of nested member

### LINKED ISSUE #453: Include all members from a nested member

rhodon-jargon:
# Flatten all nested members of a specific property

## Description

The added `MapNestedPropertiesAttribute` allows users to flatten all properties of a specific member, to avoid having to manually specify the mapping of all nested properties of that member properties as shown in #453.

Partly based on #587, but re-implemented, keeping the comments on the PR into account. I took a different approach on the implementation, to explicitly handle nested properties differently from manually specified properties. The resulting member lookup priority is as follows:

1. Manually specified property mappings using `MapPropertyAttribute`.
2. Properties from the root object, which may or may not use automatic flattening.
3. Properties from the objects at any of the specified `MapNestedPropertiesAttribute`s, which may or may not use automatic flattening. Note that these objects are checked in no explicit order, though I suspect it will be based on the order the attributes are specified in. If multiple objects contain a member with the correct name, only the first is returned and no ambiguity diagnostic is emitted. This behavior can be changed in the future.

I thought this order would be most intuitive, and it also allows overriding the default behavior using `MapperIgnoreSourceAttribute`s.

Fixes #453

## Checklist


test/Riok.Mapperly.IntegrationTests/Models/TestObject.cs
```diff
@@ -45,6 +45,8 @@ public TestObject(int ctorValue, int unknownValue = 10, int ctorValue2 = 100)
 
         public TestObjectNested? NestedNullableTargetNotNullable { get; set; }
 
+        public TestObjectExtension? Extension { get; set; }
```
latonz, line 48: I'm not sure about the naming here... Extension in C# usually refer to a different concept (AFAIK there are also proposals to allow property extensions which would then be even more confusing here...). Probably use a name like `NestedObject` or similar? (Also applies to the name of the class and other properties in the integration test classes).

test/Riok.Mapperly.Tests/Mapping/ObjectNestedPropertyTest.cs
```diff
@@ -0,0 +1,261 @@
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class ObjectNestedPropertyTest
+{
+    private static readonly TestHelperOptions ignoreNestedMemberNotUsed =
```
latonz, line 8: `private static readonly` usually starts with an `_` in this code base.

test/Riok.Mapperly.Tests/Mapping/ObjectNestedPropertyTest.cs
```diff
@@ -0,0 +1,261 @@
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class ObjectNestedPropertyTest
```
latonz, line 6: Nice test cases 👍 

Could you add test cases for the following:
use `MapNestedProperties`...
* to map to a constructor parameter
* to map to a init only property
* in combination with ignored property casing
* in a queryable projection mapping (probably in another test class)
* with a nullable object (`MapNestedProperties("A")` where `A` is nullable)

test/Riok.Mapperly.Tests/Mapping/ObjectNestedPropertyTest.cs
```diff
@@ -0,0 +1,261 @@
...
+                """
+                var target = new global::B();
+                target.Id = source.Id;
+                return target;
+                """
+            );
+    }
+
+    [Fact]
+    public void NestedPropertyWithSourcePathName()
```
latonz, line 102: What is the difference between this test case and `NestedPropertyWithMemberNameOfSource`?

test/Riok.Mapperly.Tests/Mapping/ObjectNestedPropertyTest.cs
```diff
@@ -0,0 +1,261 @@
...
+            "class A { public string Id { get; set; } public C Value { get; set; } }",
+            "class B { public string Id { get; set; } }",
+            "class C { public string MyId { get; set; } }"
+        );
+
+        TestHelper
+            .GenerateMapper(source, TestHelperOptions.AllowDiagnostics)
+            .Should()
+            .HaveDiagnostic(DiagnosticDescriptors.NestedMemberNotUsed)
+            .HaveDiagnostic(DiagnosticDescriptors.SourceMemberNotMapped)
```
latonz, line 159: Assert the generated diagnostics at least once with a snapshot to ensure the message template is rendered correctly and the location of the diagnostic is correct.

latonz (changes requested):
Thank you very much for these high-quality PRs 🥳 
I added my feedback, feel free to discuss 😊 

Since this is your first contribution to a riok project, I need to approve the CI runs manually...
Currently the integration tests for the older .NET versions are failing. You can find the VerifyTest diffs in the GH Actions artifacts (see also integration [tests contributor documentation](https://next.mapperly.riok.app/docs/contributing/tests/#integration-tests)).

latonz (approved):

latonz:
Thank you for the updates 😊 

I usually don't run the older TFM tests locally. It could work by adjusting the `TargetFramework` of the integration tests. However, I'm not sure if the older Roslyn version is used if the target framework is an older version.

PR #1247, opened 2024-04-23
feat: Map from source object

### LINKED ISSUE #1161: Mapping from the source object
When mapping to a Target field there's often a need to access multiple fields from the Source and today this requires creating a custom mapping method like so:

```csharp
public Address OrderToAddress(Order order)
{
  var ret = MapToAddress(order);
  ret.FullName = string.Concat(order.FirstName, " ", order.LastName);
  return ret;
}

[MapperIgnoreTarget(nameof(Address.FullName))]
private partial Address MapToAddress(Order order);
```

I propose a new [MapSource] (or some other name) attribute that supports using the Source object as the 'from' parameter like so:

```csharp
[MapSource(nameof(Address.FullName))]
public partial Address OrderToAddress(Order order);

private string MapToFullName(Order order) => string.Concat(order.FirstName, " ", order.LastName);
```

rhodon-jargon:
# Allow mapping from source object

## Description
Adds a `MapPropertyFromSourceAttribute`, that allows using the source object directly for mappings. In most of the code, this is represented as an empty source `MemberPath`. This required a few changes, as most of the code just assumed `MemberPath`s were not empty. Currently, when a `MapPropertyFromSourceAttribute` is used, all unused source member diagnostics are suppressed, as there is no way to analyse for every situation what members of the source are used by the conversion. This can be changed to require explicit ignore attributes for all members that are used implicitly by the mapping from source.

An additional reason for wanting this, apart from the arguments discussed in #1161, is to allow these kind of mappings for required or init-only members.

Fixes #1161

## Checklist


src/Riok.Mapperly/Symbols/NonEmptyMemberPath.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.CodeAnalysis;
+
+namespace Riok.Mapperly.Symbols;
+
+public class NonEmptyMemberPath : MemberPath
+{
+    public NonEmptyMemberPath(IReadOnlyList<IMappableMember> path)
```
latonz, line 7: I understand what you are doing here... But doesn't this introduce weird semantics: an instance of `MemberPath` can be empty but doesn't have to (the class also contains a lot of logic which only is needed if it contains a non-empty path)? 🤔 

On a second thought: Wouldn't it be easier to special case the empty member path instead of the non-empty one? Idea to discuss:
* Remove the `NonEmptyMemberPath` and revert usage back to `MemberPath`
* Add a new `RootObjectMemberPath : MemberPath` (or maybe name it  `EmptyMemberPath`)
* In `SymbolAccessor.TryFindMemberPath` if the path is empty return a new instance of the `RootObjectMemberPath` instead of the usual `MemberPath`. Pass the containing `type` to it.
* Set the `MemberPath.MemberType` property to non-nullable and virtual
* In `RootObjectMemberPath` overwrite the `MemberType` to return the type of the root which was passed in via constructor.
* `MemberPath` would simply throw if `Member` or `MemberType` is accessed on an empty `MemberPath` unless it is a `RootObjectMemberPath` (which would be expected).

WDYT?

src/Riok.Mapperly/Symbols/NonEmptyMemberPath.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    /// <summary>
+    /// Gets the last part of the path or throw if there is none.
+    /// </summary>
+    public new IMappableMember Member => Path[^1];
+
+    /// <summary>
+    /// Gets the type of the <see cref="Member"/>. If any part of the path is nullable, this type will be nullable too.
+    /// </summary>
+    public new ITypeSymbol MemberType => IsAnyNullable() ? Member.Type.WithNullableAnnotation(NullableAnnotation.Annotated) : Member.Type;
```
latonz, line 22: Couldn't this be as simply as `=> base.MemberType!;` as we are sure there is always a member type in the base since there is always a path?

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberExistingTargetMapping.cs
```diff
@@ -7,17 +7,21 @@ namespace Riok.Mapperly.Descriptors.Mappings.MemberMappings;
 /// <summary>
 /// A <see cref="IMemberAssignmentMapping"/> which maps to an existing target instance.
 /// </summary>
-public class MemberExistingTargetMapping(IExistingTargetMapping delegateMapping, GetterMemberPath sourcePath, GetterMemberPath targetPath)
-    : IMemberAssignmentMapping
+public class MemberExistingTargetMapping(
+    IExistingTargetMapping delegateMapping,
+    GetterMemberPath sourcePath,
+    GetterMemberPath targetGetterPath,
+    NonEmptyMemberPath targetMemberPath
```
latonz, line 14: With the current naming and without updated comments it is really hard to tell whats the difference between those two paths. Maybe explain with a comment.

latonz (changes requested):
Thanks for this contribution! I really like the idea of `MemberPath.ToDisplayString`.
And great test coverage 👍 
I added my feedback 😊

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFromSourceTest.cs
```diff
@@ -0,0 +1,120 @@
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class ObjectPropertyFromSourceTest
+{
+    [Fact]
+    public void WithManualMappedProperty()
```
latonz, line 8: Can you also add a test case for the following scenarios?
* to map to a constructor parameter
* to map to a init only property
* to map to a tuple value
* in a queryable projection mapping (probably in another test class)
* with a nullable object (where the source is nullable)

latonz (changes requested):

src/Riok.Mapperly/Symbols/NonEmptyMemberPath.cs
latonz: I like your updated approach 😊 The only point that looks kinda weird to me is the new naming of the `GetterMemberPath` and `SetterMemberPath`. For example `*Mapping` classes storing and exposing `*Builder` properties looks strange to me (e.g. `NullMemberMapping`).
Overall I think I'd prefer keeping the old names `GetterMemberPath` and `SetterMemberPath`. This should also drastically reduce the size of this PR.

latonz: `MemberPathGetter/Setter` indicates that the `MemberPathSetter` would provide functionality to `set` a `MemberPath` which isn't the case either. For me the best fit (although not perfect) is still `GetterMemberPath` and `SetterMemberPath`.

latonz (approved):

latonz:
Thank you for this great addition to Mapperly!

PR #1384, opened 2024-07-08
Add AggressiveInlining support

### LINKED ISSUE #1381: Using [MethodImpl(MethodImplOptions.AggressiveInlining)] for generated mapping methods
According to this [benchmark](https://github.com/mjebrahimi/Performance-Wars-Benchmarking-CSharp/tree/master/AggressiveInlining-Benchmark), it shows using `[MethodImpl(MethodImplOptions.AggressiveInlining)]` for mapping `struct`s is much faster than not using it.

I must confess that much performance difference was surprising to me but in fact, it is! and why don't we advantage of it? :) 

The generated mapping methods are simple and seem safe to use with `AggressiveInlining`, so I think there's nothing wrong with it (at least it's worth a try).

Note: mapping methods in this benchmark are mapperly-generated and I just edited them to decorate with `AggressiveInlining`.

Benchmark

mjebrahimi:
# Add AggressiveInlining support

## Description

- Adds support for generating `[MethodImpl(MethodImplOptions.AggressiveInlining)]` for mapping methods.
- Adds expected code to verify snapshots to pass the tests.

Fixes #1381 (issue)

I did some [benchmarks](https://github.com/mjebrahimi/mapperly-benchmark) and the results were amazing.

- **27x** faster for simple `struct` types
- **1.3x** faster for complex `struct` types
- Very slightly faster for `class` types

image

## Checklist


latonz:
I added a few points to be discussed here: https://github.com/riok/mapperly/issues/1381#issuecomment-2218352641

mjebrahimi:
I updated the code according to our discussion in #1381
**Current status: Work in progress (need some help to complete the code)**

- I created that enum and named it `AggressiveInliningMode` as I think it's better naming.
- I didn't set number value for enums because it's not going to be `[Flag]`able
- I added AggressiveInliningMode to `MapperAttribute` with defaults to `ValueTypes` (**what you do think about the default value?**)
- I added support for this new property wherever I found necessary (such as `MapperConfiguration`, `MapperConfigurationMerger`, `TestSourceBuilder`, `TestSourceBuilderOptions`) (**did I miss anything?**)
- I added `GenerateMethodImplAttributeListIfNeeded` method to `MethodMapping` class. (**is everything ok with it and it's location?**)
- I needed to read the value of `MapperAttribute.AggressiveInliningMode` property but I couldn't find how to (**need some help for this**)
- As we must skip applying AgressiveInlining if the method has already [MethodImpl] attribute, I wrote a method `HasAttribute<TAttribute>` in 
`Riok.Mapperly/Helpers/MethodDeclarationSyntaxExtensions.cs` (**however I have some considerations**)

Developers can use [MethodImpl] attribute in several ways (including and not limited to bellows)
```csharp
[MethodImpl]
[MethodImplAttribute]
[System.Runtime.CompilerServices.MethodImpl]
[System.Runtime.CompilerServices.MethodImplAttribute]
[global::System.Runtime.CompilerServices.MethodImpl]
[global::System.Runtime.CompilerServices.MethodImplAttribute]
```
We could check the string of attribute syntax with any of those possibilities.
For example:
```csharp
private static readonly string[] _methodImplPossiblities =
[
    "MethodImpl",
    "MethodImplAttribute",
    "System.Runtime.CompilerServices.MethodImpl",
    "System.Runtime.CompilerServices.MethodImplAttribute",
    "global::System.Runtime.CompilerServices.MethodImpl",
    "global::System.Runtime.CompilerServices.MethodImplAttribute",
];
private static bool HasMethodImpleAttribute(MethodDeclarationSyntax methodDeclarationSyntax)
{
    var attributes = methodDeclarationSyntax.AttributeLists.SelectMany(p => p.Attributes);
    return attributes.Any(attributeSyntax => Array.Exists(_methodImplPossiblities, p => attributeSyntax.ToString().Contains(p, StringComparison.Ordinal)));
}
```
But it seems it's not a clean and robust approach
Even worse when developers use type aliasing! in this way we can't recognize it at all, for example:
```csharp
using AliasName = System.Runtime.CompilerServices.MethodImplAttribute;

[Mapper]
public static partial class MyMapper
{
    [AliasName(MethodImplOptions.AggressiveInlining)]
    public static partial DateTime MapTo(DateTime dt);
}
```
So after some research, I found this way that can handle all of those. (this is the current implementation)
```csharp
public static bool HasAttribute<TAttribute>(this MethodDeclarationSyntax methodDeclarationSyntax) where TAttribute : Attribute
{
    var csharpCompilation = CSharpCompilation.Create(
        assemblyName: null,
        syntaxTrees: [methodDeclarationSyntax!.SyntaxTree],
        references: [MetadataReference.CreateFromFile(typeof(TAttribute).Assembly.Location)]
    );

    var semanticModel = csharpCompilation.GetSemanticModel(methodDeclarationSyntax.SyntaxTree);

    var attributes = methodDeclarationSyntax.AttributeLists.SelectMany(p => p.Attributes);

    return attributes.Any(attributeSyntax =>
    {
        var symbols = semanticModel.GetSymbolInfo(attributeSyntax).CandidateSymbols.OfType<IMethodSymbol>();
        return symbols.Any(p => string.Equals(p.ContainingType.ToString(), typeof(TAttribute).FullName, StringComparison.Ordinal));
    });
}
```
This method works fine, however, it always Compiles the code to understand the meaning, and I'm not sure if it's ok or not due to performance problems (**what do you think?**)

Next Step:
After completing the implementation I go for writing some unit/integration tests to ensure its functionality in different conditions. (**Anything else?**)

latonz:
Thank you for the updates 😊 💪 

* I'm not sure if `AggressiveInliningMode` is a better name, this limits future changes without needing breaking changes 🤔  `Types` is much more specific IMO.
* I would prefer a flags enum as this allows more flexibility for the future (for example we may want an extra enum entry etc.). Also this aligns with similar public Mapperly APIs (e.g. `RequiredMappingStrategy`).
* I'm not sure whether we should check only the source type of the mapping against the configuration, what are your thoughts? The more I think about it, the more I think a simple boolean enable/disable configuration might be easier 🤔🤔 
* Default value: why do you think we should default to value types only instead of all? Because of the assembly size / compilation time vs. performance gain?
* Instead of the syntax based evaluation whether the `MethodImpl` attribute is present I think it would be much easier to do it on the semantic model. Idea:
  * Add a boolean member on the `MethodMapping` on whether to the `MethodImpl` attribute needs to be added
  * Whenever instantiating a `MethodMapping` with an `IMethodSymbol` set this member to true if the attribute is absent on the user declared method symbol and the configuration is enabled for the given types (e.g. in various methods of `UserMethodMappingExtractor`, e.g. `BuildUserDefinedMapping`)
  * Evaluation whether the attribute is present or not should be as simple as `ctx.SymbolAccessor.HasAttribute<MethodImplAttribute>(methodSymbol)`
  * Access to the configuration in `UserMethodMappingExtractor` is available through `ctx.Configuration.Mapper.AggressiveInliningMode` (you cannot read the configuration in the emit / build phase, it is only available when building the mappings)

What else should be done?
* Update the documentation
  * Update the benchmarks
  * Add a section about this feature
* Add a unit test (ensures different scenarios work)
* Add an integration test (ensures the basic feature works on different .NET versions)

mjebrahimi:
Thank you so much for your help.

I just wanted to sync my branch with the upstream and used the re-sync (with discard commits) feature of GitHub, but didn't know it closes the pull request! 😅

I'm working on it and I'm gonna re-open this PR with new updates by today ✌️

mjebrahimi:
Thanks @latonz your idea was very helpful to me in understanding the codebase better and finding the right path to implement this feature.

I came with nice updates. 😊

**About the previous conversation:**

- Default Value: My only concern about defaulting to `ValueTypes` is that:
  - Given that the performance improvement of `AggressiveInlining` is mostly for value types, rather than reference types.
  - Developers mostly use class types for their DTOs/Entities and map them.
  - Given the previous, using AggressiveInlining results in increasing the assembly size and compilation time without almost any gain.
  - For more caution, there may be other disadvantages to using AggressiveInlining too much, that we are not yet aware of.

- Configuration: The more I think about having an enum, the more it gets complex unnecessarily.
  - We can have a class that contains a struct and so on. Should we decide based on the root mapping method or per each mapping method?
  - We can have a method that maps a struct/value type to a class/ref type. Should we decide based on input type or output type? 
  - Is there any need to add more types to that enum in the future? For example what? Is it necessary?

- **In the end:** I think we don't need that much complexity (over-engineering) and I do agree with you with a `bool` flag to enable/disable it with defaults to `true` (the current implementation).

**About the current implementation:**

- According to your idea I added a `bool` argument named `enableAggressiveInlinig` for MappingMethod types.
- For the below types in the `UserMethodMappingExtractor` class, the value for `enableAggressiveInlinig` is read from `Configuration and !HasAttribute<MethodImple>()`.
  - UserDefinedNewInstanceGenericTypeMapping
  - UserDefinedExistingTargetMethodMapping
  - UserDefinedNewInstanceMethodMapping
  - UserDefinedNewInstanceRuntimeTargetTypeParameterMapping
- For these types, the value for `enableAggressiveInlinig` is read only from `Configuration`.
- Based on my understanding these mapping methods are generated as _non-partial_ and _are not the root/user_ mapping method, so we don't need to check if the root/user method has `[MethodImple]` attribute, we can generate AggressiveInlining if it's enabled in the Configuration even when the root/user mapping method has `[MethodImple]` attribute. (**please correct me if I'm wrong**)
  - ArrayForEachMapping
  - ArrayForMapping
  - EnumNameMapping
  - ForEachAddEnumerableMapping
  - ForEachSetDictionaryMapping
  - MethodMapping
  - NewInstanceMethodMapping
  - NewInstanceObjectFactoryMemberMapping
  - NewInstanceObjectMemberMethodMapping
  - NewValueTupleExpressionMapping
  - NullDelegateMethodMapping
  - ObjectMemberMethodMapping
  - UserDefinedExistingTargetMethodMapping
  - UserDefinedNewInstanceGenericTypeMapping
  - UserDefinedNewInstanceMethodMapping
  - UserDefinedNewInstanceRuntimeTargetTypeMapping
  - UserDefinedNewInstanceRuntimeTargetTypeParameterMapping
- I found these types don't need the `enableAggressiveInlinig` parameter, so passed the `false` to them. (although all tests were passed, I'm not sure it's correct or if our tests are not complete enough to catch if there is any problem with it. **What do you think?**)
  - EnumFromStringSwitchMapping
  - EnumToStringMapping
  - ForEachAddEnumerableObjectFactoryMapping
  - ForEachSetDictionaryObjectFactoryMapping
  - QueryableProjectionMapping
  - NoOpMapping
  - UnimplementedMapping

**Support for `UnsafeAccessor` methods:**

- According to this benchmark `AggressiveInlining` works fine with `UnsafeAccessor` methods and is beneficial.
- Benchmark-UnsafeAccessor
- I added a `bool` argument named `enableAggressiveInlinig` to the below types and changed `DescriptorBuilder` class a bit to read the `Configuration` and pass the value to that parameter.
  - UnsafeAccessorContext
  - UnsafeConstructorAccessor
  - UnsafeFieldAccessor
  - UnsafeGetPropertyAccessor
  - UnsafeSetPropertyAccessor

**Testing:**

I ensured all the previous tests were passed correctly and added some new tests to ensure it works fine in different scenarios.

**Final benchmark (the outcome🤩):**

I added a new benchmark and the result eliminated all my fatigue :)

Benchmark

**Documentation:**

The `.mdx` files and the tools that is used for docs were unfamiliar to me, I'm trying to learn it and need some help. 
**How can I build/run the docs and see the preview of the output?**

**CICD Pipeline:**

I'm trying to run/pass most of the workflows locally or on my fork using GitHub Actions.
**Could you approve and run the pipeline manually to check out the output?**

**The checklist:**

- [x] Implement feature
- [x] Pass the previous tests (since the feature is enabled by default, all the unit/integration tests use it and were passed)
- [x] Add more tests for different scenarios
- [x] Add benchmarks
- [ ] Complete the documents (in progress)
- [ ] Pass the CICD tests locally (in progress)

latonz:
I'll close this for now... If you don't have the time/motivation to update again, I'll do it as soon as I find the time.

mjebrahimi:
Unfortunately, I have been too busy and didn't have enough time to complete it.
Hope for future contributions :)

PR #1419, opened 2024-07-26
Support inline expression with casting

trejjam:
# Support inline expression with casting

## Description

When using `Use = nameof()`, which contains casting, e.g. `((MyEnum) value & MyEnum.OptionA) > 0` generated mapping does not resolve the fully qualified type of the casted type.

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/InlineExpressionRewriter.cs
+++ b/src/Riok.Mapperly/Descriptors/InlineExpressionRewriter.cs
@@ -70,7 +70,11 @@ public class InlineExpressionRewriter(SemanticModel semanticModel, Func<IMethodS
     {
         var result = base.VisitCastExpression(node);
 
-        if (result is CastExpressionSyntax typedResult && semanticModel.GetSymbolInfo(node.Type).Symbol is ITypeSymbol namedTypeSymbol)
+        if (
+            CanBeInlined
+            && result is CastExpressionSyntax typedResult
+            && semanticModel.GetSymbolInfo(node.Type).Symbol is ITypeSymbol namedTypeSymbol
+        )
         {
             var fullyQualifiedType = FullyQualifiedIdentifier(namedTypeSymbol);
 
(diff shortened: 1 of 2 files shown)


src/Riok.Mapperly/Descriptors/InlineExpressionRewriter.cs
```diff
@@ -66,6 +66,38 @@ public override SyntaxNode VisitTypeArgumentList(TypeArgumentListSyntax node)
...
+        return result;
+    }
+
+    public override SyntaxNode? VisitBinaryExpression(BinaryExpressionSyntax node)
+    {
+        var result = base.VisitBinaryExpression(node);
+
+        if (
+            result is BinaryExpressionSyntax typedResult
+            && typedResult.Kind() is SyntaxKind.AsExpression
```
latonz, line 89: Nitpick: `typedResult.Kind() is SyntaxKind.AsExpression` could be simplified to `typedResult.IsKind(SyntaxKind.AsExpression)`

test/Riok.Mapperly.Tests/Helpers/InlineExpressionRewriterTest.cs
```diff
@@ -20,7 +20,7 @@ public class InlineExpressionRewriterTest
     [InlineData("new TestRecord[2].ToString()", true, "new global::TestRecord[2].ToString()")]
     [InlineData("base.ToString()", false)] // CS0831
     [InlineData("(1,2).ToString()", false)] // CS8143
-    [InlineData("((string?)source)?.ToString()!", false)] // CS8072
+    [InlineData("((string)source)?.ToString()!", false)] // CS8072
```
latonz, line 23: Could you explain this change?

latonz (changes requested):
Thank you very much for this contribution, I added my feedback 😊

test/Riok.Mapperly.Tests/Helpers/InlineExpressionRewriterTest.cs
trejjam: Sure,

Inline expression rewriter is used in projections, which have disabled nullable context. When a nullable type is present inside this block, the compiler is not happy (you can not use nullable annotations inside a non-nullable context).

When casts are processed in the visitor, the nullability is stripped out.

latonz: Got it, that makes sense. But since this test has `canBeInlined=false`, there is no need to change it, is there?

trejjam: It didn't come to my mind. Now visitors are applied only when `CanBeInlined=true`.

latonz (approved):

latonz:

trejjam:
Thank you for processing it. Is it possible to publish a new release? A pre-release is fine for me.

PR #1464, opened 2024-08-26
Adding support for mappers being nested under interfaces or records

david-driscoll:
# Allow mappers to be nested in records or interfaces

## Description
This is a minor update to allow mappers to be nested under an `interface` or `record` in addition to `class`.

## Checklist


src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.SymbolDeclaration.cs
```diff
@@ -28,4 +28,27 @@ public ClassDeclarationSyntax Class(string name, SyntaxTokenList modifiers, Synt
...
+    public TypeDeclarationSyntax TypeDeclaration(TypeDeclarationSyntax syntax, SyntaxList<MemberDeclarationSyntax> members)
+    {
+        var name = Identifier(syntax.Identifier.ValueText);
+        TypeDeclarationSyntax type = syntax switch
+        {
+            ClassDeclarationSyntax => ClassDeclaration(name),
+            StructDeclarationSyntax => StructDeclaration(name),
+            InterfaceDeclarationSyntax => InterfaceDeclaration(name),
+            RecordDeclarationSyntax => RecordDeclaration(Token(SyntaxKind.RecordKeyword), name),
+            _ => throw new NotSupportedException("Unsupported type declaration syntax.")
```
latonz, line 41: To simplify troubleshooting it may be helpful to include the type of the `syntax` in the error message.

latonz (changes requested):
Thank you for this great contribution!

The failing test should be fixable by simply adding a pseudo field like `private readonly string _test = test;`

latonz:
As the integration tests run on all supported .NET versions the primary constructor test needs to be excluded from .NET versions < 8.0.
But IMO these tests would fit better as unit tests instead of integration tests since the generated code is not really .NET version dependent and there is already one integration test for this scenario (`NestedTestMapper`).

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.SymbolDeclaration.cs
david-driscoll: Makes sense.

david-driscoll:
I added the primary constructor because I've run into issues with it in the past on my own source generators, so I just thought I'd add it for completeness.

I'm working on the changes now.

latonz (approved):

latonz:
LGTM, thanks for this contribution!

PR #1498, opened 2024-09-19
Support inlining of method with generic type arguments

trejjam:
# Support inlining of method with generic type arguments

## Description

Inlining of
```cs
public class Test
{
    public IReadOnlyCollection<A> MapExpression(A value) => Array.Empty<A>();
}
```

Produces:
`global::System.Array.Empty()`

instead of
`global::System.Array.Empty<global::A>()`

## Checklist


latonz (approved):
Thank you for this bug fix!

PR #1592, opened 2024-11-13
Diagnostic for mapping null mismatch

### LINKED ISSUE #602: Diagnostic for mapping null mismatch
Emit a diagnostic if the source is nullable but the target isn't. Default severity should be info.
Proposed message: `Mapping the nullable source property {0} of {1} to the target property {2} of {3} which is not nullable`

Relates https://github.com/riok/mapperly/discussions/375

jorisBarkema:
# Diagnostic for mapping null mismatch

## Description
Adds a diagnostic when a nullable source value is mapped to a non-nullable target value
Related issue: https://github.com/riok/mapperly/issues/602

Fixes #602 

## Checklist

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -8,9 +8,9 @@
       "name": "docs",
       "version": "0.0.0",
       "dependencies": {
-        "@docusaurus/core": "3.6.1",
-        "@docusaurus/plugin-ideal-image": "3.6.1",
-        "@docusaurus/preset-classic": "3.6.1",
+        "@docusaurus/core": "3.6.3",
+        "@docusaurus/plugin-ideal-image": "3.6.3",
+        "@docusaurus/preset-classic": "3.6.3",
         "@easyops-cn/docusaurus-search-local": "0.45.0",
         "@mdx-js/react": "3.0.1",
         "clsx": "2.1.0",
@@ -22,9 +22,9 @@
         "react-dom": "18.2.0"
       },
       "devDependencies": {
-        "@docusaurus/eslint-plugin": "3.6.1",
-        "@docusaurus/module-type-aliases": "3.6.1",
-        "@docusaurus/tsconfig": "3.6.1",
+        "@docusaurus/eslint-plugin": "3.6.3",
+        "@docusaurus/module-type-aliases": "3.6.3",
+        "@docusaurus/tsconfig": "3.6.3",
         "@types/react": "18.2.79",
         "@typescript-eslint/eslint-plugin": "7.16.1",
         "hast-util-to-html": "9.0.2",
@@ -113,14 +113,14 @@
       }
     },
     "node_modules/@algolia/client-abtesting": {
-      "version": "5.14.2",
-      "resolved": "https://registry.npmjs.org/@algolia/client-abtesting/-/client-abtesting-5.14.2.tgz",
-      "integrity": "sha512-7fq1tWIy1aNJEaNHxWy3EwDkuo4k22+NBnxq9QlYVSLLXtr6HqmAm6bQgNNzGT3vm21iKqWO9efk+HIhEM1SzQ==",
+      "version": "5.15.0",
+      "resolved": "https://registry.npmjs.org/@algolia/client-abtesting/-/client-abtesting-5.15.0.tgz",
+      "integrity": "sha512-FaEM40iuiv1mAipYyiptP4EyxkJ8qHfowCpEeusdHUC4C7spATJYArD2rX3AxkVeREkDIgYEOuXcwKUbDCr7Nw==",
       "dependencies": {
-        "@algolia/client-common": "5.14.2",
-        "@algolia/requester-browser-xhr": "5.14.2",
-        "@algolia/requester-fetch": "5.14.2",
-        "@algolia/requester-node-http": "5.14.2"
+        "@algolia/client-common": "5.15.0",
+        "@algolia/requester-browser-xhr": "5.15.0",
+        "@algolia/requester-fetch": "5.15.0",
+        "@algolia/requester-node-http": "5.15.0"
       },
       "engines": {
         "node": ">= 14.0.0"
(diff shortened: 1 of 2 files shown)


src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
```diff
@@ -84,6 +85,17 @@ public static bool TryBuild(
             return false;
         }
 
+        if (memberMappingInfo.IsSourceNullable && !targetMember.MemberType.IsNullable())
```
latonz, line 88: Couldn't you use `delegateMapping.TargetType.IsNullable()` instead of `!targetMember.MemberType.IsNullable()` which should solve the `ShouldPassNullValueToNullableUserMappingMethod` test.

latonz (changes requested):
I think in general all call sites of `FindOrBuildLooseNullableMapping` need to report this diagnostic (that is what you are already doing as there is only one). I think additionally we should report another diagnostic in `Riok.Mapperly.Descriptors.MappingBuilders.NullableMappingBuilder`.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
jorisBarkema: I hope you can forgive me if I am misunderstanding you, I changed it to `!delegateMapping.TargetType.IsNullable()` but that test is still failing, and I added the diagnostic for the NullableMappingBuilder

latonz: Oh my bad I mixed it up a bit. I cannot try it myself right now, but I think you would need to check source and target... `memberMappingInfo` contains the source/and target member being mapped. `delegateMapping` is the mapping used to map from the source to the target (convert the types). This means nullability could mismatch from source to the source of the delegateMapping and from the target of the delegateMapping to the target member.
So I think the correct condition would be `(!memberMappingInfo.TargetMember.MemberType.IsNullable() && delegateMapping.TargetType.IsNullable())
            || (!delegateMapping.SourceType.IsNullable() && memberMappingInfo.TargetMember.MemberType.IsNullable())`.

jorisBarkema: Everything seems to work as expected with the exception of this test: 
DisabledNullableClassPropertyToNonNullableProperty now gives a diagnostic, don't know if that's desired
And I would like your opinion on the setting to disable the warning. This was requested in the original ticket, but it can be achieved with pragma disable as well ofcourse, it's not a setting like the other settings which actually change how the mappings work. So:
a. Is this something that should be added?
b. If so, make it one setting for both these diagnostics together or separate settings? I would lean to one setting personally

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
```diff
@@ -84,6 +85,22 @@ public static bool TryBuild(
             return false;
         }
 
+        var memberTargetNullable = memberMappingInfo.TargetMember.MemberType.IsNullable();
+        var delegateTargetNullable = delegateMapping.TargetType.IsNullable();
+        var memberSourceNullable = memberMappingInfo.IsSourceNullable;
+        var delegateSourceNullable = delegateMapping.SourceType.IsNullable();
+
+        if (memberSourceNullable && !memberTargetNullable && !(delegateSourceNullable && !delegateTargetNullable))
```
latonz, line 93: Shouldn't the first two and the second two have an OR relationship instead of AND?

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -827,6 +827,26 @@ public static class DiagnosticDescriptors
             true
         );
 
+    public static readonly DiagnosticDescriptor NullableSourceValueToNonNullableTargetValue =
+        new(
+            "RMG089",
+            "Mapping nullable source to non-nullable target member",
+            "Mapping the nullable source property {0} of {1} to the target property {2} of {3} which is not nullable",
+            DiagnosticCategories.Mapper,
+            DiagnosticSeverity.Warning,
```
latonz, line 836: IMO the default severity for these two new diagnostics should be info.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
latonz: Regarding `DisabledNullableClassPropertyToNonNullableProperty`: Unfortunately, at the beginning of Mapperly, we decided to treat nullable-disabled contexts as if they were nullable-enabled contexts with all possible nullable annotations set. In retrospect, this was a bad decision, and we want to change it in the future, but it will take a lot of work. There are scenarios where nullable-disabled needs to be treated as non-nullable (for example here, in queryable projections, ...)...

Regarding the setting: The severity of these diagnostics can be set quite flexibly via editorconfig files (on a per-file or per-directory basis). However, I think enabling or disabling it for a given member is not possible with editorconfig/pragma/SuppressMessageAttribute, so a config is nice. I'd implement this in a separate PR, as this PR already adds a useful feature. My API suggestion for this would be to add a new property `bool SuppressNullMismatchDiagnostic` to the `MapPropertyAttribute`.

latonz (changes requested):
Thank you for the updates, I added my feedback 😊

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
jorisBarkema: The intention here is that the last case `delegateSourceNullable && !delegateTargetNullable` happens when the delegate function is manually defined and already has null handling, in which case we don't want to give the diagnostic. For example, test `ShouldPassNullValueToNullableUserMappingMethod` uses 

```
private string MapString(string? source)
    => source ?? "(null)";
```

Which would still give a diagnostic without that last check

latonz: But `ShouldPassNullValueToNullableUserMappingMethod ` should diagnostic for the member Value2, but not for the member Value.

latonz (changes requested):

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
jorisBarkema: It does right now, looking at the ShouldPassNullValueToNullableUserMappingMethod verified file

latonz (approved):

latonz:
I made a mistake while rebasing this PR, which resulted in an inability to push to the original branch. As a result, the PR was automatically closed.

To resolve this, I’ve opened a new PR (#1612) with the same changes. Please note that while the PR is new, the commit author remains the original one, so the history should still reflect the original work.

Apologies for the inconvenience, and thank you for your understanding!

jorisBarkema:
Alright no problem, thanks for the reviews and for maintaining this project!

Issue #1614, opened 2024-11-27
Improve null handling


latonz:
Currently, Mapperly upgrades non-nullable reference types to annotated nullable reference types early in the mapping pipeline. Additionally, when nullables are involved, Mapperly incorporates null handling into the mapping process and continues the pipeline with non-nullable types. This approach centralizes null handling, making it simple and consistent.

However, this design has led to several bugs, indicating the need for a more robust solution. We need to define a clear concept to address these issues and improve the handling of nullable reference types without compromising simplicity or maintainability.

Related Bugs:
* https://github.com/riok/mapperly/issues/1611
* https://github.com/riok/mapperly/issues/1293

See also https://github.com/riok/mapperly/issues/1117#issuecomment-2503737493.

TonEnfer:
I think that when doing this task, you should also take into account similar conversions from #1117:

| Source       | Is nullable | Target                                    | Is nullable | Expected result                                                                                                                            |
| ------------ | ----------- | ----------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
| `class A {}` | true        | `class B { static B Create(A source) }`   | false       | `target = global::B.Create(source ?? throw new ArgumentNullException(nameof(source)));`                                                    |
| `class A {}` | false       | `class B { static B? Create(A source) }`  | false       | `target = global::B.Create(source) ?? throw new ArgumentNullException();`                                                                  |
| `class A {}` | false       | `struct B { static B? Create(A source) }` | false       | `target = global::B.Create(source) ?? throw new ArgumentNullException();` or <p/> `target = global::B.Create(source).GetValueOrDefault();` |

Also pay attention to the results of these two tests:
https://github.com/riok/mapperly/blob/39376faa613efad21e32921cbad884aa4c1c3444/test/Riok.Mapperly.Tests/Mapping/StaticMethodConversionTest.cs#L135-L146

CoreDX9:
Can null value assignments be configured separately for each property?
Like this:
``` csharp
[AllowNullPropertyAssignment(false, Include = ["ArrayProp1", "ListProp2"])]
public static partial void UpdateEntity(this UpdateDto dto, MyEntity entity);
```
This way, null values can be used in the collection navigation properties to indicate that the collection navigation properties are not modified when update entity using EF Core.

latonz:
This is not yet possible.

johnnyelwailer:
Greetings. I have a related use case.
Consider this code:
```csharp
[...]

PR #1616, opened 2024-11-30
Add support for `From` and `To` method conversions (#1117)

### LINKED ISSUE #1117: Support `From` and `To` static factory methods
Many objects provide static methods instead of constructors for semantic reasons. Unfortunately as Mapperly is unable to use them users have to manually provide wrapper methods. Based [upon](https://github.com/riok/mapperly/issues/1107#issuecomment-1935306780).

#### Solution
Mapperly should look for static methods that meet the following criteria:
- static methods on the target type named From{SourceType.Name} accepting the source type as a single parameter and the target type as return type
- static methods on the source type named To{TargetType.Name} accepting the source type as single parameter returning the target type

#### Implementation
- This should be used as a fallback for if there are no compatible constructors. Should look for the target types `From` methods before looking for the source `To` methods.
- Perhaps the sources `To` methods could include non static methods.

```csharp
var target = new global::Mapperly_gRPC.HelloReply();
target.Message = response.Message;
target.ResponseTime = Timestamp.FromDateTimeOffset(response.ResponseTime);
return target;
```

See #1107.

TonEnfer:
# Add support for `From` and `To` method conversions (#1117)

## Description

* Added support for conversion by calling the ToTTarget method on the source object
* Added support for conversion by calling the static Create, CreateFrom, FromTSource methods on the integer type, as well as by calling the static ToTTarget method on the source type

Despite the discussion in the issue, the ToString mapping builder remains untouched as it contains a lot of logic.
Instead, the new mapping builder eliminates the mapping to `string`.

The previously existing mapping builders for `DateTime` mapping have been removed, as their role is now performed by the new mapping builder for static methods.
However, `MappingConversionType` has only been extended, the previously existing `MappingConversionType.DateTimeToDateOnly` and `MappingConversionType.DateTimeToTimeOnly` are left for backward compatibility, their processing has been moved to the new mapping builder

Fixes #1117

## Checklist


src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+            // otherwise try to find method ignoring the nullability
+            method = candidates.FirstOrDefault(x => SymbolEqualityComparer.Default.Equals(x.ReturnType, nonNullableTargetType));
+
+            if (method == null)
+            {
+                continue;
+            }
+
+            mapping = new StaticMethodMapping(method);
+            return true;
```
latonz, line 129: This doesn't seem to be covered by tests, could you add tests for this?

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertInstanceMethodMappingBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+        if (method is not null)
+            return new SourceObjectMethodMapping(ctx.Source, ctx.Target, method.Name);
+
+        if (!targetIsNullable)
+            return null;
+
+        // otherwise try to find method ignoring the nullability
+        method = methodCandidates.FirstOrDefault(x => SymbolEqualityComparer.Default.Equals(x.ReturnType, nonNullableTarget));
+
+        return method is null ? null : new SourceObjectMethodMapping(ctx.Source, ctx.Target, method.Name);
```
latonz, line 39: This doesn't seem to be covered by tests, could you add some tests for this?

test/Riok.Mapperly.Tests/Mapping/DateTimeTest.cs
```diff
@@ -1,18 +0,0 @@
-namespace Riok.Mapperly.Tests.Mapping;
-
-public class DateTimeTest
-{
-    [Fact]
-    public void DateTimeToDateOnly()
```
latonz, line 6: I think it makes sense to keep this test to ensure backward compatibility, just move it to the new test class.

test/Riok.Mapperly.Tests/Mapping/StaticMethodConversionTest.cs
```diff
@@ -0,0 +1,373 @@
+using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class StaticMethodConversionTest
+{
+    #region CreateMethod
```
latonz, line 8: Instead of regions, I'd prefer separate test classes...

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+        return false;
+    }
+
+    private static IMethodSymbol[] GetMethodCandidates(ICollection<IMethodSymbol> allMethods, string methodName, ITypeSymbol parameterType)
+    {
+        return allMethods
+            .Where(m =>
+                string.Equals(m.Name, methodName, StringComparison.Ordinal)
+                && m is { IsStatic: true, ReturnsVoid: false, IsAsync: false, Parameters.Length: 1 }
+                && FilterParameterType(m.Parameters[0], parameterType)
```
latonz, line 142: I think it could make sense to create a new method symbol extension method `bool ValidateSignature(this IMethodSymbol method, ITypeSymbol returnType, IReadOnlyCollection<ITypeSymbol> argTypes)` which checks:
* does it return a type which is assignable to the return type (probably use `SymbolAccessor.CanAssign`)
* does the parameter count match and is each argument assignable to the parameter

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+            )
+            .ToArray();
+    }
+
+    private static bool FilterParameterType(IParameterSymbol parameter, ITypeSymbol targetType)
+    {
+        if (SymbolEqualityComparer.Default.Equals(parameter.Type, targetType))
```
latonz, line 149: Shouldn't this include nullability?

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+            x.OriginalDefinition.SpecialType == SpecialType.System_Collections_Generic_IEnumerable_T
+        );
+
+        if (parameter.Type.IsArrayType(out var arrayTypeSymbol))
+        {
+            return SymbolEqualityComparer.Default.Equals(
+                arrayTypeSymbol.ElementType,
+                targetIsArray ? targetArrayTypeSymbol!.ElementType : targetType
+            );
+        }
```
latonz, line 167: This early return could happen before the enumerable check of the target, couldn't it?

latonz (changes requested):
Thank you for the PR! I’m excited to see this feature added to Mapperly. I’ve shared my feedback 😊

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+            : null;
+    }
+
+    private static bool IsDateTimeToDateOnlyConversion(MappingBuilderContext ctx)
+    {
+        return ctx.Source.SpecialType == SpecialType.System_DateTime
+                && ctx.Types.DateOnly != null
+                && ctx.Target is INamedTypeSymbol namedSymbol
+                && SymbolEqualityComparer.Default.Equals(namedSymbol, ctx.Types.DateOnly)
+            || true;
```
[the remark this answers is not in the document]
TonEnfer: Oops. I have no idea where this came from 🤦

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertInstanceMethodMappingBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+        // ignore `ToString` mapping for backward compatibility
+        if (nonNullableTarget.SpecialType == SpecialType.System_String)
+            return null;
+
+        var methodName = GetMappingMethodName(ctx);
+
+        var methodCandidates = ctx
+            .SymbolAccessor.GetAllMethods(ctx.Source, methodName)
+            .Where(x => x is { IsStatic: false, ReturnsVoid: false, IsAsync: false, Parameters.Length: 0 })
+            .ToArray();
```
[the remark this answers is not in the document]
TonEnfer: The only reason I chose `ToArray` is because the list is not mutable and should not be mutable further down the code.
I'll change it to `ToList` since it works better.

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+        const string create = "Create";
+        const string from = "From";
+
+        yield return create;
+        yield return $"{create}{from}";
+
+        if (!ctx.Source.IsArrayType(out var arrayType))
+            yield return $"{from}{ctx.Source.Name}";
+        else
+            yield return $"{from}{arrayType.ElementType.Name}Array";
```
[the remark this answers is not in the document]
TonEnfer: I think you are right, I will add this

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertInstanceMethodMappingBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+        return method is null ? null : new SourceObjectMethodMapping(ctx.Source, ctx.Target, method.Name);
+    }
+
+    private static string GetMappingMethodName(MappingBuilderContext ctx)
+    {
+        var nonNullableTarget = ctx.Target.NonNullable();
+
+        if (!nonNullableTarget.IsArrayType(out var arrayType))
+        {
+            return $"To{nonNullableTarget.Name}";
```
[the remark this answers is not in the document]
TonEnfer: I guess that would make sense, but I don't know how I would get language keyword names. Could you give me a hint?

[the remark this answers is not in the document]
latonz: I don't think there is an easy way to do this, you may need to handle it with a dictionary or a switch...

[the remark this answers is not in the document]
TonEnfer: It looks like this code produces almost what is needed.
```csharp
if (ctx.Source.SpecialType is >= SpecialType.System_Boolean and < SpecialType.System_String)
{
    var simpleName = CultureInfo.InvariantCulture.TextInfo.ToTitleCase(ctx.Source.ToDisplayString());
}
```
Example:
```csharp
var allNames =
    Enumerable.Range((int)SpecialType.System_Boolean, SpecialType.System_String - SpecialType.System_Boolean)
    .Select(x => CultureInfo.InvariantCulture.TextInfo.ToTitleCase(ctx.Compilation.GetSpecialType((SpecialType)x).ToDisplayString()))
    .ToList();

// allNames: [Bool, Char, Sbyte, Byte, Short, Ushort, Int, Uint, Long, Ulong, Decimal, Float, Double];
```

Unsigned types and `sbyte` seem problematic. It seems that for them it is better to check both variants of the name - `Ulong` and `ULong`, for example.

Instead, I would prefer to replace `StringComparison.Ordinal` with `StringComparison.OrdinalIgnoreCase` when looking up a method by name. It seems like this shouldn't cause any problems.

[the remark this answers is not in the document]
latonz: Using ignore case comparison sounds good to me 👍

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
TonEnfer: I don't think so, because `targetType` is always a non-`nullable` type.
In this case it turns out that it doesn't matter whether the method argument is `nullable` or not, `targetType` can always be assigned to it.
I'll rename the variable `targetType` to `nonNullableTargetType` and add a comment so that it doesn't cause questions.

Or do you see something that I don't? After your question, I'm in doubt

latonz: I don't think that would work for differences in generic type arguments? `List<string>` vs. `List<string?>`.

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertInstanceMethodMappingBuilder.cs
TonEnfer: In fact, the code here (as well as in `ParseMappingBuilder`) is unreachable, because `ctx.Target` is always non-Nullable

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,187 @@
...
+                var method = candidates.Find(x => symbolAccessor.ValidateSignature(x, nonNullableTargetType, sourceType));
+                if (method != null)
+                    return new StaticMethodMapping(method);
+            }
+            else // currently this branch is never executed because `targetType` is never null. See https://github.com/riok/mapperly/issues/1614
+            {
+                var method = candidates.Find(x => symbolAccessor.ValidateSignature(x, targetType, sourceType));
+                if (method != null)
+                    return new StaticMethodMapping(method);
+            }
```
latonz, line 123: I'd remove this branch and add a reference or a failing test case in #1614.

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -370,6 +370,85 @@ internal bool TryFindMemberPath(ITypeSymbol type, StringMemberPath path, [NotNul
...
+        }
+
+        for (var i = 0; i < method.Parameters.Length; i++)
+        {
+            if (method.Parameters[i] is not { IsParams: true } isParamsParameter)
+            {
+                if (!CanAssign(argTypes[i], method.Parameters[i].Type))
+                {
+                    return false;
+                }
```
latonz, line 396: Nit: an early continue could reduce the nesting level for the rest of the for loop body. Maybe it could even make sense to extract this into another method `ValidateParamater`

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -370,6 +370,85 @@ internal bool TryFindMemberPath(ITypeSymbol type, StringMemberPath path, [NotNul
...
+                    // for assignable to IEnumerable<T>
+                    elementType = iEnumerableInterface.TypeArguments.First();
+                }
+                else
+                {
+                    // for Span<T> and ReadOnlySpan<T>
+                    elementType = ((INamedTypeSymbol)method.Parameters[i].Type).TypeArguments.First();
+                }
+
+                var argsToEnd = new Span<ITypeSymbol>(argTypes, i, argTypes.Length);
```
latonz, line 420: Couldn't this be simplified to `argTypes.AsSpan(i)`?

test/Riok.Mapperly.Tests/Mapping/StaticMethodConversionTest.cs
```diff
@@ -0,0 +1,3006 @@
...
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class StaticMethodConversionTest
+{
+    [
+        Theory,
+        InlineData("byte", "class  A { public static A Create(byte source) => new(); }", "return global::A.Create(source);"),
+        InlineData("byte", "struct A { public static A Create(byte source) => new(); }", "return global::A.Create(source);"),
+        InlineData("byte", "class  A { public static A CreateFrom(byte source) => new(); }", "return global::A.CreateFrom(source);"),
+        InlineData("byte", "struct A { public static A CreateFrom(byte source) => new(); }", "return global::A.CreateFrom(source);"),
```
latonz, line 13: This code base usually structures one attribute per attributelist (`[Theory]\n[InlineData]\n...`)

test/Riok.Mapperly.Tests/Mapping/StaticMethodConversionTest.cs
```diff
@@ -0,0 +1,3006 @@
...
+            "return global::A.CreateFromByte(source);"
+        ),
+        InlineData(
+            "byte",
+            "struct A { public static A CreateFromByte(byte source) => new(); }",
+            "return global::A.CreateFromByte(source);"
+        ),
+        InlineData("byte", "class  A { public static A FromByte(byte source) => new(); }", "return global::A.FromByte(source);"),
+        InlineData("byte", "struct A { public static A FromByte(byte source) => new(); }", "return global::A.FromByte(source);"),
+        InlineData("sbyte", "class  A { public static A Create(sbyte source) => new(); }", "return global::A.Create(source);"),
```
latonz, line 26: This test suite seems overly extensive for the current purpose. In my opinion, a single representative test case for each scenario would be sufficient. For example:
* One test case per method name.
* One to handle keyword-specific cases.
* One for nullable and non-nullable value/reference types.
* One to verify method name case sensitivity.
This would ensure the key constellations are covered without unnecessary redundancy.

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,187 @@
...
+            yield return $"{from}{keywordName}{array}";
+
+            yield return $"{from}{array}";
+            yield return $"{create}{from}{array}";
+            yield return $"{create}{from}";
+            yield return create;
+        }
+        else
+        {
+            yield return $"{create}{from}{ctx.Source.Name}";
```
latonz, line 154: An early `yield break` could reduce the nesting level here...

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertStaticMethodMappingBuilder.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        if (ctx.Source.IsArrayType(out var arrayType))
+        {
+            yield return $"{create}{from}{arrayType.ElementType.Name}{array}";
+            yield return $"{from}{arrayType.ElementType.Name}{array}";
+
+            if (!arrayType.ElementType.HasKeyword(out var keywordName))
+                yield break;
+
+            yield return $"{create}{from}{keywordName}{array}";
```
latonz, line 143: After reviewing this PR code multiple times, I believe inlining the create, from, and array constants improves readability. While this codebase typically extracts "magic strings" into constants, in this specific case, inlining seems more appropriate as the sole purpose is to concatenate these strings. This approach keeps the code straightforward and easier to follow.

The same applies to `ConvertInstanceMethodMappingBuilder.GetMappingMethodNames`.

latonz (changes requested):
Thanks for the updates! I added my feedback 😊

src/Riok.Mapperly/Descriptors/MappingBuilders/ConvertInstanceMethodMappingBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+            // try to find method with equal nullability return type
+            var method = methodCandidates.Find(x => SymbolEqualityComparer.IncludeNullability.Equals(x.ReturnType, ctx.Target));
+            if (method is not null)
+                return new SourceObjectMethodMapping(ctx.Source, ctx.Target, method.Name);
+
+            if (!targetIsNullable)
+                continue;
+
+            // otherwise try to find method ignoring the nullability
+            method = methodCandidates.Find(x => SymbolEqualityComparer.Default.Equals(x.ReturnType, nonNullableTarget));
```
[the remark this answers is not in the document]
TonEnfer: https://github.com/riok/mapperly/pull/1616#discussion_r1871897178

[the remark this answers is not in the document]
latonz: My bad 👍

TonEnfer:
Hi @latonz. I fixed your comments, but I can't figure out how to make tests pass in CI/CD. Could you help me with this?

latonz (approved):

latonz:

PR #1717, opened 2025-02-11
feat: do not apply deep cloning to queryable projection mappings

latonz:
Never apply deep cloning to queryable projections, as these are projections and deep cloning doesn't have an effect on the projection.


latonz:
Technically, this is a breaking change, but since I can't think of any scenario where this could be useful or intentionally used, I'll release it as a feature for now. If there's any feedback regarding this during the preview releases, I'll postpone it to the next breaking change and likely make it configurable.

PR #1720, opened 2025-02-15
feat: support named mappings for existing target objects

### LINKED ISSUE #1649: Support existing target mappings with named mappings (`MapProperty.Use`)
Mapperly should support `MapProperty.Use` on mappings which target an existing instance.

<details>
<summary>Sample Test</summary>

```csharp
    [Fact]
    public Task ShouldUseReferencedMappingOnSelectedPropertiesWithExistingInstance()
    {
            """
            [MapProperty(nameof(A.Values1), nameof(B.Values1), Use = nameof(MapV1)]
            [MapProperty(nameof(A.Values2), nameof(B.Values2), Use = nameof(MapV2)]
            public partial B Map(A source);

            [UserMapping(Default = true)]
            private void DefaultMapping(List<string> source, List<string> target) {}
            private void MapV1(List<string> source, List<string> target) {}
            private void MapV2(List<string> source, List<string> target) {}
            """,
            """
            class A
            {
                public List<string> Values { get; } = [];
                public List<string> Values1 { get; } = [];
                public List<string> Values2 { get; } = [];
            }
            """,
            """
            class B
            {
                public List<string> Values { get; } = [];
                public List<string> Values1 { get; } = [];
                public List<string> Values2 { get; } = [];
            }
            """
        );
    }
```

</details>

See https://github.com/riok/mapperly/discussions/1648#discussion-7709404

### LINKED ISSUE #1700: MapProperty Use mapping function with existing object
According to the docs I can define user mapping function allowing me  to use the existing object: https://mapperly.riok.app/docs/configuration/existing-target/

E.g. in my case I tried this:
```cs
private void MapDocuments(DocumentDto[] target, [MappingTarget] ICollection<Document> source)
{
    ...
}
```

Then the `MapProperty` attribute allows us to define a custom mapper function like:
```cs
[MapProperty(nameof(MyEntityDto.Documents), nameof(MyEntity.Documents), Use = nameof(MapDocuments))]
public partial void Map(MyDto dto, [MappingTarget] MyEntity entity);
```

However above use case fails with the error: "RMG061: The referenced mapping named MapDocuments was not found"

`MapProperty` should allow us to use void methods that require the existing object.

 
Additonally:
In my specific case above I am using collection types. Initially I expected the `MapDocuments` method to be automaticly used as a UserMapping, without actually declaring `MapProperty`, but it never used my method, even when using the classic signature: `ICollection<Document > Map(DocumentDto[] dtos) {...} `

**Additional context**
Not sure if it's a feature or bug.

clegoz:
# feat: support named mappings for existing target objects

## Description

Fixes #1649
Fixes #1700

## Checklist


latonz:
Thanks for your contribution. Could you please explain the reason behind your choice of a different approach for implementing this compared to the new instance mappings (using `UseNamedMappingBuilder.TryBuildMapping`)? I'm concerned that your implementation doesn't seem to address diagnostics like `RMG072` and `RMG073`. Could you please add tests to verify this and maybe adjust your implementation to match the existing approach (see `UseNamedMappingBuilder.TryBuildMapping`)?

clegoz:
I made changes accordingly to warn when types do not match, and also added `CompositeMemberExistingTargetMapping` (not sure about the name) to map source and target types to the referenced mapping types.

latonz:
Thanks for the updates. I saw that you are still not using `UseNamedMappingBuilder`, why did you decide to not use that (there may be a reason I'm just not seeing directly while reviewing)? IMO it would make sense to use the same approach for the existing instance mappings as for the new instance mappings (add `public static IExistingTargetMapping? TryBuildExistingTargetMapping(MappingBuilderContext ctx)` to `UseNamedMappingBuilder`).

clegoz:
I tried to implement it in `UseNamedMappingBuilder` before couldn't figure out how to call `TryBuildExistingTargetMapping`. I realized that instead of using named mapping directly, I needed to build new mapping using the mapping key. It took some time, sorry about that.

Please let me know if there are any changes to be made.

latonz:
Thanks for this contribution 🥳

PR #1804, opened 2025-04-15
feat: Allow to disable null mismatch diagnostics RMG089 per member

### LINKED ISSUE #1613: Allow to disable null mismatch diagnostics (RMG089 and RMG090) for specific members
Add a new property `bool SuppressNullMismatchDiagnostic` to the `MapPropertyAttribute`. When set to true, `RMG089` and `RMG090` are not emitted.

See also https://github.com/riok/mapperly/pull/1592#discussion_r1845968984 and https://github.com/riok/mapperly/issues/602#issuecomment-1662204524.

3sRykaert:
# [Allow to disable null mismatch diagnostics RMG089](https://github.com/riok/mapperly/pull/1804/commits/e0c30ec39c3ded451220fb4606f61e39ba0d97e8)

## Description

Code change for https://github.com/riok/mapperly/issues/1613 (only yet for RMG089)
Any feedback is already appreciated.

Fixes #1613 

## Checklist


src/Riok.Mapperly/Descriptors/TypeMappingConfiguration.cs
```diff
@@ -7,7 +7,12 @@ namespace Riok.Mapperly.Descriptors;
 /// <param name="StringFormat">The format to apply to <see cref="IFormattable"/>.</param>
 /// <param name="FormatProviderName">The name of the format provider to apply to <see cref="IFormattable"/>.</param>
 /// <param name="UseNamedMapping">The name of the mapping to use to convert the source type to the target type.</param>
-public record TypeMappingConfiguration(string? StringFormat = null, string? FormatProviderName = null, string? UseNamedMapping = null)
+public record TypeMappingConfiguration(
+    string? StringFormat = null,
+    string? FormatProviderName = null,
+    string? UseNamedMapping = null,
+    bool SuppressNullMismatchDiagnostic = false
```
kakone, line 14: Comment is missing for `SuppressNullMismatchDiagnostic` parameter

3sRykaert: added now

3sRykaert:
I'm still struggling with the unittest for RMG90 when setting SuppressNullMismatchDiagnostic to true, that test is failing cause the configation of the SuppressNullMismatchDiagnostic stays false, but I can't figure out why. Anyone able to help with this?

latonz:
> I'm still struggling with the unittest for RMG90 when setting SuppressNullMismatchDiagnostic to true, that test is failing cause the configation of the SuppressNullMismatchDiagnostic stays false, but I can't figure out why. Anyone able to help with this?

You should be able to debug this by adding a conditional breakpoint in `Riok.Mapperly.Diagnostics.DiagnosticCollection.ReportDiagnostic` and setting the condition to the diagnostic id you want to debug (e.g. `descriptor.Id == "RMG090"`). With this breakpoint, you can debug the failing unit test. Once the breakpoint is hit, you can walk up the stack trace to identify why the diagnostic was reported.

src/Riok.Mapperly/Configuration/MemberMappingConfiguration.cs
```diff
@@ -18,6 +18,8 @@ public MemberMappingConfiguration(IMemberPathConfiguration Target)
 
     public string? Use { get; set; }
 
+    public bool SuppressNullMismatchDiagnostic { get; set; }
+
     public bool IsValid => Use == null || FormatProvider == null && StringFormat == null;
 
     public TypeMappingConfiguration ToTypeMappingConfiguration() => new(StringFormat, FormatProvider, Use);
```
kakone, line 25:
```suggestion
    public TypeMappingConfiguration ToTypeMappingConfiguration() => new(StringFormat, FormatProvider, Use, SuppressNullMismatchDiagnostic);
```

src/Riok.Mapperly/Descriptors/MappingBuilders/NullableMappingBuilder.cs
```diff
@@ -37,7 +37,7 @@ private static bool TryBuildNonNullableMappingKey(MappingBuilderContext ctx, out
 
         mappingKey = new TypeMappingKey(sourceNonNullable ?? ctx.Source, targetNonNullable ?? ctx.Target);
 
-        if (sourceIsNullable && !targetIsNullable)
+        if (ctx.MappingKey.Configuration.SuppressNullMismatchDiagnostic == false && sourceIsNullable && !targetIsNullable)
```
kakone, line 40: `ctx.MappingKey.Configuration.SuppressNullMismatchDiagnostic` will always be `false` because if you look at the `TypeMappingKey` class, `TypeMappingConfiguration.Default` will be used.

I'm not really sure what needs to be done but I think it's necessary to use `ctx.Configuration.Members.ExplicitMappings` :
```csharp
if (sourceIsNullable && !targetIsNullable && !SuppressNullMismatchDiagnostic(ctx))
```
```csharp
private static bool SuppressNullMismatchDiagnostic(MappingBuilderContext ctx)
{
    return ctx.Configuration.Members.ExplicitMappings
        .FirstOrDefault(x => string.Equals(x.Target.RootName, ctx.Target.Name, StringComparison.Ordinal))
        ?.SuppressNullMismatchDiagnostic is true;
}
```

3sRykaert: > `ctx.MappingKey.Configuration.SuppressNullMismatchDiagnostic` will always be `false` because if you look at the `TypeMappingKey` class, `TypeMappingConfiguration.Default` will be used.
> 
> I'm not really sure what needs to be done but I think it's necessary to use `ctx.Configuration.Members.ExplicitMappings`.

This seems indeed the way to go:
image
But can i always use "FirstOrDefault" or how do I know which explicit mapping I need to take?

3sRykaert:
> > I'm still struggling with the unittest for RMG90 when setting SuppressNullMismatchDiagnostic to true, that test is failing cause the configation of the SuppressNullMismatchDiagnostic stays false, but I can't figure out why. Anyone able to help with this?
> 
> You should be able to debug this by adding a conditional breakpoint in `Riok.Mapperly.Diagnostics.DiagnosticCollection.ReportDiagnostic` and setting the condition to the diagnostic id you want to debug (e.g. `descriptor.Id == "RMG090"`). With this breakpoint, you can debug the failing unit test. Once the breakpoint is hit, you can walk up the stack trace to identify why the diagnostic was reported.

Yes, I have already done that, but I find it too complex to find out why `SuppressNullMismatchDiagnostic` remains false here (although i specify it in the test as true)
image
I think it's cause in the unittest the `TypeMappingConfiguration.Default `is used (i believe this is the same as @kakone says here above)
image

src/Riok.Mapperly/Configuration/MemberMappingConfiguration.cs
3sRykaert: > 

thnx, that is indeed also missing. Pushed that one

latonz:
The `NullableTypeWithSuppressNullMismatchDiagnosticTrueShouldIgnoreWarningRMG090` test fails due to the following:
* `[MapProperty(\"IdA\", \"IdB\", SuppressNullMismatchDiagnostic = true)]` ignores RMG090 for the mapping of the property `IdA` to `IdB`
* The user-defined mapping `B Map(A? source)` also has a null mismatch. The RMG090 gets emitted for this mapping, but not for the property mapping `IdA` => `IdB`.

IMO this is expected behaviour and could be resolved with a `SuppressMessage` attribute, couldn't it?

src/Riok.Mapperly/Descriptors/MappingBuilders/NullableMappingBuilder.cs
latonz, line 40:
```suggestion
        if (!ctx.MappingKey.Configuration.SuppressNullMismatchDiagnostic && sourceIsNullable && !targetIsNullable)
```

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFlatteningTest.cs
```diff
@@ -814,4 +814,89 @@ public void ManualUnflattenedPropertyDeepNullablePathShouldNotNullInitialize()
...
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            "[MapProperty(\"IdA\", \"IdB\", SuppressNullMismatchDiagnostic = true)] private partial B Map(A source);",
+            "class A { public int? IdA { get; set; } }",
+            "class B { public int IdB { get; set; } }"
+        );
+
+        var mapperGenerationResult = TestHelper.GenerateMapper(source, TestHelperOptions.AllowDiagnostics);
+
+        mapperGenerationResult.Diagnostics.Should().BeEmpty();
```
latonz, line 887: Why use `TestHelperOptions.AllowDiagnostics` and asserting for now diagnostics at the same time? Same in other added tests...

latonz (changes requested):
Thanks for your contribution! 😊

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFlatteningTest.cs
3sRykaert: I don't know the source good enough how to check for "No Diagnostics" to be "thrown". Any suggestion how I can achieve that?

3sRykaert:
> The `NullableTypeWithSuppressNullMismatchDiagnosticTrueShouldIgnoreWarningRMG090` test fails due to the following:
> 
> * `[MapProperty(\"IdA\", \"IdB\", SuppressNullMismatchDiagnostic = true)]` ignores RMG089 for the mapping of the property `IdA` to `IdB`
> * The user-defined mapping `B Map(A? source)` also has a null mismatch. The RMG090 gets emitted for this mapping, but not for the property mapping `IdA` => `IdB`.
> 
> So RMG090 is always emitted on a type mapping but not on a member mapping. So I'm not sure whether it makes sense to suppress it via a member on the `MapProperty` attribute as it is not related to a member. Couldn't these RMG090 already be suppressed with a `SuppressMessage` attribute? Otherwise we would need to think about adding `SuppressNullMismatchDiagnostic` to the `UserMapping` attribute 🤔

I was just trying to do what was said in #1613 . So if I understand, correctly, this ain't possible with the `MapProperty` attribute for RMG090?

latonz:
@3sRykaert When I wrote the issue, I didn't conduct a thorough analysis. In my opinion, it works for RMG090 when dealing with generated mappings, but not for user-defined mappings.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFlatteningTest.cs
latonz: If you don't provide `TestHelperOptions.AllowDiagnostics`, it defaults to `TestHelperOptions.Default`, which throws (and fails the test) on any diagnostic except for `RMG066`, which is ignored by default. So for your use case, you should simply be able to remove the `TestHelperOptions.AllowDiagnostics` argument.

3sRykaert:
@latonz maybe we can then split up the issues in 2 separate ones:

- one for RMG089 which can be this
- and one for RM090 which is still todo

If you agree, I'll adjust the code to only keep RMG089 into account.

3sRykaert (commented):
changed as suggested

latonz (approved):

PR #1833, opened 2025-05-11
Include mapping configuration

### LINKED ISSUE #513: Include mapping configuration
Introduce two new attributes to include the mapping configuration of another mapper / mapping method.
Each mapping configuration is identified by a name which is by default the name of the method, but can be overwritten by `NamedMappingAttribute`, eg. `[NamedMapping("myName")]`.
A mapping configuration can be applied to the current mapping method by applying `IncludeMappingConfigurationAttribute`: `[IncludeMappingConfiguration("myName")]` / `[IncludeMappingConfiguration(nameof(OtherMappingMethod))]`.
If an `IncludeMappingConfigurationAttribute` references an ambiguous mapping configuration, a diagnostic should be emitted.

Proposed API:
```c#
[AttributeUsage(AttributeTargets.Method)]
public class IncludeMappingConfigurationAttribute : Attribute
{
    public IncludeMappingConfigurationAttribute(string name);
}

[AttributeUsage(AttributeTargets.Method)]
public class NamedMappingAttribute : Attribute
{
    public NamedMappingAttribute(string name);
}
```

Usage:
```c#
class Fruit { public string Name { get; set; } public decimal PricePerUnit { get; set; } }
class FruitDto { public string Title { get; set; } }

class Apple : Fruit { public int Weight { get; set; } }
class AppleDto : FruitDto { public int Weight { get; set; } }

[Mapper]
public partial static class MyMapper
{
    [MapProperty(nameof(Fruit.Name), nameof(FruitDto.Title))]
    [MapperIgnoreSource(nameof(Fruit.PricePerUnit))]
    private partial FruitDto ToFruit(Fruit fruit);

    [IncludeMappingConfiguration(nameof(ToFruit))]
    public partial static AppleDto Map(Apple apple);
}
```

faddiv:
# Include mapping configuration

## Work in progress Info

Hi @latonz,

I decided to try implementing this feature, because I feel this is an important feature, and we also could use.
Since the basics of this started working as expected, and I found the integration points that feel correct, I think I can create a complete implementation.
Currently, it is still work in progress, but this way you are aware that I'm working on it and you can guide me.

## Description

Introduce two new attributes to include the mapping configuration of another mapper/mapping method.
Each mapping configuration is identified by a name which is by default the name of the method, but can be overwritten by NamedMappingAttribute, eg. [NamedMapping("myName")].
A mapping configuration can be applied to the current mapping method by applying IncludeMappingConfigurationAttribute: [IncludeMappingConfiguration("myName")] / [IncludeMappingConfiguration(nameof(OtherMappingMethod))].
If an IncludeMappingConfigurationAttribute references an ambiguous mapping configuration, a diagnostic should be emitted.

API:
```csharp
[AttributeUsage(AttributeTargets.Method)]
public class IncludeMappingConfigurationAttribute : Attribute
{
    public IncludeMappingConfigurationAttribute(string name);
}

[AttributeUsage(AttributeTargets.Method)]
public class NamedMappingAttribute : Attribute
{
    public NamedMappingAttribute(string name);
}
```

Usage:
```csharp
class Fruit { public string Name { get; set; } public decimal PricePerUnit { get; set; } }
class FruitDto { public string Title { get; set; } }

class Apple : Fruit { public int Weight { get; set; } }
class AppleDto : FruitDto { public int Weight { get; set; } }

[Mapper]
public partial static class MyMapper
{
    [MapProperty(nameof(Fruit.Name), nameof(FruitDto.Title))]
    [MapperIgnoreSource(nameof(Fruit.PricePerUnit))]
    private partial FruitDto ToFruit(Fruit fruit);

    [IncludeMappingConfiguration(nameof(ToFruit))]
    public partial static AppleDto Map(Apple apple);
}
```

Fixes #513

## Checklist


test/Riok.Mapperly.Tests/Mapping/IncludeMappingConfigurationTest.cs
```diff
@@ -0,0 +1,148 @@
...
+            [Mapper]
+            [UseStaticMapper(typeof(OtherMapper))]
+            public static partial class TestMapper
+            {
+                [IncludeMappingConfiguration(nameof(OtherMapper.Map))]
+                public static partial B MapAnother(A a);
+            }
+            """
+        );
+        return TestHelper.VerifyGenerator(source);
```
faddiv, line 106: Should external mapping inclusion work like this?

faddiv: Ahh, nope. You wrote about a type parameter in the include mapping configuration.

latonz:
Thank you for this contribution — at first glance, it looks pretty good! 😊 
Unfortunately, I don't currently have the time to review a PR this big. Hopefully, I will have the time this week or at the beginning of next week. Thank you for your understanding.

test/Riok.Mapperly.Tests/Mapping/IncludeMappingConfigurationTest.cs
latonz: Originally, my concern was that using only the method name for external mappers could result in name collisions. To avoid this, the idea was to use a qualified name for external mappers - for example:
* `[IncludeMappingConfiguration("OtherMapper.Map")]`
* `[IncludeMappingConfiguration(nameof(@OtherMapper.Map))]`
* `[IncludeMappingConfiguration<OtherMapper>(nameof(Map))]`
* `[IncludeMappingConfiguration(typeof(OtherMapper), nameof(Map))]`

I'm fine with not supporting these qualified names for now. However, in that case, I think we need to clearly define a resolution order to handle potential conflicts (we may need to do this anyway? To fully answer this, I’ll need to take a closer look at the topic and the PR. I’ll dig into it and follow up...):
1. Methods in the same mapper
2. Methods in base classes
3. Methods in external mappers

If a name collision occurs, we should report a diagnostic. I'm not 100% sure, if we should report it for all name collisions though (if there is a matching named mapping present in the current mapper and one in an external mapper, the one in the current mapper should probably take precedence without a diagnostic or only an informational diagnostic 🤔).

faddiv: In the end I dropped this part.

faddiv:
> Thank you for this contribution — at first glance, it looks pretty good! 😊 Unfortunately, I don't currently have the time to review a PR this big. Hopefully, I will have the time this week or at the beginning of next week. Thank you for your understanding.

No problem, it is still a long way to go. 🙂 
Currently, I have at most one hour daily to work on this anyway.

test/Riok.Mapperly.Tests/Mapping/IncludeMappingConfigurationTest.cs
```diff
@@ -0,0 +1,232 @@
...
+            class B { public string DestinationName1 { get; set; } }
+            class BDerived : B { public string DestinationName2 { get; set; } }
+            """
+        );
+
+        return TestHelper.VerifyGenerator(source);
+    }
+
+    [Fact]
+    public Task ResolvesMatchingCandidateIfOnlyOneMatches()
```
faddiv, line 109: The included mapping has been slightly reworked. Now it ignores a mapping if the types mapped don't match. But if multiple could be used, then it still reports it.
The error messages should be decided. I don't think the best ones are used in every case.

faddiv:
Added the `NamedMapping` implementation too. Currently, it is just a simple alias and is treated as the default name. If duplicated with unnamed mapping it is reported.
Question: What should happen if a mapping has the same name as an unnamed mapping or another named mapping?
Also, I think this named mapping should be used by another setting that references methods by name. Like `Use` on `MapProperty`.

faddiv:
I saw that nested mappings processed separatelly. First I just tried to use recursive mapping there too. This lead to duplicate error reportings.
Now I created a separate class: `IncludeMappingBuilder`. This makes the connections and error reporting, and the generation parts go through the mapping contexts. Since I think the `MappingBuilderContext` is intended to be immutable, I kept it this way.

faddiv:
I did everything needed in this PR and opened it for a more detailed code review.

docs/docs/configuration/analyzer-diagnostics/index.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 18
+sidebar_position: 20
```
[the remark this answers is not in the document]
faddiv: As I see, these numbers should be in strict order. Since I didn't put the new documentation in the last place, I adjusted the rest of the numbers. I also adjusted the non-unique numbers so that the sort order is clear.

docs/docs/configuration/include-mapping.md
```diff
@@ -0,0 +1,87 @@
...
+    [IncludeMappingConfiguration(nameof(ToFruit))]
+    public partial static AppleDto Map(Apple apple);
+}
+```
+
+In this example, the `Map` method for `Apple` reuses the configuration from `ToFruit`, ensuring consistent property mapping.
+
+## Diagnostics
+
+If an `IncludeMappingConfigurationAttribute` refers to an ambiguous mapping configuration (e.g., multiple configurations exist with the same name), the mapper will emit a diagnostic message to help you resolve the ambiguity. This configuration only considers the compatible mappings.
```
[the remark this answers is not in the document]
faddiv: As I said above, in this PR I would like to restrict the implementation to the same class. If this is complete, you should create a new issue where the implementation plan can be laid out.
I'm happy to implement that too.
Until then, I will create a Restrictions section, add this, and also, I think the restriction on mapped types would be in a better place in this new section.

docs/docs/configuration/include-mapping.md
```diff
@@ -0,0 +1,87 @@
...
+``` 
+
+Or refer to a method directly:
+
+```csharp
+[IncludeMappingConfiguration(nameof(ToFruit))] 
+public partial static AppleDto Map(Apple apple);
+```
+
+> **Note:** The `IncludeMappingConfigurationAttribute` only uses a mapping if the mapped types are the same or base
```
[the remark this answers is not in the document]
faddiv: This is the ssection, I moved to the restrictions.

docs/docs/configuration/include-mapping.md
```diff
@@ -0,0 +1,87 @@
...
+> types of the mapped type. This means that the configuration will only be applied if the source and target types of
+> the included mapping are compatible. Specifically, the types must either be the same, or the included mapping’s
+> types must be base types of the method where the attribute is applied.
```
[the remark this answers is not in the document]
faddiv: I should be more careful. I use AI for document generation. Usually, it is better than me, especially in proper grammar usage.

docs/docs/configuration/include-mapping.md
```diff
@@ -0,0 +1,87 @@
+﻿sidebar_position: 17
+description: Include settings from another mapper method
```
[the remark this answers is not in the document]
faddiv: Changed. I also updated the file name.

docs/docs/configuration/include-mapping.md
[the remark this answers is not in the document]
latonz: You state `the mapper will emit a diagnostic message to help you resolve the ambiguity`. My idea would be to replace this with `the mapper will emit RMGXXX to help you resolve the ambiguity`.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -37,7 +38,17 @@ SupportedFeatures supportedFeatures
...
+            new MembersMappingConfiguration(
+                null,
+                [],
+                [],
+                [],
+                [],
+                [],
+                mapper.IgnoreObsoleteMembersStrategy,
+                mapper.RequiredMappingStrategy,
+                IncludedMapping: null
```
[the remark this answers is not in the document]
latonz: This is not relevant anymore if we switch to the proposed approach.

src/Riok.Mapperly.Abstractions/NamedMappingAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Riok.Mapperly.Abstractions;
+
+/// <summary>
+/// An attribute that specifies a custom name for the mapping configuration defined by the attributed method.
+/// This name can be used to uniquely identify the mapping configuration and reference it from other mapping methods.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method)]
+[Conditional("MAPPERLY_ABSTRACTIONS_SCOPE_RUNTIME")]
+public sealed class NamedMappingAttribute : Attribute
```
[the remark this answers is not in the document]
faddiv: Removed. In my opinion, this would be good for other named references as well. At least then it can be implemented together.

test/Riok.Mapperly.Tests/Mapping/IncludeMappingConfigurationTest.cs
```diff
@@ -0,0 +1,417 @@
...
+
+public class IncludeMappingConfigurationTest
+{
+    [Fact]
+    public Task UsesMapPropertyFromTargetMapper()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+            [MapProperty(nameof(A.SourceName), nameof(B.DestinationName))]
+            partial B OtherMapper(A a);
```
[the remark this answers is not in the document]
faddiv: Updated all in the tests.

test/Riok.Mapperly.Tests/Mapping/IncludeMappingConfigurationTest.cs
```diff
@@ -0,0 +1,417 @@
...
+            """,
+            "class A { public string SourceName { get; set; } }",
+            "class B { public string DestinationName { get; set; } }"
+        );
+
+        return TestHelper.VerifyGenerator(source);
+    }
+
+    [Fact]
+    public Task AppliesOnDifferentMethodDeclaration()
```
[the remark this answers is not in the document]
faddiv: Fixed

test/Riok.Mapperly.Tests/Mapping/IncludeMappingConfigurationTest.cs
```diff
@@ -0,0 +1,417 @@
...
+            """,
+            "class A { public string Source { get; set; } }",
+            "class B { public string Target { get; set; } }"
+        );
+
+        return TestHelper.VerifyGenerator(source);
+    }
+
+    [Fact]
+    public Task AppliesOnDifferentMethodDeclaration2()
```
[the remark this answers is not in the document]
faddiv: Fixed

faddiv:
> It can then be passed to the mapping pipeline without requiring any further adjustments. WDYT?

I started this alternate implementation. I keep it in a separate branch until every issue is fixed. You can sneak peek here: https://github.com/faddiv/mapperly/tree/feature/Include-mapping-configuration-alternate-impl

latonz:
Thanks for the updates, it's looking promising! 😊
Instead of reading the`IncludeMappingConfigurationAttribute` in the context, I’d suggest moving that logic into the configuration reader, since it conceptually still belongs to the configuration resolution phase.

src/Riok.Mapperly/Configuration/EnumMappingConfiguration.cs
```diff
@@ -15,4 +15,15 @@ EnumNamingStrategy NamingStrategy
 )
 {
     public bool HasExplicitConfigurations => ExplicitMappings.Count > 0 || IgnoredSourceMembers.Count > 0 || IgnoredTargetMembers.Count > 0;
+
+    public EnumMappingConfiguration MergeWith(EnumMappingConfiguration? result2Enum)
```
faddiv: I created these `MergeWith` methods to do the merging. Is this ok, or should I do something like `MapperConfigurationMerger`.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -9,17 +13,26 @@ namespace Riok.Mapperly.Configuration;
 public class MapperConfigurationReader
 {
     private readonly AttributeDataAccessor _dataAccessor;
+    private readonly MappingBuilder _mappings;
+    private readonly GenericTypeChecker _genericTypeChecker;
+    private readonly DiagnosticCollection _diagnostics;
```
faddiv: I extended the `MapperConfigurationReader` with the required dependencies. I moved the diagnostics here also. So it is now a dependency instead of a parameter.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -46,26 +59,122 @@ SupportedFeatures supportedFeatures
 
     public MappingConfiguration MapperConfiguration { get; }
 
-    public MappingConfiguration BuildFor(
+    public MappingConfiguration BuildFor(SimpleMappingBuilderContext ctx, MappingConfigurationReference reference, bool supportsDeepCloning)
```
faddiv: I had to pass the `SimpleMappingBuilderContext` because of the dependencies in the `MappingBuilder`. There are several static methods that use all kinds of dependencies from the context.
I recommend refactoring these into injectable dependencies, so this type of entanglement could be cleaned up.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -46,26 +59,122 @@ SupportedFeatures supportedFeatures
...
+
+        var includeMapping = _dataAccessor.AccessFirstOrDefault<IncludeMappingConfigurationAttribute>(reference.Method)?.Name;
+        if (includeMapping is null)
+        {
+            return configuration;
+        }
+
+        var typeMapping =
+            (ITypeMapping?)_mappings.FindOrResolveNamed(ctx, includeMapping, out var ambiguousName)
+            ?? _mappings.FindExistingInstanceNamedMapping(ctx, includeMapping, out ambiguousName);
```
faddiv: This is where the mapping builder is needed with the mapping context. At this point, the mapping collection might not have walked over the referenced map. This is true for the existing instance. I think I need to create or modify the current test to ensure it works correctly in that case.

src/Riok.Mapperly/Configuration/MembersMappingConfiguration.cs
```diff
@@ -9,7 +9,7 @@ public record MembersMappingConfiguration(
     IReadOnlyCollection<MemberValueMappingConfiguration> ValueMappings,
     IReadOnlyCollection<MemberMappingConfiguration> ExplicitMappings,
     IReadOnlyCollection<NestedMembersMappingConfiguration> NestedMappings,
-    IgnoreObsoleteMembersStrategy IgnoreObsoleteMembersStrategy,
+    IgnoreObsoleteMembersStrategy? IgnoreObsoleteMembersStrategy,
     RequiredMappingStrategy RequiredMappingStrategy
```
faddiv: For now, I made only the `IgnoreObsoleteMembersStrategy` nullable, so it can also be included.
Maybe the `RequiredMappingStrategy` is also needed. This could be tested with a new test.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: Also, I had to give up the correct method selection. The test is still there, but at this point, it is not feasible to select from multiple methods. It needs to be refactored in such a way that the client gets all the duplication and selects one. If you agree, I won't pursue this, and remove the test. (Maybe a future enchantment could solve it.)

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/IgnoredMembersBuilder.cs
```diff
@@ -92,7 +92,7 @@ private static IEnumerable<string> GetIgnoredAtMemberMembers(MappingBuilderConte
 
     private static IEnumerable<string> GetIgnoredObsoleteMembers(MappingBuilderContext ctx, MappingSourceTarget sourceTarget)
     {
-        var obsoleteStrategy = ctx.Configuration.Members.IgnoreObsoleteMembersStrategy;
+        var obsoleteStrategy = ctx.GetIgnoreObsoleteMembersStrategy();
```
faddiv: I extracted this method. Which returns with the value, or with the default, so it can be merged from the included mapping.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/MembersMappingStateBuilder.cs
```diff
@@ -1,4 +1,4 @@
-using Riok.Mapperly.Configuration;
+﻿using Riok.Mapperly.Configuration;
```
faddiv, line 1: Encoding is changed here. I'm not sure what I should do with it.

src/Riok.Mapperly/Diagnostics/DiagnosticCollection.cs
```diff
@@ -29,6 +29,21 @@ internal void ReportDiagnostic(DiagnosticDescriptor descriptor, Location? locati
...
+                ReferenceEquals(d.Descriptor, diagnostic.Descriptor)
+                && Equals(d.Location, diagnostic.Location)
+                && Equals(d.Severity, diagnostic.Severity)
+                && Equals(d.WarningLevel, diagnostic.WarningLevel)
+                && Equals(d.GetMessage(), diagnostic.GetMessage())
+            )
+        )
+        {
+            return;
+        }
```
faddiv: I extended the diagnostic collection with a report duplication check. Unfortunately, the reports are repeated several times because the `ConfigurationReader` runs several times.
First, I tried to separate the `MappingBuilderContext` to copy only and create scenarios, but it breaks lots of tests, no matter what. This made some unrelated tests change, since they also output duplicate reports.

faddiv: Also, I had to create my equality check, since the `diagnostic.Equals` has a bug: On args, the equality check reports false even when they should be equal. (They use sequenceEqual to compare lists, but the comparison might check only reference equality.)

test/Riok.Mapperly.IntegrationTests/Mapper/TestMapper.cs
```diff
@@ -109,6 +110,9 @@ public TestObjectDto MapToDto(TestObject src)
         [UserMapping(Default = false)]
         private int ComputeSum(TestObject testObject) => testObject.SumComponent1 + testObject.SumComponent2;
 
+        [IncludeMappingConfiguration("MapToDtoInternal-Renamed")]
+        private partial TestObjectDto MapToDtoInternalInclude(TestObject testObject);
```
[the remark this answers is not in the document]
faddiv: Explicit naming is now out of scope, so I think this can be closed. (This changed to implicit.)

test/Riok.Mapperly.Tests/Mapping/NamedMappingTest.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Riok.Mapperly.Tests.Mapping;
+
+public class NamedMappingTest
+{
+    [Fact]
+    public Task IncludeMappingConfigurationUsesNamedAlias()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+            [NamedMapping("mapperAlias")]
```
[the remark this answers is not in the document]
faddiv: Explicit naming is now out of scope, so I think this can be closed.

test/Riok.Mapperly.Tests/_snapshots/IncludeMappingConfigurationTest.IncludesConfigurationFromBaseClassMapping#Mapper.g.verified.cs
```diff
@@ -0,0 +1,22 @@
...
+        var target = new global::B();
+        target.DestinationName1 = a.SourceName1;
+        return target;
+    }
+
+    [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "0.0.1.0")]
+    public static partial global::BDerived MapDerived(global::ADerived a)
+    {
+        var target = new global::BDerived();
+        target.DestinationName2 = a.SourceName2;
```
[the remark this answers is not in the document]
faddiv: This sort order comes from the `SymbolAccessor.GetAllMembersCore`. It first gets the members from the current type, then from the base type. Additionally, it goes through a dictionary and so on.
So I don't think I can affect it.

test/Riok.Mapperly.Tests/_snapshots/IncludeMappingConfigurationTest.ReportsOnMultipleCandidatesButNoneMatches.verified.txt
```diff
@@ -0,0 +1,92 @@
...
+      Location: /*
+
+[IncludeMappingConfiguration(nameof(UnrelatedMapper))]
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+partial B MapAnother(A a);
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+}
+*/
+ : (17,0)-(18,26),
+      Message: The referenced mapping named UnrelatedMapper was not found,
```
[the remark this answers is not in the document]
faddiv: This result has changed. But now any ambiguity is reported. So there is not much point in this test now.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: `MappingBuilder _mappings;` is replaced with MappingCollection. I realised that I don't need that.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: Now I don't need to pass this.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: Only the second comment stands now.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -46,26 +58,122 @@ SupportedFeatures supportedFeatures
...
+        if (!visitedMethods.Add(methodSymbol))
+        {
+            _diagnostics.ReportDiagnostic(DiagnosticDescriptors.CircularReferencedMapping, methodSymbol, methodSymbol.ToDisplayString());
+            return null;
+        }
+
+        var includedReference = new MappingConfigurationReference(methodSymbol, typeMapping.SourceType, typeMapping.TargetType);
+
+        var includedConfiguration = BuildWithIncludedMappings(visitedMethods, includedReference, supportsDeepCloning);
+        return includedConfiguration != null ? configuration.MergeWith(includedConfiguration) : configuration;
```
faddiv: I wonder what should happen, if the included mapping ignores something, but the current one maps it. 
Or both maps something.
Or both maps something differently.

faddiv:
I made a few improvements on the tests.
What do you think? When will you have time for the review? Currently, I have nothing more to improve on the implementation.

latonz:
I'll have a look this week or the beginning of the next week. Thank you for your efforts!

docs/docs/configuration/reusing-mapping-configurations.md
```diff
@@ -0,0 +1,96 @@
+﻿---
+sidebar_position: 17
+description: Reusing Mapping Configurations
+---
+
+# Including Mapping Configurations
+
+Mapperly supports reusing mapping configurations across different mapping methods using two attributes:
+`NamedMappingAttribute` and `IncludeMappingConfigurationAttribute`. This enables sharing and modularizing
```
latonz: The named mapping attribute is not implemented in this PR, is it? Include a sample with `nameof()`.

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
```diff
@@ -202,3 +202,13 @@ RMG017  | Mapper   | Warning  | An init only member can have one configuration a
...
 RMG027  | Mapper   | Warning  | A constructor parameter can have one configuration at max
 RMG028  | Mapper   | Warning  | Constructor parameter cannot handle target paths
+
+## Release 4.1
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+RMG091 | Mapper | Error | Circular referenced mapping detected
```
latonz: Adjust message to `Circular referenced mapping configuration detected`

src/Riok.Mapperly/Configuration/EnumMappingConfiguration.cs
latonz: I like the current approach, however, I think the naming can be optimized.
WDYT about `EnumMappingConfiguration Include(EnumMappingConfiguration? configToInclude)`?

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -46,26 +58,122 @@ SupportedFeatures supportedFeatures
...
             MapperConfiguration.SupportedFeatures
         );
+
+        var includeMapping = _dataAccessor.AccessFirstOrDefault<IncludeMappingConfigurationAttribute>(reference.Method)?.Name;
+        if (includeMapping is null)
+        {
+            return configuration;
+        }
+
+        var typeMapping =
```
latonz, line 100: Nit: IMO the readability could be improved a bit by moving this and following parts of the method to a `MappingConfiguration BuildFor(string name, bool supportsDeepCloning)`. WDYT?

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -46,26 +58,122 @@ SupportedFeatures supportedFeatures
...
+            (ITypeMapping?)_mappings.FindNamedNewInstanceMapping(includeMapping, out var ambiguousName)
+            ?? _mappings.FindExistingInstanceNamedMapping(includeMapping, out ambiguousName);
+        var methodSymbol = typeMapping switch
+        {
+            UserDefinedNewInstanceMethodMapping udm => udm.Method,
+            UserDefinedExistingTargetMethodMapping udm => udm.Method,
+            _ => null,
+        };
```
latonz: Couldn't you simply check for `IUserMapping`?

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -46,26 +58,122 @@ SupportedFeatures supportedFeatures
...
+            return null;
+        }
+
+        var includedReference = new MappingConfigurationReference(methodSymbol, typeMapping.SourceType, typeMapping.TargetType);
+
+        var includedConfiguration = BuildWithIncludedMappings(visitedMethods, includedReference, supportsDeepCloning);
+        return includedConfiguration != null ? configuration.MergeWith(includedConfiguration) : configuration;
+    }
+
+    private bool IsMappingValid(
```
latonz: IMO a more idiomatic name would fit better here, e.g. `ValidateIncludedMapping`

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
latonz: For now, I wouldn't special-case this for now and treat it the same as it would have been applied to the mapping method directly.

src/Riok.Mapperly/Configuration/MembersMappingConfiguration.cs
latonz: Yep I think required strategy is also needed.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/IgnoredMembersBuilder.cs
latonz: Couldn't the `GetIgnoreObsoleteMembersStrategy` method be moved into the `MappingConfiguration` class?

src/Riok.Mapperly/Descriptors/SimpleMappingBuilderContext.cs
```diff
@@ -83,6 +86,13 @@ public virtual bool IsConversionEnabled(MappingConversionType conversionType) =>
     public void ReportDiagnostic(DiagnosticDescriptor descriptor, ISymbol? symbolLocation, params object[] messageArgs) =>
         _diagnostics.ReportDiagnostic(descriptor, symbolLocation?.GetSyntaxLocation() ?? _diagnosticLocation, messageArgs);
 
+    public IgnoreObsoleteMembersStrategy GetIgnoreObsoleteMembersStrategy()
+    {
+        return Configuration.Members.IgnoreObsoleteMembersStrategy.GetValueOrDefault(
+            _configurationReader.MapperConfiguration.Members.IgnoreObsoleteMembersStrategy.GetValueOrDefault()
+        );
```
latonz: Couldn't this be simplified?

```suggestion
        return Configuration.Members.IgnoreObsoleteMembersStrategy.GetValueOrDefault(
            Configuration.Mapper.IgnoreObsoleteMembersStrategy
        );
```

src/Riok.Mapperly/Diagnostics/DiagnosticCollection.cs
latonz: Mhmm this adds a `O(n)` overhead for each diagnostic added 🤔 
Couldn't we use a map of resolved configurations in the configuration reader?

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -763,6 +763,33 @@ public static class DiagnosticDescriptors
         true
     );
 
+    public static readonly DiagnosticDescriptor CircularReferencedMapping = new(
+        "RMG091",
+        "Circular referenced mapping detected",
```
latonz: Only the mapping configuration has a circular reference, but not the mapping itself. I'd adjust the message accordingly. e.g. `Circular included mapping configuration detected`.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
latonz: Removing the test sgtm.

src/Riok.Mapperly/AnalyzerReleases.Shipped.md
faddiv: This is different from the one you said at the DiagnosticDescriptor. I applied which is there for now.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: It is removed.

src/Riok.Mapperly/Configuration/EnumMappingConfiguration.cs
faddiv: Better name. Renamed.

src/Riok.Mapperly/Configuration/MembersMappingConfiguration.cs
faddiv: Created a test and implemented it.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/BuilderContext/IgnoredMembersBuilder.cs
faddiv: Moved it there.

src/Riok.Mapperly/Diagnostics/DiagnosticCollection.cs
faddiv: It worked. 🙂 
I think this also improves performance slightly.
Although the other duplications appeared again.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: It is not as easy as it seems. I left it for now.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: It worked and simplified this part.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
faddiv: Renamed

src/Riok.Mapperly/Descriptors/SimpleMappingBuilderContext.cs
faddiv: Applied during my modifications

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
faddiv: I applied this version.

faddiv:
I think I managed to fix the test failures, but if there are still some, I'll take care of them another day with the rest of feedbacks.

faddiv:
Also, I realized, the .NET 6.0 integration tests don't run.

docs/docs/configuration/reusing-mapping-configurations.md
faddiv: I re-read and reworked the documentation based on the current implementation.

faddiv:
Hopefully, I managed to fix all the build errors, except one:
Please add the enchantment label.

latonz (approved):

PR #1835, opened 2025-05-13
fix: nullable user implemented value type mappings in queryable projections

### LINKED ISSUE #1796: Projection with nullable reference types doesn't work after upgrade to 4.2.0
**Describe the bug**
After upgrading to 4.2.0 a projection that uses nullable value types and a custom (inlined) "Use" method doesn't seem to work correctly anymore.

**Declaration code**
```cs
[Mapper]
public static partial class TestMapper
{
    public static partial IQueryable<TestDto> ProjectToTestDto(this IQueryable<Test> query);

    [MapProperty(nameof(Test.Something), nameof(TestDto.Something), Use = nameof(MapSomething))]
    private static partial TestDto MapTest(Test test);

    [UserMapping(Default = false)]
    private static decimal MapSomething(decimal? something) => something ?? 0;

}

public class Test
{
    public decimal? Something { get; set; }
}

public class TestDto
{
    public decimal Something { get; set; }
}
```

Code from 4.2.0:
```csharp
new global::TestDto()
          {
              Something = x.Something != null ? x.Something.Value ?? 0 : default,
          }
```

Code from 4.1.1:
```csharp
new global::TestDto()
            {
                Something = x.Something ?? 0,
            }
```

Actual c# error:
* CS0019: Operator '??' cannot be applied to operands of type 'decimal' and 'int'

- Mapperly Version: [e.g. 1.0.1] 4.2.0
- .NET Version: 9.0.4
- Target Framework: net9.0
- Compiler Version:  '4.13.0-3.25167.3 (73eff2b5)'. Language version: 13.0.
- C# Language Version: 13.0
- IDE: VS 17.13.6
- OS: Windows 11

latonz:
Fixes #1796 
Fixes https://github.com/riok/mapperly/issues/1796#issuecomment-2877675738


test/Riok.Mapperly.Tests/_snapshots/QueryableProjectionUserImplementedTest.UserImplementedAttributedNullableValueTypeToNullableMemberTypeMismatch#Mapper.g.verified.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "0.0.1.0")]
+    public static partial global::System.Linq.IQueryable<global::B> Map(this global::System.Linq.IQueryable<global::A> query)
+    {
+#nullable disable
+        return global::System.Linq.Queryable.Select(
+            query,
+            x => new global::B()
+            {
+                Value = (decimal?)x.Value ?? 1,
```
latonz, line 14: This cast is not needed.

PR #1843, opened 2025-06-02
fix: prevents incorrect RMG020 warning

### LINKED ISSUE #1793: Incorrect warning RMG020 for nested types when custom method return types do not match target type exactly
**Describe the bug**
The warning "RMG020 The member .. on the mapping source type ...   is not mapped to any member on the mapping target type .." is incorrectly shown on nested mappings, when target types of the mapping target of a custom mapping method do not match exactly. The code is generated correctly and is using the custom method, but the warnings are shown as if the custom mapping method was not used. 

In my example (see below) the record-class `TargetWrapper` excepts an `IEnumerable<Target>` while the `CustomMapToTargetArray()` returns a `Target[]`. (This issue also only arises when one of the nested types is nullable and the other not, in my example `ICollection<Source?>` vs. `IEnumerable<Target>`) If the signature of the custom method is changed to `IEnumerable[] CustomMapToTargetArray(ICollection<Source?> sources)` the warning goes away. 

**Repro and Declaration code**
```csharp
namespace MapperlyIssue;

public record Source(string A, string B);
public record SourceWrapper(ICollection<Source?> Values);

public record Target(string A);
public record TargetWrapper(IEnumerable<Target> Values);

[Mapper]
public static partial class WarningIssue
{
    // Warning "RMG020: The member B on the mapping source type MapperlyIssue.Source is not mapped to 
    // any member on the mapping target type MapperlyIssue.Target" incorrectly shown here.
    public static partial TargetWrapper MapToTargetWrapper(SourceWrapper source);

    private static Target[] CustomMapToTargetArray(ICollection<Source?> sources)
        => sources.Where(s => s is not null).Select(s => s!.MapToTarget()).ToArray();
    
    [MapperIgnoreSource(nameof(Source.B))]
    private static partial Target MapToTarget(this Source source);
}
```

```csharp
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.2.0.0")]
        public static partial global::MapperlyIssue.TargetWrapper MapToTargetWrapper(global::MapperlyIssue.SourceWrapper source)
        {
            var target = new global::MapperlyIssue.TargetWrapper(CustomMapToTargetArray(source.Values));
            return target;
        }
```

* RMG020: The member B on the mapping source type MapperlyIssue.Source is not mapped to any member on the mapping target type MapperlyIssue.Target

- Mapperly Version: [4.2.0]
- Nullable reference types: [enabled]
- .NET Version: [NET 9.0.200]
- Target Framework: [net9.0]
- Compiler Version: [4.13.0-3.25072.11 (40a6eabe)], 
- C# Language Version: [e.g. 13.0]
- IDE: [Rider 2024.3.7]
- OS: [Windows 13.3]

**Additional context**

Andrei-hub11:
# Fix: prevents incorrect RMG020 warning

## Description

Fixes issue where RMG020 "The member .. on the mapping source type ... is not mapped to any member on the mapping target type ." warnings were incorrectly reported when user mappings with `MapperIgnoreSource` attributes were used in collection element mapping scenarios.

_For more details, see issue #1793._

**Solution:**
Modified `EnumerableMappingBuilder` to detect when user mappings exist for non-nullable element types and use `FindOrBuildLooseNullableMapping` instead of `FindOrBuildMapping`. This preserves user mapping configurations (like `MapperIgnoreSource` attributes) and prevents false positive diagnostics.

Fixes #1793

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
@@ -3,6 +3,7 @@ using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Descriptors.Enumerables;
 using Riok.Mapperly.Descriptors.Mappings;
 using Riok.Mapperly.Descriptors.Mappings.ExistingTarget;
+using Riok.Mapperly.Descriptors.Mappings.UserMappings;
 using Riok.Mapperly.Diagnostics;
 using Riok.Mapperly.Emit.Syntax;
 using Riok.Mapperly.Helpers;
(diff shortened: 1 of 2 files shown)


Andrei-hub11:
@latonz I tried applying the suggestion in [#1793 (comment)](https://github.com/riok/mapperly/issues/1793#issuecomment-xxx) — specifically ignoring IUserMapping in MappingBuilderContext.BuildDelegatedMapping. Unless I misunderstood something, the suggested fix didn’t resolve the issue. Please let me know if you'd like me to explore a different approach, make any changes, or if the current solution looks good to you.

latonz:
Thank you for your contribution! Why didn't the idea (ignoring `IUserMapping` in `MappingBuilderContext.BuildDelegatedMapping`) work?

Andrei-hub11:
@latonz I'm not sure, but I tried something like this:

````
var existingMapping = FindMapping(source, target);

if (existingMapping is IUserMapping)
{
    return null;
}
````

The early return happens when I'm debugging the test that reproduces the error, but the warning still appears.

latonz:
The easiest way to identify the source of a diagnostic is to set a breakpoint in `DiagnosticCollection.ReportDiagnostic`. To narrow it down, you can add a condition to the breakpoint that matches the specific `RMG` number you want to trace.

PR #1881, opened 2025-07-09
Add Default MappingTypeConversion

### LINKED ISSUE #1638: Disable explicit cast for widening conversions
Currently there's always an explicit cast when casting one numeric type to another. However, it's not needed for widening conversions! Consider the following example:

```csharp
public class DtoA
{
    public byte Count { get; set; }
}

public class DtoB
{
    public int Count { get; set; }
}

[Mapper]
{
    public static partial DtoB Map(DtoA a);
}
```

It produces this mapper:
```csharp
    {
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.1.1.0")]
        public static partial global::MapperlyExample.DtoB Map(global::MapperlyExample.DtoA a)
        {
            var target = new global::MapperlyExample.DtoB();
            target.Count = (int)a.Count;
            return target;
        }
    }
```

The cast is redundant. Some might also view the explicit cast as a code smell, since if you change the `DtoA` Count to be long:

```csharp
public class DtoA
{
    public long Count { get; set; }
}
```

the mapping code stays the same - but now you have potential data loss as it's a narrowing conversion!

Either:

* Change default behavior not to produce explicit casts in mappers on numeric types widening conversions
OR
* Introduce parameter to control that behavior i.e. `[Mapper(NoExplicitWideningCast = true)]`

zchung:
# Add Default MappingTypeConversion

## Description

Added a new flag for a Default MappingTypeConversion, this is a breaking change that will set the default configuration for a mapper to exclude the ExplicitCast Conversion, this will prevent accidental data loss for Explicit Casting. 

A lot of the changes are in the test cases, I had to "Explicitly" Add the flags to correctly test the new behavior on existing test cases to ensure they would work

Fixes #1638 

## Checklist

### DIFF
--- a/docs/docs/breaking-changes/5-0.md
+++ b/docs/docs/breaking-changes/5-0.md
@@ -1,0 +1,44 @@
+---
+sidebar_position: -5
+description: How to upgrade to Mapperly v5.0 and a list of all its breaking changes
+---
+
+# v5.0
+
+[![Release notes v5.0.0](https://img.shields.io/badge/Release_notes-v5.0-green?style=flat-square)](https://github.com/riok/mapperly/releases/tag/v5.0.0)
+[![NuGet v5.0.0](https://img.shields.io/badge/NuGet-v5.0-blue?style=flat-square)](https://www.nuget.org/packages/Riok.Mapperly/5.0.0)
+[![API Diff v5.0.0 / v4.3.0](https://img.shields.io/badge/API--Diff-v5.0_%2F_v4.0-yellow?style=flat-square)](https://www.fuget.org/packages/Riok.Mapperly/5.0.0/lib/netstandard2.0/diff/4.0.0/)
+
+## Migration guide from v4.3.0
+
+- Default enabled conversions no longer include explicit casts. Use `MappingConversionType.All` to get the previous behavior, or enable explicit casts explicitly with `MappingConversionType.All | MappingConversionType.ExplicitCast`.
+
+## Default enabled conversions exclude explicit casts
+
+Starting with v5.0, Mapperly introduces a new `MappingConversionType.Default` value that excludes explicit casts.
+This value is now used as the default for `MapperAttribute.EnabledConversions`.
+
+Previously, the default was `MappingConversionType.All`, which included all conversions including explicit casts.
+The new default behavior excludes explicit casts to be safer by default, as explicit casts can potentially throw runtime exceptions.
+
+### Migration options:
+
+1. **Recommended**: Review your mappings and explicitly enable `ExplicitCast` where needed:
+   ```csharp
+   [Mapper(EnabledConversions = MappingConversionType.ExplicitCast)]
+   public partial class MyMapper
+   {
+       public partial Target Map(Source source);
+   }
+   ```
+
+2. **Quick fix**: Use `MappingConversionType.All` to restore the previous behavior:
+   ```csharp
+   [Mapper(EnabledConversions = MappingConversionType.All)]
+   public partial class MyMapper
+   {
+       // Your existing mappings
+   }
+   ```
+
+This change makes Mapperly safer by default while still allowing explicit casts when explicitly requested.
--- a/test/Riok.Mapperly.IntegrationTests/Mapper/DerivedMapper.cs
+++ b/test/Riok.Mapperly.IntegrationTests/Mapper/DerivedMapper.cs
@@ -2,7 +2,7 @@ using Riok.Mapperly.Abstractions;
 
 namespace Riok.Mapperly.IntegrationTests.Mapper
 {
-    [Mapper]
+    [Mapper(EnabledConversions = MappingConversionType.ExplicitCast | MappingConversionType.ImplicitCast)]
     public partial class BaseMapper
     {
         public virtual partial long IntToLong(int value);
@@ -10,13 +10,13 @@ namespace Riok.Mapperly.IntegrationTests.Mapper
         public partial short IntToShort(int value);
     }
 
-    [Mapper]
+    [Mapper(EnabledConversions = MappingConversionType.ImplicitCast)]
     public partial class DerivedMapper : BaseMapper
     {
         public override partial long IntToLong(int value);
     }
 
-    [Mapper]
+    [Mapper(EnabledConversions = MappingConversionType.ExplicitCast | MappingConversionType.ImplicitCast)]
     public partial class DerivedMapper2 : BaseMapper
     {
         public sealed override partial long IntToLong(int value);
--- a/test/Riok.Mapperly.IntegrationTests/Mapper/StaticTestMapper.cs
+++ b/test/Riok.Mapperly.IntegrationTests/Mapper/StaticTestMapper.cs
@@ -6,7 +6,7 @@ using Riok.Mapperly.IntegrationTests.Models;
 
 namespace Riok.Mapperly.IntegrationTests.Mapper
 {
-    [Mapper(EnumMappingStrategy = EnumMappingStrategy.ByValue)]
+    [Mapper(EnumMappingStrategy = EnumMappingStrategy.ByValue, EnabledConversions = MappingConversionType.All)]
     public static partial class StaticTestMapper
     {
         [UserMapping(Default = true)]
--- a/test/Riok.Mapperly.IntegrationTests/Mapper/TestMapper.cs
+++ b/test/Riok.Mapperly.IntegrationTests/Mapper/TestMapper.cs
@@ -15,10 +15,11 @@ namespace Riok.Mapperly.IntegrationTests.Mapper
     [Mapper(
         IncludedMembers = MemberVisibility.All,
         IncludedConstructors = MemberVisibility.All,
-        EnumMappingStrategy = EnumMappingStrategy.ByValue
+        EnumMappingStrategy = EnumMappingStrategy.ByValue,
+        EnabledConversions = MappingConversionType.All
     )]
 #else
-    [Mapper(EnumMappingStrategy = EnumMappingStrategy.ByValue)]
+    [Mapper(EnumMappingStrategy = EnumMappingStrategy.ByValue, EnabledConversions = MappingConversionType.All)]
 #endif
     public partial class TestMapper
     {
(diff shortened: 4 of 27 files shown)


docs/docs/breaking-changes/5-0.md
```diff
@@ -0,0 +1,44 @@
...
+
+# v5.0
+
+[![Release notes v5.0.0](https://img.shields.io/badge/Release_notes-v5.0-green?style=flat-square)](https://github.com/riok/mapperly/releases/tag/v5.0.0)
+[![NuGet v5.0.0](https://img.shields.io/badge/NuGet-v5.0-blue?style=flat-square)](https://www.nuget.org/packages/Riok.Mapperly/5.0.0)
+[![API Diff v5.0.0 / v4.3.0](https://img.shields.io/badge/API--Diff-v5.0_%2F_v4.0-yellow?style=flat-square)](https://www.fuget.org/packages/Riok.Mapperly/5.0.0/lib/netstandard2.0/diff/4.0.0/)
+
+## Migration guide from v4.3.0
+
+- Default enabled conversions no longer include explicit casts. Use `MappingConversionType.All` to get the previous behavior, or enable explicit casts explicitly with `MappingConversionType.All | MappingConversionType.ExplicitCast`.
```
latonz: `MappingConversionType.All` includes `MappingConversionType.ExplicitCast` already, doesn't it? Shouldn't it be `MappingConversionType.Default | MappingConversionType.ExplicitCast`?

test/Riok.Mapperly.IntegrationTests/Mapper/DerivedMapper.cs
```diff
@@ -2,21 +2,21 @@
 
 namespace Riok.Mapperly.IntegrationTests.Mapper
 {
-    [Mapper]
+    [Mapper(EnabledConversions = MappingConversionType.ExplicitCast | MappingConversionType.ImplicitCast)]
```
latonz: Wouldn't it be simpler to just use the old `MappingConversionType.All`?

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
```diff
@@ -30,7 +30,37 @@ public record TestSourceBuilderOptions(
     public static readonly TestSourceBuilderOptions Default = new();
     public static readonly TestSourceBuilderOptions AsStatic = new(Static: true);
     public static readonly TestSourceBuilderOptions WithDeepCloning = new(UseDeepCloning: true);
+    public static readonly TestSourceBuilderOptions WithDeepCloningAndExplicitCast = new TestSourceBuilderOptions(
```
latonz: Instead of adding all of these entries individually, I’d suggest using `TestSourceBuilderOptions.WithDeepCloning with { ... }`. We should keep this focused on the most common options to avoid cluttering the list.

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
```diff
@@ -190,7 +190,7 @@ public Task WithClassBaseTypeShouldWork()
             using System.Collections.Generic;
             using Riok.Mapperly.Abstractions;
 
-            [Mapper]
+            [Mapper(EnabledConversions = MappingConversionType.ExplicitCast)]
```
latonz: Rather than enabling the explicit cast everywhere, I think it’s better to update the test, at least in cases where it’s not specifically verifying the explicit cast behavior.

latonz (commented):
Thank you for your contribution and efforts!

Since this introduces a breaking change, we'll postpone merging it until fall, when we plan to include it in our next scheduled breaking-change release (together with other breaking changes).

docs/docs/breaking-changes/5-0.md
zchung: oh you are right. Actually I think I should just remove the `MappingConversionType.All` part from that guide because 
`MappingConversionType.Default | MappingConversionType.ExplicitCast` also means the same but `MappingConversionType.ExplicitCast` on its own satisfies the sentence, what do you think?

test/Riok.Mapperly.IntegrationTests/Mapper/DerivedMapper.cs
zchung: I'm glad you asked this, I originally had All on everything but I felt like that was a lazy short cut since in a testing scenario we already have the All condition tested in many different places but for the specific Enabled conversions now that the default excludes Explicit cast it seemed prudent to have test cases that cover these scenarios now. I can change it to All if you like though

docs/docs/breaking-changes/5-0.md
latonz: Only setting `MappingConversionType.ExplicitCast` would disable all other conversions, IMO `MappingConversionType.Default | MappingConversionType.ExplicitCast` should be correct here.

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
zchung: I see, my intention here is because there are several test cases with the same scenario I wanted to avoid repeating the with {...} multiple times, this seemed a little cleaner.

test/Riok.Mapperly.IntegrationTests/Mapper/DerivedMapper.cs
latonz: I would modify the code to use `All` only where necessary (where the explicit cast should be tested). In all other cases, I would modify the test case to avoid explicit casts.

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
zchung: Agreed, I originally didn't update the test to keep the original design intact but if you're okay with changing the test then that is probably cleaner.

zchung:
@latonz I've updated the unit tests now and removed the explicit casting where applicable and used All for any other cases. I've also removed the extra parameters in the TestOptions class and simplified it to have options for the All Conversion Type only. This should hopefully simplify the test cases.

test/Riok.Mapperly.IntegrationTests/Mapper/DerivedMapper.cs
```diff
@@ -20,7 +18,5 @@ public partial class DerivedMapper : BaseMapper
     public partial class DerivedMapper2 : BaseMapper
     {
         public sealed override partial long IntToLong(int value);
-
-        public new partial short IntToShort(int value);
```
latonz, line 24: Instead of just removing the `new` override test, could you add another method testing the `new override?`

test/Riok.Mapperly.IntegrationTests/Mapper/StaticTestMapper.cs
```diff
@@ -16,8 +16,6 @@ public static partial class StaticTestMapper
 
         public static partial long ImplicitCastInt(int value);
 
-        public static partial int ExplicitCastInt(uint value);
```
latonz, line 19: Here, the explicit cast was tested explicitly. Instead of removing it, I'd enable the explicit cast (using `.All`) for this mapper.

test/Riok.Mapperly.IntegrationTests/Mapper/TestMapper.cs
```diff
@@ -41,8 +41,6 @@ public TestMapper()
 
         public partial long ImplicitCastInt(int value);
 
-        public partial int ExplicitCastInt(uint value);
```
latonz, line 44: Here, the explicit cast was tested explicitly. Instead of removing it, I'd enable the explicit cast (using `.All`) for this mapper.

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
```diff
@@ -30,7 +30,16 @@ public record TestSourceBuilderOptions(
     public static readonly TestSourceBuilderOptions Default = new();
     public static readonly TestSourceBuilderOptions AsStatic = new(Static: true);
     public static readonly TestSourceBuilderOptions WithDeepCloning = new(UseDeepCloning: true);
+    public static readonly TestSourceBuilderOptions WithDeepCloningAndConversionTypeAll = new TestSourceBuilderOptions(
+        UseDeepCloning: true,
+        EnabledConversions: MappingConversionType.All
+    );
+    public static readonly TestSourceBuilderOptions WithAll = new TestSourceBuilderOptions(EnabledConversions: MappingConversionType.All);
```
latonz: A more descriptive name might be `AllConversions`.

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
```diff
@@ -30,7 +30,16 @@ public record TestSourceBuilderOptions(
     public static readonly TestSourceBuilderOptions Default = new();
     public static readonly TestSourceBuilderOptions AsStatic = new(Static: true);
     public static readonly TestSourceBuilderOptions WithDeepCloning = new(UseDeepCloning: true);
+    public static readonly TestSourceBuilderOptions WithDeepCloningAndConversionTypeAll = new TestSourceBuilderOptions(
+        UseDeepCloning: true,
+        EnabledConversions: MappingConversionType.All
+    );
```
latonz: A simpler and more descriptive name might be `AllConversionsWithDeepCloning`

test/Riok.Mapperly.Tests/TestSourceBuilderOptions.cs
```diff
@@ -30,7 +30,16 @@ public record TestSourceBuilderOptions(
...
+    public static readonly TestSourceBuilderOptions WithAll = new TestSourceBuilderOptions(EnabledConversions: MappingConversionType.All);
     public static readonly TestSourceBuilderOptions WithReferenceHandling = new(UseReferenceHandling: true);
+    public static readonly TestSourceBuilderOptions WithReferenceHandlingAndConversionTypeAll = new(
```
latonz: A simpler and more descriptive name might be `AllConversionsWithReferenceHandling`

latonz (changes requested):
Thank you for the updates! I have a couple of small nitpicks.

latonz (approved):

PR #1886, opened 2025-07-15
Fix syntax errors in tests

faddiv:
# Fix syntax errors in tests

## Description

I found several syntax errors in the test input sources. Some are made by me. I would like to push a fix.

I found this while investigating `CreateNameOfMemberPath` and whether it could be used for the full path referencing of other methods. Interestingly, there were too many cases when the fallback code ran, which led to this realization. The fallback still runs, but in cases where the error is intentional.

## Checklist


latonz (approved):
Thank you for this contribution! 🚀

PR #1890, opened 2025-07-22
Reference External mappings

### LINKED ISSUE #1856: Reference external mappings
Several APIs in Mapperly allow referencing other mappings. Currently, only mappings within the same mapper class are supported. This functionality should be extended to support referencing mappings from any mapping class using a fully qualified name, with support for "[full nameof](https://mapperly.riok.app/docs/configuration/flattening/#full-nameof)" (e.g., `MyNamespace.MyMappingClass.MyMappingMethod` or `nameof(@MyNamespaceMyMappingClass.MyMappingMethod)`).

This enhancement should apply to the following attributes:
* `MapValueAttribute.Use`
* `MapPropertyAttribute.Use`
* `MapPropertyFromSourceAttribute.Use`
* `IncludeMappingConfigurationAttribute.Name`

Rel. https://github.com/riok/mapperly/issues/1855 
Follows up on: https://github.com/riok/mapperly/issues/513 and https://github.com/riok/mapperly/pull/1833

faddiv:
# Reference External mappings

## Description

Implements external mapping for static methods and instance methods through field reference.

Fixes #1856

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
+++ b/src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
@@ -1,4 +1,3 @@
-using System.Diagnostics;
 using Microsoft.CodeAnalysis;
 using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Configuration;


src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
```diff
@@ -259,6 +258,96 @@ private static IMemberPathConfiguration CreateNameOfMemberPath(InvocationExpress
         return new SymbolMemberPath(memberPath.ToImmutableEquatableArray());
     }
 
+    private static MethodReferenceConfiguration CreateMethodReferenceConfiguration(
```
faddiv: My first thought was to reuse the `IMemberPathConfiguration` for this purpose, but in the end, I decided it is a better approach to keep it separated. This config holds the name and the type where it is declared.
To keep the existing features intact, I needed to grab the containing class symbol and see if it the Mapper class that is processed. This way the target type is null, when it is not needed.

faddiv: The current implementation doesn't check the @ symbol. Should it do for consistent behavior?

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -79,13 +79,13 @@ MapperConfiguration defaultMapperConfiguration
...
         ReserveMethodNames();
         ExtractUserMappings();
+        ExtractExternalMappings();
 
         // ExtractObjectFactories needs to be called after ExtractUserMappings due to configuring mapperDescriptor.Static
         var objectFactories = ExtractObjectFactories();
         var constructorFactory = new InstanceConstructorFactory(objectFactories, _symbolAccessor, _unsafeAccessorContext);
         var formatProviders = ExtractFormatProviders();
         EnqueueUserMappings(constructorFactory, formatProviders);
-        ExtractExternalMappings();
```
faddiv: I pulled up the `ExtractExternalMappings` after the `ExtractUserMappings`. This is needed for the IncludeConfiguration, and it seems it doesn't break anything.

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
```diff
@@ -32,6 +32,29 @@ public static IEnumerable<IUserMapping> ExtractExternalMappings(SimpleMappingBui
         return staticExternalMappers.Concat(externalInstanceMappers);
     }
 
+    public static IEnumerable<(string Name, IUserMapping Mapping)> ExtractExternalNamedMappings(
+        SimpleMappingBuilderContext ctx,
+        INamedTypeSymbol mapperSymbol
```
faddiv, line 38: I created a new ExtractExternalNamedMappings which focuses specifically the external uses in the file.

faddiv, line 38: Currently, it only works on the static usings, but I think I could make it work on the instance level mappings later with the help of the `UseMapper` attribute. WDYT?

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -60,7 +60,7 @@ internal static bool IsDelegate(this ISymbol symbol)
     internal static int GetInheritanceLevel(this ITypeSymbol symbol)
     {
         var level = 0;
-        while (symbol.BaseType != null)
+        while (symbol.BaseType is not null && symbol.BaseType.SpecialType != SpecialType.System_Object)
         {
```
faddiv, line 64: I found these and fixed them while I worked on a similar method in this class. The problem is that the object type's parent is itself; therefore, this runs into an infinite loop. If I remember correctly, I saw a report about Visual Studio crash. Maybe it is caused by this. If it is true, then a separate PR could be made, fixed, and unit tested.

src/Riok.Mapperly/Helpers/ThrowHelpers.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace Riok.Mapperly.Helpers;
+
+public static class ThrowHelpers
+{
+    [StackTraceHidden]
+    public static void ThrowIfNull(
```
faddiv, line 10: This is the replacement for the `ArgumentNullException.ThrowIfNull` since it is not available in .NET Standard, but it simplifies some code.

test/Riok.Mapperly.IntegrationTests/_snapshots/StaticMapperTest.SnapshotGeneratedSource_NET6_0.verified.cs
```diff
@@ -776,12 +776,12 @@ uint x when targetType.IsAssignableFrom(typeof(int)) => ExplicitCastInt(x),
...
                 global::System.Collections.Generic.IEnumerable<global::Riok.Mapperly.IntegrationTests.Models.TestObject> x when targetType.IsAssignableFrom(typeof(global::System.Collections.Generic.IEnumerable<global::Riok.Mapperly.IntegrationTests.Dto.TestObjectDto>)) => MapAllDtos(x),
+                global::Riok.Mapperly.IntegrationTests.Models.TestObject x when targetType.IsAssignableFrom(typeof(global::Riok.Mapperly.IntegrationTests.Dto.TestObjectDto)) => MapToDto(x),
+                global::Riok.Mapperly.IntegrationTests.Dto.TestObjectDto x when targetType.IsAssignableFrom(typeof(global::Riok.Mapperly.IntegrationTests.Models.TestObject)) => MapFromDto(x),
                 object x when targetType.IsAssignableFrom(typeof(object)) => DerivedTypes(x),
+                global::Riok.Mapperly.IntegrationTests.Models.ConstantValuesObject x when targetType.IsAssignableFrom(typeof(global::Riok.Mapperly.IntegrationTests.Dto.ConstantValuesDto)) => MapConstantValues(x),
                 _ => throw new global::System.ArgumentException($"Cannot map {source.GetType()} to {targetType} as there is no known type mapping", nameof(source)),
```
faddiv, line 785: For some reason, they swapped places. I don't know if this is a problem.

test/Riok.Mapperly.Tests/Mapping/ReferenceExternalMappingsTests.cs
```diff
@@ -0,0 +1,91 @@
+﻿namespace Riok.Mapperly.Tests.Mapping;
+
+public class ReferenceExternalMappingsTests
```
faddiv: Should these tests go into a separate test class as I did now, or should they be spread amongst the respective test classes? I plan to make additional tests, but now I just created ones that test the basic functionality.

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
```diff
@@ -172,6 +173,7 @@ private static TData Create<TData>(
             ),
             _ when arg.IsNull => null,
             _ when targetType == typeof(IMemberPathConfiguration) => CreateMemberPath(arg, syntax, symbolAccessor),
```
faddiv, line 177: On the CreateMemberPath, I found a potential bug:
If the `nameof` contains only a single class type without the property path, then it creates an empty `StringMemberPath`, which later causes an exception. Are there any bug reports on this issue? It should be fixed.

test/Riok.Mapperly.Tests/Mapping/ReferenceExternalMappingsTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    [Fact]
+    public Task MapValueUseOnInstanceSupportsExternalMappings()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+            OtherMapper mapper = new();
+
+            [MapValue("Value", Use = nameof(mapper.NewValue))]
+            internal partial B Map(A source);
```
faddiv: Hehh, managed the external mapping this way. I think this has potential when someone uses instance-level mapping.

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
```diff
@@ -259,6 +258,97 @@ private static IMemberPathConfiguration CreateNameOfMemberPath(InvocationExpress
         return new SymbolMemberPath(memberPath.ToImmutableEquatableArray());
     }
 
+    private static MethodReferenceConfiguration CreateMethodReferenceConfiguration(
```
faddiv: Also, I will look through and refine the implementation, but for now, it works.

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
faddiv: Solved, and not even the UseMapper is needed.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
faddiv: It doesn't fix anything. I wonder why it behaves differently for me.

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
latonz: AFAIK there are no bug reports regarding this. Feel free to open an issue and/or a PR fixing this.

src/Riok.Mapperly/Configuration/MethodReferenceConfiguration.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Helpers;
+
+namespace Riok.Mapperly.Configuration;
+
+public record class MethodReferenceConfiguration(string Name, INamedTypeSymbol? TargetType = null, ISymbol? TargetMember = null)
```
latonz: The class keyword is not needed, is it?

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
```diff
@@ -259,6 +250,70 @@ private static IMemberPathConfiguration CreateNameOfMemberPath(InvocationExpress
...
+        TypedConstant arg,
+        AttributeArgumentSyntax? syntax,
+        SymbolAccessor? symbolAccessor
+    )
+    {
+        ThrowHelpers.ThrowIfNull(symbolAccessor);
+
+        if (arg.Kind != TypedConstantKind.Primitive)
+        {
+            throw new InvalidOperationException($"Cannot create {nameof(StringMemberPath)} from {arg.Kind}");
```
latonz: This is not a `StringMemberPath` is it?

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -265,4 +265,28 @@ internal static bool HasKeyword(this ITypeSymbol typeSymbol, [NotNullWhen(true)]
...
     }
+
+    internal static bool ExtendsType(this ITypeSymbol derivedType, ITypeSymbol baseType)
+    {
+        var type = derivedType;
+        if (baseType.TypeKind == TypeKind.Interface)
+        {
+            return type.AllInterfaces.Any(x => SymbolEqualityComparer.Default.Equals(x, baseType));
+        }
+        while (type is not null)
```
latonz: Nit: add empty line before blocks.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -265,4 +265,28 @@ internal static bool HasKeyword(this ITypeSymbol typeSymbol, [NotNullWhen(true)]
...
+            if (SymbolEqualityComparer.Default.Equals(baseType, type))
+            {
+                return true;
+            }
+
+            type = type.BaseType;
+            if (type is not null && type.BaseType is not null && type.SpecialType == SpecialType.System_Object)
+            {
+                Debugger.Break();
+            }
```
latonz: Is this really needed or a leftover?

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -265,4 +265,28 @@ internal static bool HasKeyword(this ITypeSymbol typeSymbol, [NotNullWhen(true)]
         keywordName = null;
         return false;
     }
+
+    internal static bool ExtendsType(this ITypeSymbol derivedType, ITypeSymbol baseType)
```
latonz: As we already have `Implements` a better matching name would be `Extends`.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -265,4 +265,28 @@ internal static bool HasKeyword(this ITypeSymbol typeSymbol, [NotNullWhen(true)]
...
+    {
+        var type = derivedType;
+        if (baseType.TypeKind == TypeKind.Interface)
+        {
+            return type.AllInterfaces.Any(x => SymbolEqualityComparer.Default.Equals(x, baseType));
```
latonz: We could use the `Implements` method here, couldn't we?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
```diff
@@ -156,9 +157,18 @@ private static bool TryBuildMethodProvidedSourceValue(
...
-            sourceValue = new MethodProvidedSourceValue(method.Name);
+            string? targetName = null;
+            if (methodReference.TargetMember is not null)
+            {
+                targetName = methodReference.TargetMember.Name;
+            }
+            else if (methodReference.TargetType is not null)
+            {
+                targetName = methodReference.TargetType.FullyQualifiedIdentifierName();
+            }
```
latonz: add a NL before and after blocks.

src/Riok.Mapperly/Helpers/ThrowHelpers.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace Riok.Mapperly.Helpers;
+
+public static class ThrowHelpers
+{
+    [StackTraceHidden]
```
latonz, line 9: This is not available in .net standard, have you tested this (it may get polyfilled)?

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
```diff
@@ -259,6 +250,70 @@ private static IMemberPathConfiguration CreateNameOfMemberPath(InvocationExpress
         return new SymbolMemberPath(memberPath.ToImmutableEquatableArray());
     }
 
+    private static MethodReferenceConfiguration CreateMethodReferenceConfiguration(
+        TypedConstant arg,
+        AttributeArgumentSyntax? syntax,
+        SymbolAccessor? symbolAccessor
```
latonz, line 258: why is `symbolAccessor` nullable if it is just checked for not-null one line later?

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
latonz: IMO whenever we don't use the default c# nameof logic, we should ensure `@` prefix is used. I'd implement it directly in `TryGetFullNameOfSyntax`.

test/Riok.Mapperly.Tests/Mapping/ReferenceExternalMappingsTests.cs
latonz: I think it makes more sense to put these tests next to the tests of the feature these are testing.

test/Riok.Mapperly.Tests/Mapping/ReferenceExternalMappingsTests.cs
```diff
@@ -0,0 +1,228 @@
+﻿namespace Riok.Mapperly.Tests.Mapping;
+
+public class ReferenceExternalMappingsTests
+{
+    [Fact]
+    public Task MapValueUseOnStaticSupportsExternalMappings()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+            [MapValue("Value", Use = nameof(OtherMapper.NewValue))]
```
latonz: IMO these should use fullnameof with the `@` prefix. Otherwise we override the c# behaviour of nameof....

src/Riok.Mapperly/Symbols/MapperMethodReference.cs
```diff
@@ -0,0 +1,5 @@
+﻿using Microsoft.CodeAnalysis;
+
+namespace Riok.Mapperly.Symbols;
+
+public record MapperMethodReference(IMethodSymbol Method, INamedTypeSymbol? TargetType, ISymbol? TargetMember);
```
latonz: Add a comment explaining when which field is set?
Did I get this correctly?
* Neither `TargetType` and `TargetMember` is set if a method of the current mapper is referenced
* `TargetType` is set if a method of an external type is referenced.
* `TargetMember` is set if an instance method of a local member is referenced.

latonz: Why this intermediate step and not resolving `MethodProvidedSourceValue` directly?

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -477,7 +478,25 @@ private bool TryFindPath(ITypeSymbol type, StringMemberPath path, bool ignoreCas
...
     }
 
-    public IOperation? GetOperation(SyntaxNode node) => compilationContext.GetSemanticModel(node.SyntaxTree)?.GetOperation(node);
+    public TOperation? GetOperation<TOperation>(SyntaxNode node)
+        where TOperation : class, IOperation => compilationContext.GetSemanticModel(node.SyntaxTree)?.GetOperation(node) as TOperation;
+
+    public ITypeSymbol? GetContainingTypeSymbol(SyntaxNode? node)
+    {
+        var classNode = node;
+        while (classNode is not (null or BaseTypeDeclarationSyntax))
```
latonz: IMO `classNode is not null and not BaseTypeDeclarationSyntax` would be easier to read here.

latonz: Why not using `node` directly?

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
latonz: This change now attempts to extract all external mappings upfront.  
Previously, we resolved referenced named mappings with `Use =` only when they were actually used,  
to avoid unnecessary work.  
Could you clarify the intention behind this change?

latonz (changes requested):

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
faddiv: I followed the patterns of other methods in this regard. Looking through this class, it seems the null check can be extracted to the top level. Should I do it in this PR?

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
faddiv: Yep. Fixed.

src/Riok.Mapperly/Configuration/MethodReferenceConfiguration.cs
faddiv: Nope. Removed.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
faddiv: Added. BTW, I'm sometimes afraid to format the code, as I would like to keep unrelated changes to a minimum, and other parts are changed by Rider. The reason is mainly the too-long lines. Is there a rule about line length on this project?

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
faddiv: Fixed

faddiv: The nullability check was messed up in the first version of this method, but now it is ok, so I simplified it.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
faddiv: Changed

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
faddiv: This is unnecessary here anyway, so I removed it.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
faddiv: This becomes obsolete.

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
faddiv: Leftover. Removed.
For a while, I experienced an interesting issue: The base type of the object was itself, and this method went into an infinite loop. It is now gone.

src/Riok.Mapperly/Helpers/ThrowHelpers.cs
faddiv: Yes, it is polyfilled. I'm not sure if this is very useful here, so I can remove it if you wish.

src/Riok.Mapperly/Symbols/MapperMethodReference.cs
faddiv: Added documentation comment.
1. The `TargetType` and `TargetMember` are not set if the method is on the mapper or its base class.
2. If the first is not true, then the second is always set.
3. `TargetMember` is set if the method is an instance-level method, and points to the field or property that holds a reference to the target type.

faddiv: I kept the original structure of the program, which kept the resolution and creation separate. I can merge it into one, if you wish.

test/Riok.Mapperly.Tests/Mapping/ReferenceExternalMappingsTests.cs
faddiv: Ok. I will move them when everything is ok and I create the integration tests and docs too. Until then, this conversation should remain open.

test/Riok.Mapperly.Tests/Mapping/ReferenceExternalMappingsTests.cs
faddiv: Changed the behavior and tests according to the feedback previously.

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
faddiv: Unfortunately, for IncludeMapping, it didn't work if I left these later. I think it was not necessary for the other use cases.
But I will reevaluate if this could be done on-demand.

test/Riok.Mapperly.IntegrationTests/_snapshots/StaticMapperTest.SnapshotGeneratedSource_NET6_0.verified.cs
latonz: I think this actually is a problem, as the line before this already matches everything, doesn't it? But this is already resolved?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
latonz: We use csharpier. It should be configured with `printWidth: 140`.

src/Riok.Mapperly/Helpers/ThrowHelpers.cs
latonz: Then I'd keep it for now.

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
latonz: Yes, I'd do it here.

src/Riok.Mapperly/Symbols/MapperMethodReference.cs
latonz: I'd merge it...

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
faddiv: I take it back. It turns out, it is not possible. There is one place where the symbol accessor can't be used. `MapperGenerator.BuildDefaults`

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
faddiv: I see, I hope you don't mind if I put a `max_line_length` into .editorconfig.  It helps my editor.

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
faddiv: I named this to TryGetFullNameOfSyntax, but it was misleading.
I renamed it to TryGetNameOfSyntax and used it for that. Besides that, I created a method `IsFullNameOfSyntax` that can check the @ prefix without generating the full name string. I replaced this check in the old method also, and the ToFullString is called only when the fallback version runs. (This happens in case of invalid paths.)

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
faddiv: As I see the linter minds it, so I reverted.

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
faddiv: Ok, I looked into it again. Only, the MapValue works separately.
For `ExtractExternalNamedMappings()` basically the same is needed as `ExtractUserMappings()`. This means extracting all the user mappings and cache with their name. 
For now, I separated the entire thing from the external unnamed mappings, so the two things don't get mixed up. I'm still investigating if it can be done during ExtractUserMappings.

faddiv: It can't be mixed up. Both extract only the necessary UserMapping information and cache it.
These Usage values are always applied, so I doubt that unnecessary work happens in this regard.
However, I see unnecessary work on attribute access, as it is read twice and with reflection. In my opinion, that could be cached in some way.

src/Riok.Mapperly/Symbols/MapperMethodReference.cs
faddiv: I merged this

src/Riok.Mapperly/Helpers/SymbolExtensions.cs
```diff
@@ -265,4 +265,19 @@ internal static bool HasKeyword(this ITypeSymbol typeSymbol, [NotNullWhen(true)]
         keywordName = null;
         return false;
     }
+
+    internal static bool Extends(this ITypeSymbol? derivedType, ITypeSymbol baseType)
+    {
+        while (derivedType is not null)
+        {
+            if (SymbolEqualityComparer.Default.Equals(baseType, derivedType))
```
faddiv: For some reason, this is needed, since sometimes the same symbol type is not the same reference, but still points exactly to the same type.

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
faddiv: I created one: https://github.com/riok/mapperly/issues/1906

test/Riok.Mapperly.IntegrationTests/_snapshots/StaticMapperTest.SnapshotGeneratedSource_NET6_0.verified.cs
faddiv: It is already resolved.

test/Riok.Mapperly.Tests/Mapping/ReferenceExternalMappingsTests.cs
faddiv: I moved these now.

src/Riok.Mapperly/Symbols/MapperMethodReference.cs
faddiv: Merge is done

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
faddiv: I gave another thought to it. It would be possible if the extraction did not depend on a `SimpleMappingBuilderContext`, since the information is needed during its creation. So I see only this as a possible solution at the moment.

faddiv: I still need to rethink this. It interferes with the UseMapper.

faddiv: The interference with UseMapper is fixed, but I still see no other way than to pre-collect these. Where the mapping is selected, no information that could be used to build these reference.
Example: `UseNamedMappingBuilder.TryBuildMapping`

faddiv:
I just noticed the fails today. I fixed them.
One of the problems is that `nameof` can't be used on fields. Some of the new integration tests had to be disabled below .NET 8.0.

faddiv:
Hi @latonz ,
I'm not sure if you are on holiday, but if you have time, could you check my PR. It has been done for a while.

src/Riok.Mapperly/Symbols/MapperMethodReference.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace Riok.Mapperly.Symbols;
+
+/// <summary>
+/// Describes a reference to a mapper method, including its symbol, declaring type can't be used directly,
+/// and the member providing the instance for instance methods.
+/// </summary>
+/// <param name="Method">The symbol of the referenced method.</param>
+/// <param name="TargetType">The declaring type if the method can't be used directly; otherwise, <c>null</c>.</param>
+/// <param name="TargetMember">The instance member for instance methods; otherwise, <c>null</c>.</param>
+public record MapperMethodReference(IMethodSymbol Method, INamedTypeSymbol? TargetType, ISymbol? TargetMember);
```
latonz: This class is now unused, isn't it?

test/Riok.Mapperly.IntegrationTests/Mapper/UseExternalInstanceMapper.cs
```diff
@@ -0,0 +1,33 @@
...
+    public partial class UseExternalInstanceMapper
+    {
+        [UseMapper]
+        private readonly MyOtherMapper _myOtherMapper = new();
+
+        private readonly ExternalMapperMethods _externalMapper = new();
+
+        public partial IdObjectDto Map(IdObject source);
+
+#if NET8_0_OR_GREATER
```
latonz, line 17: Why are all integration tests only .NET 8+? Also add tests without nameof.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyFromSourceTest.cs
```diff
@@ -212,4 +212,71 @@ public void NullableNestedPropertyWithMemberNameOfSource()
...
+    }
+
+    [Fact]
+    public Task ShouldSupportExternalMappingsOnInstanceProperty()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+            OtherMapper Mapper { get; } = new();
+
+            [MapPropertyFromSource(nameof(B.FullName), Use = nameof(@Mapper.ToFullName))]
```
latonz, line 267: Also add a test that uses string literal names instead of nameof. As far as I know, these are not supported yet, but I think they should be. Otherwise, the API feels inconsistent, allowing nameof but rejecting equivalent string literals is not how C# usually works.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyValueMethodTest.cs
```diff
@@ -437,4 +437,95 @@ public void InvalidMethodSignatureAsyncShouldDiagnostic()
...
+
+        return TestHelper.VerifyGenerator(source);
+    }
+
+    [Fact]
+    public Task ExternalMappingWorksWithFullNamespacePath()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            """
+            [MapValue("Value", Use = nameof(@Some.Namespace.OtherMapper.NewValue))]
```
latonz, line 513: Also include a test with string literal names instead of nameof (the same for other unit tests too). E.g. `[IncludeMappingConfiguration("Some.Namespace.OtherMapper.NewValue")]`

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyUseNamedMappingTest.cs
```diff
@@ -683,4 +683,73 @@ public record B(int Value);
             .HaveDiagnostic(DiagnosticDescriptors.ReferencedMappingNotFound, "The referenced mapping named MapValue was not found")
             .HaveAssertedAllDiagnostics();
     }
+
+    [Fact]
+    public Task ShouldSupportExternalMappingsOnStaticMapping()
```
latonz, line 688: Also add tests referencing methods on external mappers that don't exist or are not accessible.

latonz (changes requested):

test/Riok.Mapperly.IntegrationTests/Mapper/UseExternalInstanceMapper.cs
faddiv: The `nameof` on fields is not supported by .NET below 8, so I had to disable it.
Ok, I will add additional tests somewhere this week.

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -519,4 +547,66 @@ private IEnumerable<IMappableMember> GetAllAccessibleMappableMembersCore(ITypeSy
             .Select(x => MappableMember.Create(this, x))
             .WhereNotNull();
     }
+
+    public INamedTypeSymbol? GetTypeByMetadataName(string targetTypeName, SyntaxNode? contextNode)
```
faddiv: I created a method that resolves the class by name within the given context. In the tests I've seen, there are more than 400 possibilities to choose from in that simple case. Although the result is cached, I'm doubting it is the best approach. Should I allow only the full path in case of a string?

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyUseNamedMappingTest.cs
faddiv: Added tests for non-existing references.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyUseNamedMappingTest.cs
```diff
@@ -683,4 +683,227 @@ public record B(int Value);
...
+                }
+                """,
+            ]
+        );
+
+        return TestHelper.VerifyGenerator(source);
+    }
+
+    [Fact]
+    public Task ShouldSupportExternalMappingsOnStringWithFullNameSpecification()
```
faddiv: A little bit overgrown, the test cases, and it needs to be duplicated if I wish to test every case on every attribute (MapProperty, MapValue, MapPropertyFromSource, IncludeMapping). Are you ok with that, or should I use a different approach?

test/Riok.Mapperly.Tests/TestSourceBuilder.cs
```diff
@@ -59,6 +67,32 @@ public static string MapperWithBody([StringSyntax(StringSyntax.CSharp)] string b
         );
     }
 
+    public static string MapperWithBodyInBlockScopedNamespace(
+        [StringSyntax(StringSyntax.CSharp)] string body,
+        TestSourceBuilderOptions? options = null
```
faddiv: I added some new things to TestSourceBuilder, so I can test the mapper behavior when classes are in different namespaces.

faddiv:
Hi @latonz ,
I thought that the re-request review would be enough to request your input. (You requested this previously.)
If not, how can I notice you? Currently, I have the questions above, which are needed so I can continue the work.

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
latonz: Only supporting full name works perfectly fine (for external mapping methods), but we need to make sure it’s documented.

latonz:
Thanks for the updates, and sorry for the delay, this kept slipping down my to-do list. I’ll do a full review once you’ve addressed [this](https://github.com/riok/mapperly/pull/1890#discussion_r2359614256).

test/Riok.Mapperly.Tests/Mapping/NamedMappingTest.cs
```diff
@@ -135,4 +135,93 @@ public Task MapPropertyFromSourceWithUseUsesNamedMapping()
...
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypesInBlockScopedNamespace(
+            """
+            [MapProperty(nameof(A.StringValue), nameof(B.StringValue), Use = "OtherNamespace.OtherMapper.CustomModifyString")]
+            public partial B Map(A source);
+            """,
+            TestSourceBuilderOptions.Default,
+            "class A { public string StringValue { get; set; } }",
+            "class B { public string StringValue { get; set; } }"
+        );
```
faddiv, line 150: I was wondering if NamedMapping and external mapping can be combined. It just needed a slight enchantment, so I did it.

faddiv:
Hi @latonz,
I'm done with everything.

docs/docs/configuration/constant-generated-values.mdx
```diff
@@ -63,3 +65,155 @@ public partial CarDto Map(Car car);
 [NamedMapping("CustomGetSourceSystem")]
 string GetSourceSystem() => "C1";
 ```
+
+### External method reference
```
faddiv, line 69: I was wondering if these External mapping informations should go into its own section, since it is a little bit overgrown, and has too much duplication. Also, it is hard to find. I think I'm not the only one who thinks the same. I just found this: https://github.com/riok/mapperly/issues/1960

latonz: Sgtm, can be done in another MR.

latonz (dismissed):

latonz:

PR #1895, opened 2025-07-28
feat: support ref keyword in user defined methods for the target parameter

### LINKED ISSUE #1892: Copy `ref` keyword if UserMapping function has it when function is used
When using the existing object function plain arrays are overwritten, we want to merge the arrays, so we have a user defined mapping function for this. This needs the target to be passed as a reference to update the ref in the caller.

I'd like the source generator to be aware of the `ref` keyword in mappers, so reference to arrays can be updated (and other value types of course the need both the target and the source to determine the new value).

```csharp
class A
{
    public string[] arr { get; set; }
}

[Mapper(AllowNullPropertyAssignment = false, UseDeepCloning = true, RequiredMappingStrategy = RequiredMappingStrategy.Source)]
public static partial class Maps
{
    [MapProperty(nameof(A.arr), nameof(A.arr), Use = nameof(MapArray))] // Use it required otheriwse it will generate it's own
    public static partial void Merge([MappingTarget] this A target, A second);

    private static void MapArray([MappingTarget] ref string[] target, string[] second) => target = [.. target, .. second.Except(target)];
}

```

Generated code

```csharp
[global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.3.0.0")]
public static partial void Merge(this global::A configuration, global::A second)
{
    MapArray(configuration.arr, second.arr);
}
```

Desired generated code

```csharp
[global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.3.0.0")]
public static partial void Merge(this global::A configuration, global::A second)
{
    var temp = configuration.arr;
    MapArray(ref temp, second.arr);
    confguration.arr = temp;
}
```

I've tried to use Array.Resize but that also needs to update the reference in the caller and it can't. I tried to get it to inline the MapArray function but it won't do that, with or without the `ref` in the definition.

Other option is to let normal mapping methods (with a return value) also accept the target instance.

**Additional context**
This might help to implement this as well: https://github.com/riok/mapperly/issues/1416

EraYaN:
# support ref keyword in user defined methods for the target parameter

## Description

This will emit a ref keyword and supporting lines so pass by value types can actually be updated by the existing target methods (immutable records, array etc.) when the user defined methods has a ref keyword.

Fixes #1892

## Checklist

This is an initial attempt at solving this issue, does this agree with how the project would expect solving this issue?I has to extend the `MethodParameter` struct and pass it down a bit further to get the extra required context to the point where the Arguments are made. Not sure if there is some clean up that is possible to unify some of the other code paths.


latonz:
Thank you for this contribution, whats the reason for introducing the temporary field in the generated code?

EraYaN:
If the member access is a property, you can not pass it as a reference. Since it's really a function call. So it need to take the reference and then assign it back.

```csharp
var a = new ClassWithProperty();

// this
FunctionWithRefParam(ref a.Property);
// becomes this (kind of)
FunctionWithRefParam(ref a.get_Property());
```

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.Invocation.cs
```diff
@@ -170,6 +170,16 @@ public static ArgumentListSyntax ArgumentListWithoutIndention(IEnumerable<Argume
 
     private ArgumentListSyntax ArgumentList(IEnumerable<ExpressionSyntax> argSyntaxes) => ArgumentList(argSyntaxes.Select(Argument));
 
+    private ArgumentListSyntax ArgumentList(IEnumerable<MethodArgument> argSyntaxes) =>
+        ArgumentList(argSyntaxes.Select(ma => Argument(default, RefKindToken(ma.Parameter.RefKind), ma.Argument)));
```
latonz: Nit: this code base uses null in this case instead of default.

src/Riok.Mapperly/Symbols/MethodParameter.cs
```diff
@@ -3,15 +3,15 @@
 
 namespace Riok.Mapperly.Symbols;
 
-public readonly record struct MethodParameter(int Ordinal, string Name, ITypeSymbol Type)
+public readonly record struct MethodParameter(int Ordinal, string Name, ITypeSymbol Type, RefKind RefKind)
```
latonz: Since `RefKind` is only used rarely, I'd set a default value of `RefKind.None`.

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedExistingTargetMethodMapping.cs
```diff
@@ -32,6 +32,28 @@ public override IEnumerable<StatementSyntax> Build(TypeMappingBuildContext ctx,
         // we explicitly cast to be able to use the default interface implementation or explicit implementations
         if (Method.ReceiverType?.TypeKind != TypeKind.Interface)
         {
+            // if target parameter expects the ref keyword we need to add some extra lines so the reference is correctly stored back into the property.
+            if (targetParameter.RefKind == RefKind.Ref)
+            {
+                var indented = ctx.SyntaxFactory.AddIndentation();
+                yield return ctx.SyntaxFactory.Block(
```
latonz: Instead of wrapping the code in a block, we usually rely on unique variable names.  
You only used the block to prevent a name collision, correct?  
You can generate a unique name with `ctx.NameBuilder.New("tempRef")`.  
Consider using something like `Method.Name + "Ref"` as a name for better context.

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedExistingTargetMethodMapping.cs
```diff
@@ -32,6 +32,28 @@ public override IEnumerable<StatementSyntax> Build(TypeMappingBuildContext ctx,
         // we explicitly cast to be able to use the default interface implementation or explicit implementations
         if (Method.ReceiverType?.TypeKind != TypeKind.Interface)
         {
+            // if target parameter expects the ref keyword we need to add some extra lines so the reference is correctly stored back into the property.
+            if (targetParameter.RefKind == RefKind.Ref)
```
latonz, line 36: Why do we only support the target parameter? Couldn't there also be use-cases for the source parameter being a ref?

latonz: If we decide to also support source ref params, we also need to support `UserImplementedMethodMapping` and add diagnostics for queryable mappings etc.

test/Riok.Mapperly.Tests/Mapping/UserMethodTest.cs
```diff
@@ -50,6 +50,21 @@ static void MapList(List<int> src, List<string> dst) { }
         return TestHelper.VerifyGenerator(source);
     }
 
+    [Fact]
+    public Task InstanceMapperShouldUseStaticExistingTargetMethodWithRefKeyword()
```
latonz, line 54: Please also add an integration test.

latonz (changes requested):
Thank you for this contribution 😊 I added my feedback.

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedExistingTargetMethodMapping.cs
EraYaN: I couldn't figure out how to use the ref for the source param in any reasonable way, and it does make the code gen more complicated. So i figured that it might be worth not supporting.

latonz: I’m fine with supporting only the target parameter for now, as long as we document it clearly.

src/Riok.Mapperly/Emit/Syntax/SyntaxFactoryHelper.Invocation.cs
EraYaN: The current code calls the `Argument` function which is defined as:

```csharp
/// <summary>Creates a new ArgumentSyntax instance.</summary>
public static ArgumentSyntax Argument(ExpressionSyntax expression)
    => SyntaxFactory.Argument(default, default, expression);
```

Hence my use of `default` to not cause any differences (in this case it should be the same I believe). I'll adjust it.

EraYaN:
I've also rebased on latest `main`.

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
```diff
@@ -119,7 +119,8 @@ internal virtual void EnableReferenceHandling(INamedTypeSymbol iReferenceHandler
         ReferenceHandlerParameter ??= new MethodParameter(
             ReferenceHandlerParameterIndex,
             DefaultReferenceHandlerParameterName,
-            iReferenceHandlerType
+            iReferenceHandlerType,
+            RefKind.None
```
latonz: Explicit none parameter should not be required anymore...

src/Riok.Mapperly/Descriptors/Mappings/MethodMapping.cs
```diff
@@ -39,7 +39,7 @@ public abstract class MethodMapping : ITypeMapping
     protected MethodMapping(ITypeSymbol sourceType, ITypeSymbol targetType)
     {
         TargetType = targetType;
-        SourceParameter = new MethodParameter(SourceParameterIndex, DefaultSourceParameterName, sourceType);
+        SourceParameter = new MethodParameter(SourceParameterIndex, DefaultSourceParameterName, sourceType, RefKind.None);
```
latonz: Explicit none parameter should not be required anymore...

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedExistingTargetMethodMapping.cs
```diff
@@ -32,6 +32,24 @@ public override IEnumerable<StatementSyntax> Build(TypeMappingBuildContext ctx,
...
+            {
+                var targetRefVarName = Method.Name + "TargetRef";
```
latonz: My bad, `Method.Name` doesn’t really make sense in this context. I had confused it with the name of the target field. Since the target name isn’t easily accessible, I’d simply use
```csharp
var targetRefVarName = ctx.NameBuilder.New("targetRef");
```

test/Riok.Mapperly.IntegrationTests/UseUserMethodWithRefTest.cs
```diff
@@ -0,0 +1,30 @@
...
+            return Verifier.VerifyFile(path);
+        }
+
+        [Fact]
+        public void RunArrayMappingWithRef()
+        {
+            var modelTarget = new ArrayObject { IntArray = [10, 12] };
+            var modelSrc = new ArrayObject { IntArray = [11, 13] };
+            UseUserMethodWithRef.Merge(modelTarget, modelSrc);
+            modelTarget.IntArray.ShouldBe([10, 12, 11, 13]);
```
latonz: It probably would also make sure that the reference of the `modelTarget.IntArray` is the same before and after the mapper invocation.

test/Riok.Mapperly.IntegrationTests/UseUserMethodWithRefTest.cs
```diff
@@ -0,0 +1,30 @@
...
+using Shouldly;
+using VerifyXunit;
+using Xunit;
+
+namespace Riok.Mapperly.IntegrationTests
+{
+    public class UseUserMethodWithRefTest : BaseMapperTest
+    {
+        [Fact]
+        [VersionedSnapshot(Versions.NET6_0)]
```
latonz, line 14: Does this snapshot really need to be versioned? The generated code is the same for all .net versions, isn't it?

latonz (changes requested):

test/Riok.Mapperly.IntegrationTests/UseUserMethodWithRefTest.cs
EraYaN: But it is not the same in this case, that is kind of the point. A new array is allocated, so the reference is different.

latonz: You are absolutely correct, that's why we pass the pointer to the field and not the field itself.

EraYaN:
That last failing test, I hope it's fixed now, but no idea why it broke for .NET 4.8.

latonz:
The tests are failing due to the non-nullable not initialized array field in the integration tests: https://github.com/riok/mapperly/actions/runs/16592052238/job/46930700287?pr=1895#step:4:22

The integration test for the .net framework is failing since we generate slightly different code for the .net framework. Instead of `#nullable enable` we use `#nullable enable annotations` there. So your `[VersionedSnapshot(Versions.NET6_0)]` was correct, sorry about that. The test should pass if you add `[VersionedSnapshot(Versions.NET6_0)]`, copy the snapshot `UseUserMethodWithRefTest.SnapshotGeneratedSource.verified.cs` as `UseUserMethodWithRefTest.SnapshotGeneratedSource_NET6_0.verified.cs` and replace `#nullable enable` with `#nullable enable annotations` in `UseUserMethodWithRefTest.SnapshotGeneratedSource.verified.cs`.

EraYaN:
Alright that should then fix it, I mirrored the nullable annotation from `UseExternalMapperTest.SnapshotGeneratedSource_NET6_0.verified.cs`

EraYaN:
I understand the issue now, it needs both snapshots.

latonz (approved):

latonz:
Thank you for this great contribution! 🥳

PR #1921, opened 2025-08-15
Adding support for a "bool TryMap(.. out result)" pattern

### LINKED ISSUE #1920: Add bool TryMap(out result) pattern for runtime mapping
I originally opened a discussion:
https://github.com/riok/mapperly/discussions/1919

However, I thought I'd give it a go and attempt to contribute a solution. After reviewing the contribution guidelines, I am opening an issue so I can commit my changes and reference this issue.

The current capabilities allows me to create a partial method such as `object Map(object source, Type type)` and that generates a big switch expression that will, at runtime, go thru and "find" the matching method and execute and return the results.

If a match cannot be found, it throws an exception.

I would like to have a `bool TryMap(object source, Type type, ..., out object result)` pattern that will set the result in the out parameter and return true if the map was successful.

The reason I am doing this is that, in order to make migrating away from Automapper, I need to support an IMapper interface that is injected where multiple assemblies define their maps inside that assembly. I will update each assembly to use Mapperly, but to minimize all the places, I still want to use an IMapper that is injected.

I will scan the assemblies and find all those "profiles" or mapperly classes and when I call Map, I'll spin thru all those that are registered and return the first result that had a successful TryMap.

```csharp

public partial bool TryMap(object source, Type targetType, out object result);

private partial B MapToB(A source);
private partial D? MapToD(C source);
private partial int? MapToInt(string? source);


// Relevant new / adjusted API surface
// Expected generated code

public partial bool TryMap(object source, global::System.Type targetType, out object result)
{
	result = default;
	switch (source)
	{
		case global::A x when targetType.IsAssignableFrom(typeof(global::B)):
			result = MapToB(x);
			return true;
		default:
			return false;
	}
}

```

I considered just calling the Map method and catching the Exception, but the exception is an ArgumentException and I cannot be certain without checking the message, that the Exception is due to that particular mapper instance not being able to handle the mapping, or that the underlying method call generated an exception for whatever reason.

In addition, I generally prefer not to dictate program flow by catching exceptions (whenever I can).

**Additional context**
I haven't tried it, but an Update/TryUpdate pattern that works the same way might be a good addition in the future.

randellhodges:
## Description

Modeled after the existing process of generating a switch expression, this pattern looks for a method defined as `bool TryMap(object source, Type type, out result)` and similar and will generate method of switch statements that will set the result out parameter and return true if it was handled.

Fixes #1920

## Checklist


latonz:
See https://github.com/riok/mapperly/discussions/1919#discussioncomment-14138477

PR #1980, opened 2025-10-11
fix: handle single segment fullnameof

### LINKED ISSUE #1906: If MapProperty.Target is a nameof with single value mappery crashes
**Describe the bug**
If a map property looks like this: `[MapProperty(nameof(BananaDto.NamedBanana), nameof(Banana))]`
So the Target is a single identifier in a nameof expression, then no mapping is generated. (It causes an exception in mapperly.)

**Example**

[Mapper]
internal partial class InstanceBananaBoxMapper
{
    [MapProperty(nameof(BananaDto.NamedBanana), nameof(Banana))]
    public partial Banana MapBanana(BananaDto source);
}

- Mapperly Version: 4.2.1
- Nullable reference types: [e.g. enabled]
- .NET Version: .NET 8.0.412
- Target Framework: .net8.0
- Compiler Version: [e.g. 4.8.0-3.23524.11 (f43cd10b), obtain with `#error version` in any `.cs` file]
- C# Language Version: 12.0
- IDE: Rider 2025.1.4
- OS: Windows 11

**Additional context**
The source of the problem is that the `StringMemberPath` becomes empty in this case. The first identifier cut out, but if there is only one, nothing is left, which later causes an exception.

faddiv:
#  Handle empty path in RootName

## Description

This PR fixes the crash when the target is `fullname` but only has a single part. (f.e, Class only)
I'm unsure what it should do if it is a single part. For now, I made the simplest fix, but it doesn't really feel right.

I also created a test when the source property is a single part. This also behaves oddly:
- If the destination is a string, then it generates this: `destination.StringProperty = source.ToString()` Is this intentional?
- If it is not, then the map fails.

Fixes #1906

## Checklist

- [ ] Integration tests are added/updated (as applicable, especially if feature/bug depends on Roslyn or framework version in use)

### DIFF
--- a/src/Riok.Mapperly/Configuration/PropertyReferences/StringMemberPath.cs
+++ b/src/Riok.Mapperly/Configuration/PropertyReferences/StringMemberPath.cs
@@ -9,7 +9,7 @@ public readonly record struct StringMemberPath(ImmutableEquatableArray<string> P
     public StringMemberPath(IEnumerable<string> path)
         : this(path.ToImmutableEquatableArray()) { }
 
-    public string RootName => Path[0];
+    public string RootName => Path.FirstOrDefault() ?? string.Empty;
     public string FullName => string.Join(MemberPathConstants.MemberAccessSeparatorString, Path);
     public int PathCount => Path.Count;
     public IEnumerable<string> MemberNames => Path;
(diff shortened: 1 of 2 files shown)


latonz:
Thanks for the contribution, this is actually the correct handling.  
A single-part `fullname` should indeed be treated as a valid case, and your fix ensures we don’t crash while still covering that scenario correctly.

Regarding the `source.ToString()` behavior: yes, that’s intentional. When the destination is a string and the source is a non-string type, Mapperly converts it via `ToString()` by design. If it’s not a string, the mapping is expected to fail since there’s no implicit conversion.

PR #2011, opened 2025-10-30
feat: Add SnakeCase strategy to PropertyNameMappingStrategy

Kataane:
# Add SnakeCase Property Name Mapping Strategy

## Description

This PR adds support for `SnakeCase` as a new property name mapping strategy in Mapperly, enabling automatic mapping between PascalCase properties in C# source objects and snake_case properties in target objects.

### Motivation

In modern C# development, it's common to work with external systems or code generators that produce DTOs with snake_case property names (e.g., from JSON APIs, database schemas, or code generation tools). Currently, mapping between C# PascalCase properties and snake_case properties requires manual `[MapProperty]` attributes for each property, which is tedious and error-prone for large objects.

This feature is particularly valuable when:
- Working with auto-generated C# DTOs from external contracts that use snake_case naming
- Integrating with APIs that follow snake_case conventions (common in Python, Ruby, or REST APIs)
- The source contracts cannot be modified due to external dependencies or architectural constraints

### Changes Made

- Added `SnakeCase` enum value to `PropertyNameMappingStrategy` with XML documentation
- Implemented snake_case to PascalCase conversion for property matching in `MemberPathCandidateBuilder`
- Updated `MemberPathCandidateBuilder` with new overload supporting naming strategies
- Modified `MembersMappingBuilderContext` to use SnakeCase strategy when configured
- Updated PublicAPI snapshot to reflect new enum value
- Added comprehensive test case for SnakeCase property mapping validation

### Usage Example

```csharp
[Mapper(PropertyNameMappingStrategy = PropertyNameMappingStrategy.SnakeCase)]
public partial class MyMapper
{
    public partial TargetDto Map(SourceDto source);
}

// Maps automatically:
// source.FirstName -> target.first_name
// source.LastName -> target.last_name
```

This allows developers to map properties between PascalCase source objects and snake_case target objects (e.g., `FirstName` -> `first_name`) by simply setting `PropertyNameMappingStrategy = PropertyNameMappingStrategy.SnakeCase`.

Fixes # (issue - if applicable)

## Checklist

### DIFF
--- a/docs/docs/configuration/mapper.mdx
+++ b/docs/docs/configuration/mapper.mdx
@@ -209,29 +209,16 @@ or by setting the `IgnoreObsoleteMembersStrategy` option of the `MapperAttribute
 ### Property name mapping strategy
 
 By default, property and field names are matched using a case sensitive strategy.
-If all properties/fields differ only in casing, for example `ModelName` on the source
-and `modelName` on the target,
-the `MapperAttribute` can be used with the `PropertyNameMappingStrategy` option.
+If properties/fields differ in naming conventions, the `MapperAttribute` can be used with the `PropertyNameMappingStrategy` option.
 
-```csharp
-// highlight-start
-[Mapper(PropertyNameMappingStrategy = PropertyNameMappingStrategy.CaseInsensitive)]
-// highlight-end
-public partial class CarMapper
-{
-    public partial CarDto ToDto(Car car);
-}
+Available strategies:
 
-public class Car
-{
-    public string ModelName { get; set; }
-}
-
-public class CarDto
-{
-    public string modelName { get; set; }
-}
-```
+| Name            | Description                                                              |
+| --------------- | ------------------------------------------------------------------------ |
+| CaseSensitive   | Matches properties by their exact name (default)                         |
+| CaseInsensitive | Matches properties ignoring case differences                             |
+| SnakeCase       | Matches properties by converting to `snake_case` before comparison       |
+| UpperSnakeCase  | Matches properties by converting to `UPPER_SNAKE_CASE` before comparison |
 
 ### `null` values
 
--- a/src/Riok.Mapperly.Abstractions/PropertyNameMappingStrategy.cs
+++ b/src/Riok.Mapperly.Abstractions/PropertyNameMappingStrategy.cs
@@ -14,4 +14,16 @@ public enum PropertyNameMappingStrategy
     /// Matches a property by its name in case insensitive manner.
     /// </summary>
     CaseInsensitive,
+
+    /// <summary>
+    /// Matches a property by converting both source and target property names to snake_case before comparison.
+    /// For example, "FirstName" would match "first_name".
+    /// </summary>
+    SnakeCase,
+
+    /// <summary>
+    /// Matches a property by converting both source and target property names to SNAKE_CASE before comparison.
+    /// For example, "FirstName" would match "FIRST_NAME".
+    /// </summary>
+    UpperSnakeCase,
 }
--- a/src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
@@ -1,4 +1,6 @@
+using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Configuration.PropertyReferences;
+using Riok.Mapperly.Helpers;
 
 namespace Riok.Mapperly.Descriptors;
 
@@ -7,36 +9,61 @@ public static class MemberPathCandidateBuilder
     /// <summary>
     /// Maximum number of indices which are considered to compute the member path candidates.
     /// </summary>
-    private const int MaxPascalCaseIndices = 8;
+    private const int MaxPermutationIndices = 8;
 
     /// <summary>
-    /// Splits a name into pascal case chunks and joins them together in all possible combinations.
+    /// Splits a name into chunks and joins them together in all possible combinations.
     /// <example><c>"MyValueId"</c> leads to <c>[["MyValueId"], ["My", "ValueId"], ["MyValue", "Id"], ["My", "Value", "Id"]</c></example>
     /// </summary>
     /// <param name="name">The name to build candidates from.</param>
+    /// <param name="strategy">The naming strategy to use.</param>
     /// <returns>The joined member path groups.</returns>
-    public static IEnumerable<StringMemberPath> BuildMemberPathCandidates(string name)
+    public static IEnumerable<StringMemberPath> BuildMemberPathCandidates(string name, PropertyNameMappingStrategy strategy)
     {
-        if (name.Length == 0)
-            yield break;
+        if (string.IsNullOrEmpty(name))
+            return [];
 
-        // yield full string
-        // as a fast path (often member match by their exact name)
-        yield return new StringMemberPath([name]);
+        return BuildCandidates(name, strategy).Prepend(new StringMemberPath([name])).DistinctBy(x => x.FullName);
+    }
 
-        var indices = GetPascalCaseSplitIndices(name).Take(MaxPascalCaseIndices).ToArray();
-        if (indices.Length == 0)
-            yield break;
-
-        // try all permutations, skipping the first because the full string is already yielded
-        var permutationsCount = 1 << indices.Length;
-        for (var i = 1; i < permutationsCount; i++)
+    private static IEnumerable<StringMemberPath> BuildCandidates(string name, PropertyNameMappingStrategy strategy)
+    {
+        return strategy switch
         {
-            yield return new StringMemberPath(BuildPermutationParts(name, indices, i));
+            PropertyNameMappingStrategy.SnakeCase => BuildSnakeCaseCandidates(name, name.ToSnakeCase()),
+            PropertyNameMappingStrategy.UpperSnakeCase => BuildSnakeCaseCandidates(name, name.ToUpperSnakeCase()),
+            _ => BuildPermutations(name, char.IsUpper, skipSeparator: false),
+        };
+    }
+
+    private static IEnumerable<StringMemberPath> BuildSnakeCaseCandidates(string originalName, string snakeCaseName)
+    {
+        var snakeCasePermutations = BuildPermutations(snakeCaseName, static c => c == '_', skipSeparator: true);
+
+        // PascalCase Fallback
+        var pascalCaseSource = originalName.Contains('_', StringComparison.Ordinal) ? originalName.ToPascalCase() : originalName;
+        var pascalCasePermutations = BuildCandidates(pascalCaseSource, PropertyNameMappingStrategy.CaseSensitive);
+        return snakeCasePermutations.Concat(pascalCasePermutations);
+    }
+
+    private static IEnumerable<StringMemberPath> BuildPermutations(string name, Func<char, bool> isSeparator, bool skipSeparator)
+    {
+        var indices = GetSplitIndices(name, isSeparator).Take(MaxPermutationIndices).ToArray();
+
+        // try all permutations
+        var permutationsCount = 1 << indices.Length;
+        for (var i = 0; i < permutationsCount; i++)
+        {
+            yield return new StringMemberPath(BuildPermutationParts(name, indices, i, skipSeparator));
         }
     }
 
-    private static IEnumerable<string> BuildPermutationParts(string source, int[] splitIndices, int enabledSplitPositions)
+    private static IEnumerable<string> BuildPermutationParts(
+        string source,
+        int[] splitIndices,
+        int enabledSplitPositions,
+        bool skipSplitIndex
+    )
     {
         var lastSplitIndex = 0;
         var currentSplitPosition = 1;
@@ -45,7 +72,7 @@ public static class MemberPathCandidateBuilder
             if ((enabledSplitPositions & currentSplitPosition) == currentSplitPosition)
             {
                 yield return source.Substring(lastSplitIndex, splitIndex - lastSplitIndex);
-                lastSplitIndex = splitIndex;
+                lastSplitIndex = splitIndex + (skipSplitIndex ? 1 : 0);
             }
 
             currentSplitPosition <<= 1;
@@ -55,11 +82,11 @@ public static class MemberPathCandidateBuilder
             yield return source.Substring(lastSplitIndex);
     }
 
-    private static IEnumerable<int> GetPascalCaseSplitIndices(string str)
+    private static IEnumerable<int> GetSplitIndices(string str, Func<char, bool> isSeparator)
     {
         for (var i = 1; i < str.Length; i++)
         {
-            if (char.IsUpper(str[i]))
+            if (isSeparator(str[i]))
                 yield return i;
         }
     }
--- a/test/Riok.Mapperly.Tests/Descriptors/MemberPathCandidateBuilderTest.cs
+++ b/test/Riok.Mapperly.Tests/Descriptors/MemberPathCandidateBuilderTest.cs
@@ -1,3 +1,4 @@
+using Riok.Mapperly.Abstractions;
 using Riok.Mapperly.Descriptors;
 
 namespace Riok.Mapperly.Tests.Descriptors;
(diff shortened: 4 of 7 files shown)


src/Riok.Mapperly.Abstractions/PropertyNameMappingStrategy.cs
```diff
@@ -14,4 +14,10 @@ public enum PropertyNameMappingStrategy
     /// Matches a property by its name in case insensitive manner.
     /// </summary>
     CaseInsensitive,
+
+    /// <summary>
+    /// Matches a property by converting both source and target property names to snake_case before comparison.
+    /// For example, "FirstName" would match "first_name".
+    /// </summary>
+    SnakeCase,
```
latonz, line 22: IMO we should split this into `SnakeCase` (`snake_case`) and `UpperSnakeCase` (`SNAKE_CASE`). This also aligns with the `EnumNamingStrategy`.

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -36,6 +38,51 @@ public static IEnumerable<StringMemberPath> BuildMemberPathCandidates(string nam
...
+    /// Builds member path candidates based on the specified naming strategy.
+    /// For SnakeCase strategy, generates both PascalCase and snake_case versions to support bidirectional mapping.
+    /// Otherwise, delegates to the default BuildMemberPathCandidates method.
+    /// </summary>
+    /// <param name="name">The name to build candidates from.</param>
+    /// <param name="strategy">The property name mapping strategy.</param>
+    /// <returns>The joined member path groups.</returns>
+    public static IEnumerable<StringMemberPath> BuildMemberPathCandidates(string name, PropertyNameMappingStrategy strategy)
+    {
+        if (strategy == PropertyNameMappingStrategy.SnakeCase)
```
latonz: I'd move this into another method or even another class and use a switch expression on the strategy. And only expose the method with the strategy parameter.

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -36,6 +38,51 @@ public static IEnumerable<StringMemberPath> BuildMemberPathCandidates(string nam
...
+
+            // For bidirectional support, yield both versions:
+            // 1. If target is snake_case, convert to PascalCase to find PascalCase source
+            //    e.g., "first_name" -> "FirstName"
+            // 2. If target is PascalCase, convert to snake_case to find snake_case source
+            //    e.g., "FirstName" -> "first_name"
+            var pascalCase = name.ToPascalCase();
+
+            // Yield the converted version first as it's more likely to match
+            if (!string.Equals(name, pascalCase, StringComparison.Ordinal))
```
latonz: If snake case policy is explicitly match it should match before the pascal case match, since it is automatically set. Also we should match unflattened members.
In order of priority (for a member named my_value_id or MyValueId):
* my_value_id
* my.value_id
* my_value.id
* my.value.id
* MyValueId
* My.ValueId
* MyValue.Id
* My.Value.Id

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
```diff
@@ -246,6 +246,52 @@ public void WithPropertyNameMappingStrategyCaseInsensitive()
             );
     }
 
+    [Fact]
+    public void WithPropertyNameMappingStrategySnakeCaseOnTarget()
+    {
+        var source = TestSourceBuilder.MapperWithBodyAndTypes(
+            "partial B Map(A source);",
+            new TestSourceBuilderOptions { PropertyNameMappingStrategy = PropertyNameMappingStrategy.SnakeCase },
+            "class A { public string FirstName { get; set; } public string LastName { get; set; } }",
```
latonz, line 255: Add test cases for the following scenarios:
* exact matches (`snake_case` to `snake_case`)
* exact matches in PascalCase fallback (`MyMember` to `MyMember`)
* automatic flattening

latonz:
Please also update the documentation.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
Kataane: Hi. Thanks for the feedback, I will try to correct the comments soon.

I would like to clarify the point with PascalCase fallback.

What to do with this situation

```csharp
    "partial B Map(A source);",
    new TestSourceBuilderOptions { PropertyNameMappingStrategy = PropertyNameMappingStrategy.UpperSnakeCase },
    "class A { public string MY_VALUE { get; set; } }",
    "class B { public string MY_VALUE { get; set; } public string MyValue { get; set; } }"
);
```

Because PascalCase fallback will cause this situation:

```csharp
var target = new global::B();
target.MY_VALUE = source.MY_VALUE;
target.MyValue = source.MY_VALUE;
return target;
```

latonz: Mhmm, great question. IMO these are edge cases, but the generated code is still acceptable. The second mapping can be ignored by adding `[MapperIgnoreTarget("MyValue")]`, correct?

Kataane: Uh, okay. It's a edge cases. And the behavior is generally justified, but will it come as a surprise to the user. 

Since the library is mostly designed for explicit mapping, and if it doesn't happen, at least there are warnings. Which allows you not to shoot yourself in the foot because of a broken contract. That's why I love this library

But here we are silently mapping one property into two properties, without any warnings. 

Maybe add settings like EnablePascalCaseFallback.

latonz: I understand your arguments. But without the pascal case fallback how do you know which side of the mapping is snake_case and which is PascalCase? IMO it is quite common to map in both directions and only one side being snake_case....
Probably the best option in your sample would be only matching the exact match and only use pascal case match as a fallback.

Kataane: To summarize, there should be a result:

```
var target = new global::B();
target.MY_VALUE = source.MY_VALUE;
// target.MyValue = source.MY_VALUE; - ignore
return target;
````

Is there a mechanism somewhere in the code to match properties exactly? To tell the generator `MY_VALUE` is already taken, don't take it anymore because I found an exact match for it.

Comment (latonz) [reply test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs:255]: Thinking on this again, i think mapping both is what aligns best with Mapperly's current behaviour and expectations. Mapperly works on the list of target members and tries to find a matching source member for each target member. For example (how it works today with case-insensitive mappings):
```csharp
[Mapper(PropertyNameMappingStrategy = PropertyNameMappingStrategy.CaseInsensitive)]
static partial class Mapper
{
    public static partial A Map(B src);
}

record A
{
    public int Value1 { get; set; }

    public int value1 { get; set; }
}

record B
{
    public int Value1 { get; set; }
}
```

generates
```csharp
static partial class Mapper
{
    [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.3.0.0")]
    public static partial global::A Map(global::B src)
    {
        var target = new global::A();
        target.Value1 = src.Value1;
        target.value1 = src.Value1;
        return target;
    }
}
```

Therefore in your sample it should map both members:
```csharp
var target = new global::B();
target.MY_VALUE = source.MY_VALUE;
target.MyValue = source.MY_VALUE;
return target;
```

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
Kataane: Add the `PropertyNameMappingStrategy.cs` `UpperSnakeCase` option, as well as the logic for it.

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
Kataane: Add in the `MemberPathCandidateBuilder.cs` fallback to PascalCase

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyTest.cs
Kataane: Add diff tests for snake case and upper snake case to the `ObjectPropertyTest.cs`

Kataane:
Hi @latonz,
Could you please review the changes? I would like to add this functionality to my code as soon as possible.

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -1,66 +1,273 @@
...
-    public static IEnumerable<StringMemberPath> BuildMemberPathCandidates(string name)
+    public static IEnumerable<StringMemberPath> BuildMemberPathCandidates(string name, PropertyNameMappingStrategy strategy)
+    {
+        if (string.IsNullOrEmpty(name))
+        {
+            yield break;
+        }
+
+        // 1. Always yield the exact match first (Fastest path)
+        yield return new StringMemberPath(new[] { name });
```
latonz: this could be simplified by using an enumerable with prepend and a switch:
```csharp
if (string.IsNullOrEmpty(name))
    return Enumerable.Empty<StringMemberPath>();

var candidates = strategy switch
{
    PropertyNameMappingStrategy.SnakeCase => BuildSnakeCaseCandidates(name),
    PropertyNameMappingStrategy.UpperSnakeCase => BuildUpperSnakeCaseCandidates(name),
    _ => BuildCaseSensitivePermutations(name),
};

// the very first candidate to be matched should always be the exact match
return candidates.Prepend(new StringMemberPath(new[] { name }));
```

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -1,66 +1,273 @@
...
+
+                // Lower/Upper fallbacks
+                // Verify they are different from original to avoid dups (e.g. "id".ToLower() == "id")
+                var lower = name.ToLowerInvariant();
+                if (!string.Equals(lower, name, StringComparison.Ordinal))
+                    yield return new StringMemberPath(new[] { lower });
+
+                var upper = name.ToUpperInvariant();
+                if (!string.Equals(upper, name, StringComparison.Ordinal) && !string.Equals(upper, lower, StringComparison.Ordinal))
+                    yield return new StringMemberPath(new[] { upper });
```
latonz: Why is this needed? If the case-insensitive strategy is selected, case-insensitive matching is applied anyway? If this is really required, add a test that fails if these lines are dropped.

test/Riok.Mapperly.Tests/Descriptors/MemberPathCandidateBuilderTest.cs
```diff
@@ -29,12 +30,18 @@ public class MemberPathCandidateBuilderTest
     )]
     public void BuildMemberPathCandidatesShouldWork(string name, string[] chunks)
```
latonz: Add tests for the new strategies (probably just add the strategy as parameter of this test method)

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -1,66 +1,273 @@
...
+                break;
+
+            default:
+                foreach (var item in BuildCaseSensitivePermutations(name))
+                    yield return item;
+                break;
+        }
+    }
+
+    private static IEnumerable<StringMemberPath> BuildCaseSensitivePermutations(string name)
```
latonz: I'd prefer a name like `BuildPascalCasePermutations`

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -1,66 +1,273 @@
...
+        }
+
+        return true;
+    }
+
+    private static bool IsUpperSnakeCase(string str)
+    {
+        foreach (var t in str)
+        {
+            if (char.IsLower(t))
```
latonz: Doesn't this return false for `_`? Same for `IsSnakeCase`?

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -1,66 +1,273 @@
...
+        {
+            yield return box;
+        }
+    }
+
+    private static IEnumerable<StringMemberPath> GenerateSnakeCasePermutations(string name)
     {
-        if (name.Length == 0)
+        // Quick check before allocating array
+        if (name.IndexOf('_') == -1)
```
latonz: Why aren't you using `!name.Contains('_')`

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -1,66 +1,273 @@
...
+        foreach (var t in str)
+        {
+            if (char.IsUpper(t))
+                return false;
+        }
+
+        return true;
+    }
+
+    private static bool IsUpperSnakeCase(string str)
```
latonz: These would probably better fit in `MemberNamingUtil`... Also add unit tests...

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
```diff
@@ -1,66 +1,273 @@
...
+
+        return count;
+    }
+
+    private static List<string> BuildPermutationList(string source, int[] indices, int indicesCount, int mask, bool skipDelimiter)
+    {
+        // Pre-calculate exact capacity to avoid internal List resizing
+        // 1 base segment + 1 for every split
+        var capacity = 1 + CountSetBits(mask);
+        var list = new List<string>(capacity);
```
latonz: It looks like you attempted to optimize for performance. Have you benchmarked this yet?
From my perspective the previous code was easier to follow, and I might be biased, so I would still like to double check whether the added complexity is justified.
There is probably still room for significant optimization by using (ReadOnly)Spans, stackalloc and/or array pools. I'm not sure whether we should move these performance optimizations into a separate PR.

docs/docs/configuration/mapper.mdx
```diff
@@ -193,30 +193,189 @@ or by setting the `IgnoreObsoleteMembersStrategy` option of the `MapperAttribute
...
+
+    </TabItem>
+
+</Tabs>
+
+:::note
+Property name mapping strategies follow a priority order:
+1. Exact name match (fastest)
+2. Strategy-specific conversions (snake_case, UPPER_SNAKE_CASE)
+3. PascalCase permutations (for backward compatibility)
```
latonz: IMO this is not only for backward compatibility, it also makes sense. I'd remove the part in ()...

docs/docs/configuration/mapper.mdx
```diff
@@ -193,30 +193,189 @@ or by setting the `IgnoreObsoleteMembersStrategy` option of the `MapperAttribute
...
+4. Case variations (for CaseInsensitive strategy)
+:::
+
+:::warning
```
latonz: IMO this should only be an info, not a warning.

docs/docs/configuration/mapper.mdx
```diff
@@ -193,30 +193,189 @@ or by setting the `IgnoreObsoleteMembersStrategy` option of the `MapperAttribute
...
+
+- **Before**: `ModelName` → `modelname`
+- **After**: `ModelName` → `model_name` (with SnakeCase) or `MODEL_NAME` (with UpperSnakeCase)
+
+When migrating from `CaseSensitive` to snake case strategies:
+
+- **Before**: Only exact matches work
+- **After**: Automatic conversion between PascalCase and snake_case conventions
+
+For existing mappers using explicit `[MapProperty]` attributes, the new strategies provide fallback mappings while maintaining backward compatibility.
```
latonz: I think we can remove this block since it does not seem to provide much value.

src/Riok.Mapperly/Descriptors/MemberPathCandidateBuilder.cs
Kataane: >It looks like you attempted to optimize for performance

Yes and no. The main goal was to unify the BuildPermutationList method. I wanted it to work for any notation, so I didn't focus much on optimization, such as using (ReadOnly)Spans, stackalloc and/or array pools. 

I was wondering, though, if there will be any other notations in the future. KebabCase is not possible as a notation in C#.

If you are confused by this code, I can revert it.

Kataane:
Hi @latonz,

Thank you for your review.

I fixed the issues based on your reviews. The main point that is still unresolved is the complex login in MemberPathCandidateBuilder.cs.

PR #2020, opened 2025-11-15
Performance improvements: AttributeDataAccessor

faddiv:
# Performance improvements: AttributeDataAccessor

## Description

This PR showcases the performance improvements I had in mind:
The idea was to improve the AttributeDataAccessor since that is one of the most used parts of the code.
I replaced the reflection with a more direct approach. I also used caching on them to save some more work.
I also created a comparison benchmark for each step, so it can be compared to see how much benefit it can give.
My comments about the changes:
- Some calls don't benefit from caching since it is called once or twice, others have lots of benefits. Regardless, I did it for each other for consistency.
- I hoped for a little more benefit, but this is what I get: ~93 to ~82 ms on large compilation and nothing on small. I included the results, so you can see for yourself.
- I have more ideas that could help, but I leave it for later.
- Also, a few could return a read-only list directly, since it is `ToList`-ed. I didn't touch it for consistency.
- There was a single static call, which I could separate from the rest. This way, the `SymbolAccessor` doesn't need to be passed around and checked for null. I think this makes this class cleaner.
- Some places, not null is returned, I think the null check could be moved out, for more consistency.
- Other places, if the processing fails, it returns null. I think it helps in stability. A half-typed-in attribute/method or whatever doesn't cause an exception this way.
- Reverted: I reduced the number of different methods: TryGetAttributes, GetAttributes -> GetAttributes. 
   - In tests, this doesn't cause any problem; however, there is a weird problem, which was there from the beginning: The `GetBestTypeByMetadataName` in `WellKnownTypes` won't find some types if it is not used in the code. It can occur when some tests are run alone. This could cause a problem, and I would look for some way to implement this differently. Or let's use `TryGetAttributes` instead. 
   - In .NET 4.8, this causes problems, so I reverted on NotNullIfNotNull. I will look into other cases to revert to the original variant.
   - I also tried `TryGetAttributes` at other places, but it caused quite a lot of failures.
- I created the `NotNullIfNotNullConfiguration`, since the `NotNullIfNotNullAttribute` is internal, but the `AttributeDataAccessor` is public.
- I switched on the analyzer performance output on integration test `<ReportAnalyzer>true</ReportAnalyzer>` so I can observe how it is changed in practice. This reflected the gain in the benchmark.

Fixes # (issue)

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
+++ b/src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
@@ -107,18 +107,15 @@ public class AttributeDataAccessor(SymbolAccessor symbolAccessor) : IAttributeDa
     public MapperIgnoreObsoleteMembersAttribute? ReadMapperIgnoreObsoleteMembersAttribute(ISymbol symbol)
     {
         var attrData = GetAttribute<MapperIgnoreObsoleteMembersAttribute>(symbol);
-        if (attrData == null)
-        {
-            return null;
-        }
-
-        return new MapperIgnoreObsoleteMembersAttribute(
-            GetSimpleValueOrDefault(
-                attrData,
-                nameof(MapperIgnoreObsoleteMembersAttribute.IgnoreObsoleteStrategy),
-                IgnoreObsoleteMembersStrategy.Both
+        return attrData is not null
+            ? new MapperIgnoreObsoleteMembersAttribute(
+                GetSimpleValueOrDefault(
+                    attrData,
+                    nameof(MapperIgnoreObsoleteMembersAttribute.IgnoreObsoleteStrategy),
+                    IgnoreObsoleteMembersStrategy.Both
+                )
             )
-        );
+            : null;
     }
 
     public IEnumerable<NestedMembersMappingConfiguration> ReadMapNestedPropertiesAttribute(ISymbol symbol)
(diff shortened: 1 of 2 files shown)


benchmarks/Results/baseline/results/Riok.Mapperly.Benchmarks.SourceGeneratorBenchmarks-report-github.md
```diff
@@ -0,0 +1,14 @@
...
+
+BenchmarkDotNet v0.15.4, Windows 11 (10.0.26200.6899)
+AMD Ryzen 5 2600 3.40GHz, 1 CPU, 12 logical and 6 physical cores
+.NET SDK 9.0.306
+  [Host] : .NET 9.0.10 (9.0.10, 9.0.1025.47515), X64 RyuJIT x86-64-v3
+
+Job=InProcess  Toolchain=InProcessEmitToolchain  
+
+```
+| Method       | Mean      | Error     | StdDev    | Gen0      | Gen1     | Allocated   |
```
latonz, line 11: IMO we shouldn't checkin the generated results of the benchmarks...

src/Riok.Mapperly/Configuration/CachedAttributeDataAccessor.cs
```diff
@@ -0,0 +1,205 @@
...
+using System.Runtime.Serialization;
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Abstractions;
+
+namespace Riok.Mapperly.Configuration;
+
+public class CachedAttributeDataAccessor(IAttributeDataAccessor attributeDataAccessor) : IAttributeDataAccessor
+{
+    private readonly Dictionary<CacheKey, object?> _cache = new();
+    private readonly IAttributeDataAccessor _attributeDataAccessor = attributeDataAccessor;
```
latonz: This can be replaced with the primary ctor field, no?

src/Riok.Mapperly/Configuration/CachedAttributeDataAccessor.cs
```diff
@@ -0,0 +1,205 @@
+﻿using System.Runtime.CompilerServices;
+using System.Runtime.Serialization;
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Abstractions;
+
+namespace Riok.Mapperly.Configuration;
+
+public class CachedAttributeDataAccessor(IAttributeDataAccessor attributeDataAccessor) : IAttributeDataAccessor
```
latonz, line 8: To what extent do the performance gains result from caching?

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
```diff
@@ -15,197 +16,557 @@ namespace Riok.Mapperly.Configuration;
...
-        return arg.Kind switch
+        return attrData.AttributeClass?.TypeArguments is { Length: > 0 } typeArguments && typeArguments.Length > index
+            ? typeArguments[index]
+            : throw new InvalidOperationException($"Could not get type argument {index} of attribute {attrData.AttributeClass?.Name}");
+    }
+
+    private static TValue? GetSimpleValue<TValue>(AttributeData attrData, string propertyName)
+        where TValue : struct
+    {
+        if (TryGetTypedConstant(attrData, propertyName, out var typedConstant, out _))
```
latonz: Invert the if and return early.

benchmarks/Results/baseline/results/Riok.Mapperly.Benchmarks.SourceGeneratorBenchmarks-report-github.md
faddiv: I checked in this only so I can show you my results. When I open this PR, I will remove this.
For now, I'm interested in whether you agree with this approach, considering the performance gains.

src/Riok.Mapperly/Configuration/CachedAttributeDataAccessor.cs
faddiv: After the `AttributeDataAccessor`, it is smaller. As mentioned, in some cases, it is pointless; in other cases, it provides benefits.
For me, the large compile had these results:
- Baseline: 93.267 ms and 21185.55 KB
- Without caching: 85.923 ms and 20239.85 KB
- With caching: 82.853 ms and 19992.74 KB

src/Riok.Mapperly/Configuration/AttributeDataAccessor.cs
faddiv: I used the conditional operator where it seemed to fit, and early return where it wasn't.

src/Riok.Mapperly/Configuration/CachedAttributeDataAccessor.cs
faddiv: Done. BTW: Do you prefer using the primary ctor fields directly? I always disliked that it is not read-only.

latonz: IMO it has pros and cons. Often it saves quite a bit of boilerplate, but IMO it results in unnecessary diffs when there is a need to switch back to a regular ctor 🤷

latonz:
Hi @faddiv ,

Thank you for your effort on this and for providing the detailed performance benchmarks.
We have decided not to continue with this PR for now. While we appreciate the investigation into these optimizations, we are going to hold off on merging these changes at this time.

Thanks again for your contribution!

PR #2036, opened 2025-11-25
Fix name collision on inlined mapping

### LINKED ISSUE #2012: Lambda expression variable issue with `x` when comparing to source
The following code works as expected.

```csharp
public record Blog(List<Post> Posts);
public record Post(int Id);
public record BlogDto(int Id);

[Mapper(RequiredMappingStrategy = RequiredMappingStrategy.Target)]
public static partial class BlogMapper
{
	public static partial IQueryable<BlogDto> ProjectToDto(this IQueryable<Blog> q);

	[MapPropertyFromSource(nameof(BlogDto.Id), Use = nameof(MapId))]
	public static partial BlogDto BlogMap(Blog blog);

	private static int MapId(Blog blog) => blog.Posts.Count(TEST => TEST.Id == blog.Posts.Count);
}
```

Will generate the following snippet.

```csharp
        return global::System.Linq.Queryable.Select(
            q,
            x => new global::BlogDto(
                global::System.Linq.Enumerable.Count(x.Posts, TEST => TEST.Id == x.Posts.Count)
            )
        );
```

However if you modify the `MapId` method to use an `x` for the lambda variable, you get an incorrect mapping.
```csharp
private static int MapId(Blog blog) => blog.Posts.Count(x => x.Id == blog.Posts.Count);

        return global::System.Linq.Queryable.Select(
            q,
            x => new global::BlogDto(
                global::System.Linq.Enumerable.Count(x.Posts, x => x.Id == x.Posts.Count)
            )
        );
```

This is not a contrived example, but an actual issue I ran into. Maybe instead of the generator defaulting to `x` for the queryable select, use something more unique?

faddiv:
# Fix name collision on inlined mapping

## Description

This PR ensures that the inlined mapping in the projections doesn't cause parameter collision.

Fixes #2012

## Checklist

### DIFF
--- a/test/Riok.Mapperly.IntegrationTests/_snapshots/ProjectionMapperTest.SnapshotGeneratedSource.verified.cs
+++ b/test/Riok.Mapperly.IntegrationTests/_snapshots/ProjectionMapperTest.SnapshotGeneratedSource.verified.cs
@@ -10,59 +10,59 @@ namespace Riok.Mapperly.IntegrationTests.Mapper
 #nullable disable
             return global::System.Linq.Queryable.Select(
                 q,
-                x => new global::Riok.Mapperly.IntegrationTests.Dto.TestObjectDtoProjection(x.CtorValue)
+                x1 => new global::Riok.Mapperly.IntegrationTests.Dto.TestObjectDtoProjection(x1.CtorValue)
                 {
-                    IntValue = x.IntValue,
-                    IntInitOnlyValue = x.IntInitOnlyValue,
-                    RequiredValue = x.RequiredValue,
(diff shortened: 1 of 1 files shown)


latonz:
Mapperly usally works top to bottom. Here, the approach is inversed (the parameter names are reserved from the innermost mapping). I'd prefer doing it the other way around (that's what I tried): rewriting the inlined mapping with unique parameter names. That's actually also what we do for the source parameter.

latonz:
Feel free to reopen with a rewrite parameters approach 😊 Closing for now as no update was provided for a couple of weeks.

PR #2086, opened 2025-12-26
fix: stackoverflow on nullable readonly existing target

### LINKED ISSUE #2078: InsufficientExecutionStackException with read-only property
- [X] I have read the [documentation](https://mapperly.riok.app), including the [FAQ](https://mapperly.riok.app/docs/getting-started/faq/)
- [X] I can reproduce the bug using the latest prerelease version
- [X] I have searched existing [discussion](https://github.com/riok/mapperly/discussions) and [issue](https://github.com/riok/mapperly/issues) to avoid duplicates

**Describe the bug**
Adding a clone/copy map definition with a class that has a read-only property throws InsufficientExecutionStackException (similar to #1882 ?)

**Repro**
https://github.com/nas-rda/Mapperly1882Extended

**Declaration code**
```csharp
public class Sample
{
    public string Value { get; set; }

    public Task? ReadonlyProperty => null;
}

[Mapper(RequiredMappingStrategy = RequiredMappingStrategy.Target, UseDeepCloning = true)]
public static partial class Map
{
    public static partial void Copy(this Sample source, Sample target);
}
```

None. InsufficientExecutionStackException is thrown in MapperGenerator.

```csharp
public static partial class Map
{
    public static partial void Copy(this Sample source, Sample target)
    {
        target.Value = source.Value;
    }
}
```

None.

- Mapperly Version: 4.3.1-next.0
- .NET Version: 10.0.0
- Target Framework: net10.0
- Compiler Version: 18.0.2+b0f34d51f
- C# Language Version: 14.0
- IDE: Rider 2025.3.0.4
- OS: Windows 10 22H2

**Workaround**
Using `MapperIgnoreTarget(nameof(@Sample.ReadonlyProperty))` will not cause the exception.

faddiv:
# Fix stackoverflow on nullable readonly existing target

## Description

This PR fixes the stack overflow exception for nullable, readonly existing target mappings.

I’d also like to get your thoughts on an idea about the tests: These tests practically test the same output on different cases. Is it enough to generate a snapshot only for a single one, and just assert for the same output on the rest? I think there are too much snapsots and hard to manage. This approach can limit it.

Another thought about the bug report itself: when one of the parameters was the Task, the generator explored it deeply, which is a waste of resources. It doesn't fix the root problem, but somehow this should be limited. Since limiting this type or that couldn't really help. My idea is to limit exploring the read-only targets themselves. So if the target is read-only, then it is explored only if it is explicitly asked in some manner. WDYT?

Fixes #2078

## Checklist


latonz (approved):

latonz:
> I’d also like to get your thoughts on an idea about the tests: These tests practically test the same output on different cases. Is it enough to generate a snapshot only for a single one, and just assert for the same output on the rest? I think there are too much snapsots and hard to manage. This approach can limit it.

IMO we should decide on a case by case basis whether a snapshot test makes sense. In this case, since it is only about diagnostics, I prefer the fluent `TestHelper` assertions. They are more focused and make these tests easier to maintain. However, whenever asserting generated mapping code, I think it is simpler to use snapshot tests.

latonz:
> Another thought about the bug report itself: when one of the parameters was the Task, the generator explored it deeply, which is a waste of resources. It doesn't fix the root problem, but somehow this should be limited. Since limiting this type or that couldn't really help. My idea is to limit exploring the read-only targets themselves. So if the target is read-only, then it is explored only if it is explicitly asked in some manner. WDYT?

Yeah I think it would make sense to only try to map direct properties of the root but not explore further. Does that make sense?

PR #2088, opened 2025-12-29
feat: Assembly-wide UseStaticMapper

### LINKED ISSUE #1146: Allow assembly-wide usage of UseStaticMapper
I have a static mapper that is used in many other mappers. I don't want to add the `UseStaticMapper` attribute to each mapper class.

`UseStaticMapper` should be allowed as an assembly attribute and should apply to all mappers in the assembly.

A new optional property `UseAsStaticMapper` (or similar) on the `Mapper` attribute that enables the usage of this mapper as a static mapper in all other mappers.

faddiv:
# Assembly-wide UseStaticMapper

## Description

This PR makes the UseStaticMapper attribute usable at the assembly level, so the specified mappers are available for all other mappers without needing to be specified explicitly.

## Remarks

- I thought this would be a useful feature while working with Mapperly, then I saw there is already an open issue for this, so I implemented it. 🙂 
- Instead of `ImmutableArray<UseStaticMapperConfiguration> _globalStaticMappers;` I considered passing it to the Build method. Does that make more sense? But if yes, then some other input also, which is not a dependency but a parameter.
- ` AttributeDataAccessor.Access<UseStaticMapperAttribute<object>, UseStaticMapperConfiguration>(attributeData);` also processes the non-generic variant, so I left it this way.
- A little bit restructured the methods in `ExternalMappingsExtractor` to better fit this new feature.
- The integration test on the generic variant was removed in the end. This way, no conditional compilations are needed.

Fixes #1146

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
+++ b/src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
@@ -11,12 +11,12 @@ namespace Riok.Mapperly.Descriptors.ExternalMappings;
 internal static class ExternalMappingsExtractor
 {
     public static IEnumerable<IUserMapping> ExtractExternalMappings(
-        IEnumerable<UseStaticMapperConfiguration> globalStaticMappers,
+        IEnumerable<UseStaticMapperConfiguration> assemblyScopedStaticMappers,
         SimpleMappingBuilderContext ctx,
         INamedTypeSymbol mapperSymbol
     )
     {
-        return ExtractExternalStaticMappings(globalStaticMappers, ctx)
+        return ExtractExternalStaticMappings(assemblyScopedStaticMappers, ctx)
             .Concat(ExtractExternalStaticMappings(ExtractStaticMappersFromAttributes(ctx, mapperSymbol), ctx))
             .Concat(ExtractExternalInstanceMappings(ctx, mapperSymbol));
     }


test/Riok.Mapperly.Tests/Mapping/UseStaticMapperTest.cs
faddiv, line 1: As an afterthought: I should test what happens when a mapper references itself. (But later for now.)

faddiv: In the end, it has no negative effect, but I see an opportunity for some short-circuiting, and I cleaned up some leftover code.

docs/docs/configuration/external-mappings.mdx
```diff
@@ -73,6 +73,38 @@ it will use the provided implementation by the `BananaMapper`.
...
+        ```
+
+    </TabItem>
+    <TabItem value="global-generic" label="With generic argument">
+
+        ```csharp
+        using Mapperly.Abstractions;
+
+        [assembly: UseStaticMappers<BananaMapper>]
+        [assembly: UseStaticMappers<AppleMapper>]
```
latonz: In the other sample we don't state the generic parameter, just the `typeof`, I'd do the same here and remove the generic param sample.

docs/docs/configuration/external-mappings.mdx
```diff
@@ -73,6 +73,38 @@ it will use the provided implementation by the `BananaMapper`.
...
+
+It is also possible to configure Mapperly to always consider certain external mappers
+without needing to specify them on each mapper class.  
+This can be done through the `UseStaticMappers` attribute.
+
+<Tabs>
+    <TabItem value="global-typeod" label="With typeof parameter" default>
+
+        ```csharp
+        using Mapperly.Abstractions;
```
latonz: We never show the using in the code samples, I'd remove it here too.

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
```diff
@@ -101,6 +101,11 @@ internal static IEnumerable<IUserMapping> ExtractUserImplementedMappings(
         bool isExternal
     )
     {
+        // Ignore the mapper type itself.
+        if (SymbolEqualityComparer.Default.Equals(type, ctx.MapperDeclaration.Symbol))
+        {
+            return [];
+        }
         var methods = ctx
```
latonz, line 110: Nit: we usually add an empty NL after a block.

docs/docs/configuration/external-mappings.mdx
```diff
@@ -73,6 +73,38 @@ it will use the provided implementation by the `BananaMapper`.
 Used mappers themselves can be Mapperly backed classes.  
 The [`AutoUserMappings`](./user-implemented-methods.mdx) value also applies to the usage of external mappers.
 
+## Globally available external mappers
```
latonz: Not sure about the term "globally" here... It is still limited to the scope of the assembly...

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
```diff
@@ -10,27 +10,50 @@ namespace Riok.Mapperly.Descriptors.ExternalMappings;
...
+    }
 
-        var externalInstanceMappers = ctx
+    private static IEnumerable<IUserMapping> ExtractExternalStaticMappings(
+        IEnumerable<UseStaticMapperConfiguration> staticMappers,
+        SimpleMappingBuilderContext ctx
+    )
+    {
+        var staticExternalMappers = staticMappers.SelectMany(x =>
+            UserMethodMappingExtractor.ExtractUserImplementedMappings(
```
latonz, line 40: This now happens for every mapper. Couldn't this be improved, that it only happens once? E.g. do it in the mapper generator?

src/Riok.Mapperly/Descriptors/DescriptorBuilder.cs
```diff
@@ -20,6 +21,7 @@ public class DescriptorBuilder
 {
     private readonly MapperDescriptor _mapperDescriptor;
     private readonly SymbolAccessor _symbolAccessor;
+    private readonly ImmutableArray<UseStaticMapperConfiguration> _globalStaticMappers;
```
latonz: IMO these are not really global but assembly scoped.

test/Riok.Mapperly.IntegrationTests/Mapper/GlobalMappers.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Riok.Mapperly.Abstractions;
+using Riok.Mapperly.IntegrationTests.Dto;
+using Riok.Mapperly.IntegrationTests.Mapper;
+using Riok.Mapperly.IntegrationTests.Models;
+
+[assembly: UseStaticMapper(typeof(GlobalMappers))]
+
+namespace Riok.Mapperly.IntegrationTests.Mapper
```
latonz: I'd also use a term like `AssemblyMapper` instead of `GlobalMappers`. Same applies at several locations where the term global is used.

latonz (changes requested):
Thanks for this contribution, I added my feedback and also copilot added some feedback.

docs/docs/configuration/external-mappings.mdx
faddiv: I used it here just because the attribute must go after the usings, but I'm fine this way, too.

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
faddiv: Ehh! A weird case when the git goes into some invalid state. I managed to commit the nl there. 😕

docs/docs/configuration/external-mappings.mdx
faddiv: I modified where I used globally.

src/Riok.Mapperly/Descriptors/ExternalMappings/ExternalMappingsExtractor.cs
faddiv: I'm not sure if it can work that way. AFAIK, on changes, every sourcegenerator runs to collect the significant data, then it compares with the previous one. For this comparison, it needs to run even if there is no change.
I would test it out, since it is an interesting question, but this part depends on the `SimpleMappingBuilderContext`, so it can't be moved there anyway. 🙂

latonz: But outside this context, only the symbol- and the attribute accessors are really needed, no? That said, I’m fine with handling this in a separate PR. The idea would be to move user mapping extraction into a dedicated source-gen pipeline step. Especially with the new feature and the possible large number of assembly wide static mappers, this could significantly improve performance.

latonz: I created https://github.com/riok/mapperly/issues/2096 to handle this.

latonz:
Thanks for this contribution!

PR #2095, opened 2025-12-30
feat: Optimize source generation by using GetText

faddiv:
# Optimize source generation by using GetText

## Description

This is a small improvement I wished to send in for a while:
This way of the source text generation is better, since it internally contains a decision: if the output size is larger than the Large Object Heap limit, then it uses `LargeTextWriter`, which breaks the source apart, so the LOH allocation is avoided. On the benchmark, there is no visible change, but I think it is worth it.

Fixes # (issue)

## Checklist


latonz (approved):
Definitely worth it, thanks for the contribution! ❤️

Issue #2096, opened 2025-12-30
Extract user mappings only once in source generator pipeline


latonz:
In a single source generator invocation, the same external mappings are repeatedly discovered for each mapper. For example, if a mapper is used assembly wide as an external mapper, its mappings are rediscovered for every mapper being processed. The same happens when multiple mappers reference the same external mapper.
To improve performance, this discovery should happen only once. The extracted user mapping information per mapper declaration can then be reused by the descriptor builder.

_See [#2088](https://github.com/riok/mapperly/pull/2088/changes#r2653289240)_

faddiv:
I bring here the discussion from the PR:
Currently, it is not possible easily extract these.
As I see, these are the possibilities that  could be done:
- Bringing the settings needed for this connection to the UseStaticMapper itself. (I see AutoUserMappings, but is there another?) This would be a breaking change.
- If class level, then it needs to look at the parent class and process the Mapper parameters. But what about the assembly-level attributes?
- The UseStaticMapper still could be preprocessed, and the final IUserMapping could be created in the main step with the context dependent informations.

latonz:
I'll leave it as it is for now.

PR #2097, opened 2026-01-01
Add name to use static mapper generator step

faddiv:
# Add name to use static mapper generator step

## Description

Hi @latonz,

I doubt that the external mappers can be extracted (See #2096), since the behaviour of these methods is affected by the settings added to the mapper attribute. In other words, using the same type in different places may result in different results.

In this PR, I added an explicit name to the generator step and extended the `IncrementalGeneratorTest` to check this.
I also experimented with what happens when the static mapper doesn't change, and only a new mapper is added. I thought it would be cached or unmodified, but it gets modified, no matter what I do. The reason is that the `MapperType` is considered different between compilations. Do you think this is okay?

Fixes # (issue)

## Checklist

- [ ] Integration tests are added/updated (as applicable, especially if feature/bug depends on Roslyn or framework version in use)

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -17006,9 +17006,9 @@
       }
     },
     "node_modules/qs": {
-      "version": "6.14.0",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.0.tgz",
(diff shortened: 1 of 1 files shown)


latonz:
Thanks for the analysis and I think indeed you are correct that these cannot be extracted easily... I'm not sure whether it would make sense to enforce affected settings to be unique for a given assembly for the assembly-wide imported static mappers🤔

Regarding the changing mapping type: this is due to the compilation, which is tied to the symbol which changes with each compilation.

faddiv:
I removed the experimental test, which was failing, so the rest can be merged.

PR #2100, opened 2026-01-02
fix: don't emit unnecessary null-handling if source and target are nullable

Kataane:
# Fix the handling of nullable types in the NullDelegateMapping.

## Description

Hi @latonz.

This pull request (PR) addresses an issue with nullable type handling in the NullDelegateMapping class. It ensures that, when both the source and target types are nullable and have the same underlying type, the delegate mapping is built properly without unnecessary casts or null checks.

**Key changes:**
- Special case handling was added in NullDelegateMapping.cs for nullable source and target types with identical underlying types.
- Simplified dictionary value mapping by removing unnecessary null checks in the generated code.
- Added comprehensive test coverage for nullable array mappings.
- Test snapshots were updated to reflect the simplified generated code.

**Technical details:**
The fix adds a condition to NullDelegateMapping.Build() that checks if both the source and target types are nullable and share the same underlying type. When this condition is met, it delegates to the underlying mapping without additional null handling, resulting in cleaner generated code.

**Example of improvement:**
Before: `target[item.Key] = item.Value == null ? default : item.Value;`
After: `target[item.Key] = item.Value;`

related #1614

## Checklist

### DIFF
--- a/src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
+++ b/src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
@@ -10,99 +10,113 @@ namespace Riok.Mapperly.Descriptors.Mappings;
 /// <summary>
 /// Null aware delegate mapping. Abstracts handling null values of the delegated mapping.
 /// </summary>
-public class NullDelegateMapping : NewInstanceMapping
+/// <remarks>
+/// This mapping handles null propagation by wrapping a delegate mapping with appropriate
+/// null checks and fallback values. It optimizes away when no null handling is required.
+/// </remarks>
+public class NullDelegateMapping(
+    ITypeSymbol nullableSourceType,
+    ITypeSymbol nullableTargetType,
+    INewInstanceMapping delegateMapping,
+    NullFallbackValue nullFallbackValue
+) : NewInstanceMapping(nullableSourceType, nullableTargetType)
 {
     private const string NullableValueProperty = nameof(Nullable<>.Value);
 
-    private readonly INewInstanceMapping _delegateMapping;
-    private readonly NullFallbackValue _nullFallbackValue;
-
-    public NullDelegateMapping(
-        ITypeSymbol nullableSourceType,
-        ITypeSymbol nullableTargetType,
-        INewInstanceMapping delegateMapping,
-        NullFallbackValue nullFallbackValue
-    )
-        : base(nullableSourceType, nullableTargetType)
-    {
-        _delegateMapping = delegateMapping;
-        _nullFallbackValue = nullFallbackValue;
-
-        // the mapping is synthetic (produces no code)
-        // if and only if the delegate mapping is synthetic (produces also no code)
-        // and no null handling is required
-        // (this is the case if the delegate mapping source type accepts nulls
-        // or the source type is not nullable and the target type is not a nullable value type (otherwise a conversion is needed)).
-        IsSynthetic =
-            _delegateMapping.IsSynthetic
-            && (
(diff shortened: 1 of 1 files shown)


src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
```diff
@@ -45,6 +45,11 @@ public override ExpressionSyntax Build(TypeMappingBuildContext ctx)
         if (_delegateMapping.SourceType.IsNullable())
             return _delegateMapping.Build(ctx);
 
+        if (SourceType.IsNullable() && TargetType.IsNullable() && SymbolEqualityComparer.Default.Equals(SourceType, TargetType))
```
latonz: I don't think the symbol equality comparison does make much sense here, does it? Wouldn't `_delegateMapping.IsSynthetic` be more feasable?

src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
latonz: This needs adjustments too, no?

latonz (changes requested):

src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
Kataane: You are right. I fixed it.

test/Riok.Mapperly.Tests/Mapping/NullableTest.cs
```diff
@@ -163,7 +163,14 @@ TestSourceBuilderOptions.Default with
...
     [Fact]
-    public void NonNullableToNullableValueType()
+    public void NullableToNullableType()
+    {
+        var source = TestSourceBuilder.Mapping("int[]?", "int[]?");
+        TestHelper.GenerateMapper(source).Should().HaveSingleMethodBody("return source;");
+    }
+
+    [Fact]
+    public void NonNullableToNullableValueType2()
```
latonz, line 173: The test method name 'NonNullableToNullableValueType2' is unclear due to the '2' suffix. Consider a more descriptive name that explains what distinguishes this test from similar tests, such as 'NonNullableToNullableValueTypeCast' or 'NonNullableDateTimeToNullableDateTime'.

src/Riok.Mapperly/Descriptors/Mappings/NullDelegateMapping.cs
```diff
@@ -10,88 +10,113 @@ namespace Riok.Mapperly.Descriptors.Mappings;
...
+
+        // No null handling needed if:
+        // 1. Delegate already accepts nulls
+        // 2. Source isn't nullable and target isn't nullable value type (no conversion needed)
+        // 3. Both source and target are nullable (pass-through)
+        var delegateAcceptsNulls = delegateMapping.SourceType.IsNullable();
+        var noConversionNeeded = !sourceType.IsNullable() && !targetType.IsNullableValueType();
+        var bothNullable = sourceType.IsNullable() && targetType.IsNullable();
+
+        return delegateAcceptsNulls || noConversionNeeded || bothNullable;
```
latonz, line 69: The logic for determining if the mapping is synthetic appears incorrect. When both source and target are nullable, null handling may still be required if the delegate mapping doesn't accept nulls. The condition `bothNullable` should not automatically make the mapping synthetic, it should depend on whether the delegate can handle the null case. Consider: if source is `int[]?` and target is `int[]?`, but delegateMapping expects non-nullable `int[]`, this would incorrectly be marked as synthetic.

latonz:
Closing as no update was received for several weeks. Feel free to reopen with updates 😊

PR #2101, opened 2026-01-02
Add aggressive inlining support for generated mapping methods

Kataane:
# Add aggressive inlining support for generated mapping methods

## Description

This pull request adds support for applying `MethodImplOptions.AggressiveInlining` to generated mapping methods, allowing users to optimize performance based on their specific needs.

**Key changes:**
- Added `AggressiveInliningTypes` enum with `None`, `ValueTypes`, `ReferenceTypes`, and `All` options
- Added `AggressiveInliningTypes` property to `MapperAttribute` for configuration
- Updated mapping method generation to conditionally apply `[MethodImpl(MethodImplOptions.AggressiveInlining)]` attribute
- Extended unsafe accessor generation to support aggressive inlining
- Added comprehensive test coverage for all aggressive inlining scenarios

**Technical details:**
The implementation introduces a new configuration option that allows users to control when aggressive inlining should be applied to generated mapping methods. This is particularly useful for performance-critical scenarios where the overhead of method calls can be significant.

The feature supports granular control:
- `None`: No aggressive inlining (default, maintains backward compatibility)
- `ValueTypes`: Apply to struct mappings only
- `ReferenceTypes`: Apply to class mappings only  
- `All`: Apply to all mappings

Example usage:
```csharp
[Mapper(AggressiveInliningTypes = AggressiveInliningTypes.All)]
public partial class MyMapper
{
    public partial TargetDto Map(SourceEntity source);
}
```

**Performance impact:**
This feature allows users to achieve better performance in scenarios where mapping methods are called frequently, especially for simple mappings where the method call overhead is significant compared to the mapping logic itself.

## Checklist

## Related Issues

#1384


latonz:
Have you measured the impact of this? Does it significantly improve the performance? Does this already work if the user applies the attribute to the user declared method?

Kataane:
Hi @latonz 

> Have you measured the impact of this?
>Does it significantly improve the performance

According to previous measurements, #1384:
- 27x faster for simple struct types
- 1.3x faster for complex struct types
- Very slightly faster for class types

If you want, I can rerun the benchmark to ensure the result is repeated.

--- 

>Does this already work if the user applies the attribute to the user declared method

It works for all methods in the mapping class if her is one of the AggressiveInliningTypes types:

```
[Mapper(EnumMappingStrategy = EnumMappingStrategy.ByName, AggressiveInliningTypes = AggressiveInliningTypes.All)]
{
    [MapProperty(nameof(Car.Manufacturer), nameof(CarDto.Producer))]

    public static partial int ToInt(int value);
}
```

latonz:
After reviewing the proposal and the related discussions (such as in Issue #1381), we have decided not to move forward with adding AggressiveInlining as a default or generated feature at this time.

While explicit inlining can offer performance gains in specific micro-benchmarks, it also carries the risk of significantly increasing assembly size and compile times for larger object hierarchies. In most cases, we prefer to trust the .NET compiler to make the correct inlining decisions.

Thank you for your contribution and the work put into this!

PR #2116, opened 2026-01-15
fix: unsafe accessor generic type args

YoelDruxman:
# Fix unsafe accessor generic type args

## Description

Fixes incorrect UnsafeAccessor code generation when multiple types inherit from different instantiations of the same generic base class.

### Problem

When mapping types like:
```csharp
class A<T> { protected T _value { get; set; } }
class A1 : A<IB> { }
class A2 : A<IC> { }
```

The `UnsafeAccessor` would generate code with wrong type arguments for the second type, causing: `error CS1503: Argument 1: cannot convert from 'A2' to 'A<IB>'`

### Root Cause

`UnsafeAccessorTypeContext` caches accessors by `symbol.OriginalDefinition`, which ignores type arguments. When processing `A1 : A<IB>`, it caches the accessor. When processing `A2 : A<IC>`, it returns the cached accessor with `IB` type arguments instead of `IC`.

### Solution

Pass the actual `containingType` at invocation time to `BuildAssignment` and `BuildAccess` methods, instead of using the cached symbol's type arguments. This preserves accessor caching (for class generation and naming) while using correct type arguments at each invocation site.

## Checklist

- [X] The existing code style is followed
- [X] Performed a self-review of my code
- [X] Hard-to-understand areas of my code are commented
- [X] The documentation is updated (as applicable)
- [X] Unit tests are added/updated
- [X] Integration tests are added/updated (as applicable, especially if feature/bug depends on roslyn or framework version in use)

### DIFF
--- a/src/Riok.Mapperly/Descriptors/Enumerables/Capacity/EnsureCapacityMethodSetter.cs
+++ b/src/Riok.Mapperly/Descriptors/Enumerables/Capacity/EnsureCapacityMethodSetter.cs
@@ -1,3 +1,4 @@
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Riok.Mapperly.Symbols.Members;
 using static Riok.Mapperly.Emit.Syntax.SyntaxFactoryHelper;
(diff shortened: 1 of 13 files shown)


YoelDruxman:
This issue is stopping me from updating from `4.2.1` to the latest version.

latonz:
Thanks for the contribution 🥳  Please rebase onto main and I will merge it.

PR #2117, opened 2026-01-16
Add support for explicit Shallow Cloning

### LINKED ISSUE #2113: Give the possibility to clone objects without UseDeepCloning
Hi, I currently use AutoMapper to clone EF entities. The main use is either to support actually copies in the business logic (users can copy business objects) or to support forms which can be cancelled.

Given that I heavily use navigations, my object graph may be complex in some cases, which leads to UseDeepCloning to be impossible to use (maybe I want to clone a small object deep in the graph and I found myself with hundreds of new objects).

What I'm looking for is to be able to clone JUST the object I pass to the mapper, all the references either stay exactly the same, or I handle them manually.

The solution I'm looking for is a middleground between `UseDeepCloning` true and false. Because false outright refuses to do any work and just outputs something like this:

```csharp
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.3.1.0")]
        private static partial global::Jungheinrich.CalculationTool.Models.Tables.OfferComponentLTR ToNewComponentLTR(global::Jungheinrich.CalculationTool.Models.Tables.OfferComponentLTR ltr)
        {
            return ltr;
        }
```

But with Deep Cloning enabled I found myself in the situation where I need to fight Mapperly to "undo" most of the deep cloning of references because I just don't need it.

I'm just expecting a normal mapping like:

        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.3.1.0")]
        private static partial global::Tables.OfferComponentLTR ToNewComponentLTR(global::Tables.OfferComponentLTR ltr)
        {
            var target = new Tables.OfferComponentLTR();
            target.FirstProp = source.FirstProp;
            target.SecondProp = source.SecondProp;
            ....
        }

Of course I do expect nothing more than an object with all properties set exactly the same, so even references should be all the same.

I've tried to work with `UseDeepCloning` enabled but the burden is extremely high, and it basically defeats the simplicity of Mapperly, on my codebase the actual result is way way worse than AutoMapper, to the degree that I mostly need to fight against Mapperly to avoid it to clone parent and child objects.

**Additional context**
I know that one possible solution could be to introduce DTOs, but it would be like asking me to basically change the whole application just to enable mapperly to generate something which can already do.

samusaran:
# Add support for Shallow Cloning

## Description

The current behaviour of Mapperly when mapping between same types is either to return the original instance, or to perform a deep clone of the entire object graph.

Another use case lies in the middle of those two solutions: Shallow Cloning.

Shallow cloning an object allows a user to clone an object at the first level, while keeping all the references to other reference types unchanged. If a user needs some references to be handled differently, they can still use a user mapping on those specific properties.

This feature also aligns to the default behaviour of Automapper, and this could help users to more easily migrate from Automapper to Mapperly.

Fixes #2113

## Checklist

### DIFF
--- a/test/Riok.Mapperly.Tests/Mapping/ShallowCloneTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/ShallowCloneTest.cs
@@ -1,4 +1,6 @@
-﻿namespace Riok.Mapperly.Tests.Mapping;
+﻿using Riok.Mapperly.Abstractions;
+
+namespace Riok.Mapperly.Tests.Mapping;
 
 public class ShallowCloneTest
 {
@@ -9,7 +11,14 @@ public class ShallowCloneTest
             "A",
             "A",
             TestSourceBuilderOptions.WithShallowCloning,
-            "class A { public int Value { get; set; } public List<string> List { get; set; } }"
+            """
+            class A
+            {
+                public int Value { get; set; }
+                public List<string> List { get; set; }
+                public A NestedReference { get; set; }
+            }"
+            """
         );
         TestHelper
             .GenerateMapper(source)
@@ -19,6 +28,44 @@ public class ShallowCloneTest
                 var target = new global::A();
                 target.Value = source.Value;
                 target.List = source.List;
+                target.NestedReference = Map(source.NestedReference);
+                return target;
+                """
+            );
+    }
+
+    [Fact]
+    public void ShallowCloneWithNoUserMappingsShouldDirectAssignAllProperties()
+    {
+        var source = TestSourceBuilder.CSharp(
+            """
+            using Riok.Mapperly.Abstractions;
+
+            [Mapper(CloningStrategy = CloningStrategy.ShallowCloning)]
+            public partial class Mapper
+            {
+                [UserMapping(Default = false)]
+                public partial A Map(A source);
+            }
+
+            class A
+            {
+                public int Value { get; set; }
+                public List<string> List { get; set; }
+                public A NestedReference { get; set; }
+            }
+            """
+        );
+
+        TestHelper
+            .GenerateMapper(source)
+            .Should()
+            .HaveSingleMethodBody(
+                """
+                var target = new global::A();
+                target.Value = source.Value;
+                target.List = source.List;
+                target.NestedReference = source.NestedReference;
                 return target;
                 """
             );


samusaran:
@latonz  I would continue the discussion here.

As stated in the github issue #2113, another solution would be to deprecate `UseShallowCloning` in favour of a single `CloningBehaviour` enum, to avoid having conflicting configuration between shallow and deep cloning.

At least from a user perspective, I find this solution to be less confusing.

robertjak:
We are eagerly expecting this feature!

samusaran:
Hi @latonz , sorry for the delay. I've managed to work on the feature but i've found a small issue, but maybe you can help me.

The first iteration (where everything worked fine) used a single attribute on the mapping methods to specify that the mapping should use the Shallow Clone approach.

The thing is that when moving that attribute to the Mapper itself, it's seems to me that it's impossible to distinguish the Shallow Clone from the Deep Clone.

Let's take the following example:

```csharp
class A 
{ 
   public List<string> List { get; set; } 
}

[Mapper(CloningBehaviour = CloningBehaviour.ShallowClone)]
public static partial class ShallowCloningMapper
{
   public static partial A Copy(A src);
}
```

What we are asking is to shallow clone `A`, but since the ShallowClone behaviour is applied at the mapping level, I get "shallow clones" of everything, resulting in a strange deep clone. This is because MappingBuilder iterates on all properties of A.

In this case the resulting mapping is the following:
```csharp
                var target = new global::A();
                target.Value = source.Value;
                target.List = global::System.Linq.Enumerable.ToList(source.List);
                return target;
```

while the expected is
```csharp
                var target = new global::A();
                target.Value = source.Value;
                target.List = source.List;
                return target;
```

Do you have any suggestions on how to handle this?

This all works if I roll back to the previous behaviour of "DeepCloning at mapper level, ShallowCloning at method level"

latonz:
@samusaran I haven't reviewed your code in depth yet, but on a first glance, I think the missing part is checking for the user symbol, see the approach I outlined here https://github.com/riok/mapperly/issues/2113#issuecomment-3754334421 (4. bullet point):
> On a first glance all references to `MappingConfiguration.UseDeepCloning` need to be updated, it probably makes sense to introduce a new `bool MappingBuilderContext.UseCloning => Configuration.UseDeepCloning || (HasUserSymbol && Configuration.UseShallowCloning);`

Adapted to the new design this would be:
```csharp
bool MappingBuilderContext.UseCloning => Configuration.CloningBehaviour == CloningBehaviour.DeepCloning || (HasUserSymbol && Configuration.CloningBehaviour == CloningBehaviour.ShallowCloning);
```

On second thought, I am not sure we should introduce `MapperUseShallowCloningAttribute`. It adds noticeable complexity for limited benefit. If it turns out to be necessary, I think it would be better to introduce it in a later iteration.

samusaran:
Hi @latonz , that you for pointing me to the right `UseCloning` logic. That made the fix.

I'm only left with one deviation from the desired result, which I don't understand how to fix it (due to the mappers being called recursively). I'll point it to the code directly.

src/Riok.Mapperly.Abstractions/MapperAttribute.cs
```diff
@@ -65,8 +65,14 @@ public class MapperAttribute : Attribute
     /// when <c>false</c>, the same array is reused.
     /// when <c>true</c>, the array and each person is cloned.
     /// </summary>
+    [Obsolete("Please use the property CloningBehaviour")]
```
latonz: Could we add `[EditorBrowsable(EditorBrowsableState.Never)]` to this property alongside the Obsolete attribute? Also explain that this property takes precedence over `CloningBehaviour` if set to true (also add this to the xml docs).

```csharp
[EditorBrowsable(EditorBrowsableState.Never)]
[Obsolete("Use 'CloningBehaviour' instead. If this is set to true, `CloningBehaviour.DeepClone` is always used.")]
```

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -96,11 +96,7 @@ public static class EnumerableMappingBuilder
     private static NewInstanceMapping? TryBuildCastMapping(MappingBuilderContext ctx, ITypeMapping elementMapping)
     {
         // cannot cast if the method mapping is synthetic, deep clone is enabled or target is an unknown collection
```
latonz: Nit: The comment needs to be updated, it's not only `deep clone` anymore...

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -322,7 +318,7 @@ private static (bool CanMapWithLinq, string? CollectMethod) ResolveCollectMethod
         // if the target is an IEnumerable<T> don't collect at all
         // except deep cloning is enabled.
         var targetIsIEnumerable = ctx.CollectionInfos!.Target.CollectionType == CollectionType.IEnumerable;
-        if (targetIsIEnumerable && !ctx.Configuration.UseDeepCloning)
+        if (targetIsIEnumerable && ctx.Configuration.CloningBehaviour != CloningBehaviour.DeepCloning)
```
latonz: Shouldn't `ctx.UseCloning` be used here? Update the comment explaining why only deep cloning or replace `deep cloning` with just `cloning`.

src/Riok.Mapperly/Descriptors/MappingBuilders/EnumerableMappingBuilder.cs
```diff
@@ -202,7 +198,7 @@ private static INewInstanceMapping BuildArrayToArrayMapping(MappingBuilderContex
         // use a for loop mapping otherwise.
         if (elementMapping.IsSynthetic)
         {
-            return ctx.Configuration.UseDeepCloning
+            return ctx.Configuration.CloningBehaviour == CloningBehaviour.DeepCloning
```
latonz: Shouldn't `ctx.UseCloning` be used here? If not, add a comment explaining why.

src/Riok.Mapperly/Descriptors/MappingBuilders/MemoryMappingBuilder.cs
```diff
@@ -117,23 +117,23 @@ public static class MemoryMappingBuilder
 
     private static NewInstanceMapping? BuildMemoryToArrayMapping(MappingBuilderContext ctx, INewInstanceMapping elementMapping)
     {
-        if (!elementMapping.IsSynthetic || ctx.Configuration.UseDeepCloning)
+        if (!elementMapping.IsSynthetic || ctx.Configuration.CloningBehaviour == CloningBehaviour.DeepCloning)
```
latonz: Shouldn't `ctx.UseCloning` be used here? Update the comment explaining why only deep cloning or replace deep cloning with just cloning.

src/Riok.Mapperly/Descriptors/MappingBuilders/MemoryMappingBuilder.cs
```diff
@@ -117,23 +117,23 @@ public static class MemoryMappingBuilder
...
             return BuildSpanToArrayMethodMapping(ctx, elementMapping);
 
         return new SourceObjectMethodMapping(ctx.Source, ctx.Target, ToArrayMethodName);
     }
 
     private static NewInstanceMapping? BuildMemoryToSpanMapping(MappingBuilderContext ctx, INewInstanceMapping elementMapping)
     {
-        if (!elementMapping.IsSynthetic || ctx.Configuration.UseDeepCloning)
+        if (!elementMapping.IsSynthetic || ctx.Configuration.CloningBehaviour == CloningBehaviour.DeepCloning)
```
latonz: A few code spots still use `ctx.Configuration.CloningBehaviour == CloningBehaviour.DeepCloning` instead of `ctx.UseCloning`. Why is that?

src/Riok.Mapperly/Descriptors/MappingBuilders/SpanMappingBuilder.cs
```diff
@@ -44,7 +44,10 @@ public static class SpanMappingBuilder
             // if the source is Span/ReadOnlySpan or Array and target is Span/ReadOnlySpan
             // and element type is the same, then direct cast
             (CollectionType.Span or CollectionType.ReadOnlySpan or CollectionType.Array, CollectionType.Span or CollectionType.ReadOnlySpan)
-                when elementMapping.IsSynthetic && !ctx.Configuration.UseDeepCloning => new CastMapping(ctx.Source, ctx.Target),
+                when elementMapping.IsSynthetic && ctx.Configuration.CloningBehaviour != CloningBehaviour.DeepCloning => new CastMapping(
```
latonz: Shouldn't `ctx.UseCloning` be used here? Update the comment explaining why only deep cloning or replace deep cloning with just cloning.

src/Riok.Mapperly.Abstractions/CloningBehaviour.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Riok.Mapperly.Abstractions;
+
+/// <summary>
+/// Specifies whether and how to copy objects of the same type and complex types like collections and spans.
+/// </summary>
+public enum CloningBehaviour
```
latonz: After thinking a bit more about the naming, I am not fully convinced that `CloningBehaviour` is the best choice, especially since the `CloningBehavior` versus `CloningBehaviour` spelling is ambiguous.

My current favorite is `CloningStrategy`, since it aligns well with other Mapperly configuration APIs like `StackCloningStrategy`, `IgnoreObsoleteMembersStrategy`, `RequiredMappingStrategy` and others. Another reasonable alternative would be `CloningMode`.

test/Riok.Mapperly.IntegrationTests/ShallowCloningMapperTest.cs
```diff
@@ -0,0 +1,48 @@
...
+        public void RunIdMappingShouldWork()
+        {
+            var source = new IdObject { IdValue = 20 };
+            var copy = ShallowCloningMapper.Copy(source);
+            source.ShouldNotBeSameAs(copy);
+            copy.IdValue.ShouldBe(20);
+        }
+
+        [Fact]
+        public void RunMappingWithMapperAvoidReturningSourceReference()
```
latonz, line 41: Add a test with an assignable child property and check that it is not cloned.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -277,6 +282,7 @@ private MembersMappingConfiguration BuildMembersConfig(MappingConfigurationRefer
             nestedMembersConfigurations,
             ignoreObsolete,
             requiredMapping
+        //useShallowCloning != null
```
latonz: This can be removed, no?

src/Riok.Mapperly/Configuration/MembersMappingConfiguration.cs
```diff
@@ -11,6 +11,7 @@ public record MembersMappingConfiguration(
     IReadOnlyCollection<NestedMembersMappingConfiguration> NestedMappings,
     IgnoreObsoleteMembersStrategy? IgnoreObsoleteMembersStrategy,
     RequiredMappingStrategy? RequiredMappingStrategy
+//CloningBehaviour CloningBehaviour
```
latonz: As we only allow mapper-wide configuration for now, this can be safely removed, no?

src/Riok.Mapperly/Configuration/MembersMappingConfiguration.cs
```diff
@@ -36,6 +37,7 @@ public MembersMappingConfiguration Include(MembersMappingConfiguration? otherCon
             NestedMappings.Concat(otherConfiguration?.NestedMappings ?? []).ToList(),
             IgnoreObsoleteMembersStrategy ?? otherConfiguration?.IgnoreObsoleteMembersStrategy,
             RequiredMappingStrategy ?? otherConfiguration?.RequiredMappingStrategy
+        //CloningBehaviour // TODO what to do in this case?
```
latonz: As we only allow mapper-wide configuration for now (and no mapping-method specific config), this can be safely removed, no?

samusaran:
I've updated the codebase following your feedback.

Also I've found a quick solution to the problem I was describing the last time.

In a shallow clone mapper, setting all mappings to `[UserMapping(Default = false)]` solves all issues, and the code created is an actual shallow clone with all properties copied 1:1 (except init-only Enumerables). You can check [test/Riok.Mapperly.IntegrationTests/Mapper/ShallowCloningMapper.cs](https://github.com/riok/mapperly/pull/2117/changes#diff-3fd005b0a6fe6709fbe9bbf72095b0592aedffb4bfb354ea088387f86f310e32R7) to verify that.

This is great news 🎉

The only thing I expected to use was `AutoUserMappings = false` instead of defining the attribute for every mapping methods but it doesn't seems to work fully.

Let me know your feedback!

test/Riok.Mapperly.Tests/Mapping/ShallowCloneTest.cs
```diff
@@ -0,0 +1,73 @@
...
+            "A",
+            "A",
+            TestSourceBuilderOptions.WithShallowCloning,
+            """
+            class A
+            {
+                public int Value { get; set; }
+                public List<string> List { get; set; }
+                public A NestedReference { get; set; }
+            }"
```
latonz, line 20: The raw string literal contains an extra `"`.

test/Riok.Mapperly.IntegrationTests/ShallowCloningMapperTest.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
```
latonz, line 2: Weird (and probably unnecessary) import?

docs/docs/breaking-changes/5-0.md
```diff
@@ -27,6 +27,18 @@ but have not handled the newly discovered members.
 Ignore these members to restore the previous behavior.
 See the [private members mapping documentation](../configuration/private-members.mdx) for more details.
 
+## `UseDeepCloning` has been deprecated in favour of `CloningStrategy`
+
+To support allow the implementation of shallow cloning, the `UseDeepCloning` property has been deprecated.
```
latonz, line 32: This sentence doesn't sound quite correct...

docs/docs/breaking-changes/5-0.md
```diff
@@ -27,6 +27,18 @@ but have not handled the newly discovered members.
 Ignore these members to restore the previous behavior.
 See the [private members mapping documentation](../configuration/private-members.mdx) for more details.
 
+## `UseDeepCloning` has been deprecated in favour of `CloningStrategy`
```
latonz, line 30: Add this to the bullet point list on the top.

docs/docs/configuration/mapper.mdx
```diff
@@ -35,23 +35,27 @@ Configurations set via `MapperDefaultsAttribute` take precedence over MSBuild pr
...
+By default, Mapperly does not create copies of objects to improve performance.
 If an object can be directly assigned to the target, it will do so
 (eg. if the source and target type are both `Car[]`, the array and its entries will not be cloned).
-To create deep copies, set the `UseDeepCloning` property on the `MapperAttribute` to `true`.
+To create copies, set the `CloningStrategy` property on the `MapperAttribute` to the desired cloning strategy.
 
 ```csharp
 // highlight-start
-[Mapper(UseDeepCloning = true)]
+[Mapper(CloningStrategy = CloningStrategy.DeepCloning or CloningStrategy.ShallowCloning)]
```
latonz, line 45: Instead of using an invalid c# syntax I'd create a new second line using the second option with a `// or` comment in between.

docs/docs/configuration/mapper.mdx
```diff
@@ -35,23 +35,27 @@ Configurations set via `MapperDefaultsAttribute` take precedence over MSBuild pr
...
 // highlight-end
 public partial class CarMapper
 {
   ...
 }
 ```
 
+Using `CloningStrategy.DeepCloning` will perform deep copies, so all reference types will be cloned recursively,
+meanwhile by using `CloningStrategy.ShallowCloning` a new instance will be returned with all the valued copied 1:1 from
+the source object, without performing any cloning on child properties.
```
latonz, line 55: We need a section explaining what happens if there is a user-defined mapping for a member type (which then could probably be deep-cloned, which may or may not be expected by the user). Anyway, this needs some explanation here, also reference `UserMapping(Default = false)` here.

src/Riok.Mapperly.Abstractions/CloningStrategy.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Riok.Mapperly.Abstractions;
+
+/// <summary>
+/// Specifies whether and how to copy objects of the same type and complex types like collections and spans.
+/// </summary>
+public enum CloningStrategy
+{
+    /// <summary>
+    /// Default behaviour, the original instance will be returned
```
latonz, line 9: add a suffix like `if assignable` as this is only the case if it is assignable.

test/Riok.Mapperly.Tests/Mapping/ShallowCloneTest.cs
```diff
@@ -0,0 +1,73 @@
+﻿using Riok.Mapperly.Abstractions;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class ShallowCloneTest
```
latonz, line 5: Add tests for all other mappings affected by `UseCloning`, for example from `A[]` to `A[]` (which should clone the array but not the entries). Colocate these tests to the mapping builder tests they belong to.

latonz:
Closing this for now due to inactivity. Feel free to reopen or comment if there are any updates.

PR #2176, opened 2026-03-06
feat: Justification Property in Ignore Attributes and Diagnostic

### LINKED ISSUE #2172: Justification Property in Ignore Attributes
No direct problem, nothing is broken. I love mapperly.

I would like to document/justify why certain property mappings are ignored when using `[MapperIgnoreSource(nameof(Car.Id))]` and related attributes.

From a code generator (output) point of view there is no API change.

I would like to be able to specify the business case/reason why specific properties are not mapped from source to target or the other way around, possibly by extending the existing mapper attributes; here on the example of [MapperIgnoreSourceAttribute](https://github.com/riok/mapperly/blob/main/src/Riok.Mapperly.Abstractions/MapperIgnoreSourceAttribute.cs)

```diff
public sealed class MapperIgnoreSourceAttribute : Attribute
{
    public MapperIgnoreSourceAttribute(string source)
    {
        Source = source;
    }

    public string Source { get; }

+   public string? Justification { get; set; }
}
```

We can put comments next to the ignored attribute, but that makes it harder/near impossible to enforce.
Another way is to create a custom attribute, something like `IgnoreJustificationAttribute`, which can then be used in code analysis to enforce all (*) ignored properties have a justification. 
\* - Mapping and handling multiple ignored properties might be a bit of a headache from an analyzer POV.

**Additional context**
The usage would be comparable to `JetBrains.Annotations` where we could use freetext to specify why certain properties are not mapped and could build a corresponding analyzer to ensure that all property ignores have a reason why they are not mapped. Currently this information might be lost.

Note: I did not consider the impact this could have in the code-gen itself, when it must look-up metadata for a mapping and if it could cause problems when discovering the attached metadata, if there is suddenly more syntax nodes appearing (the `, Justification = "Must never expose PII from a public endpoint"`)

Note2: This is a very-very nice-to-have feature since there is definately other workaround available; but I think it would be a nice extension to mapperly and I would also be happy to contribute towards the implementation.

OlliMartin:
# Justification Property in Ignore Attributes and Diagnostic

**Note:** This Pull Request is a draft/request for comments. I would like to make sure my changes go into the right direction before developing something that goes against the architecture.

## Description

This Pull Request aims to address the feature request of #2172.

With the addition of the `Justification` property to `MapperIgnoreX` attributes (more on that below) it is possible to implement a diagnostic to enforce developers to document their decision why a mapping should be ignored.

After reviewing the code I identified the following attributes to be relevant for the diagnostic (please correct me if that's incorrect, not complete):
- `MapperIgnoreSourceAttribute`
- `MapperIgnoreTargetAttribute`
- `MapperIgnoreSourceValueAttribute` 
- `MapperIgnoreTargetValueAttribute` 

In this first draft I only included handling for `MapperIgnoreSource` and I would add the handling for the other attributes including additional tests once my approach is confirmed.

Fixes #2172 

## Open Points

- I already adjusted the code (added a Justification) of three `Verify`/snapshot tests because they would write the new diagnostic otherwise. I do not see a good option to globally ignore the `Hidden` diagnostic, as the generator output is just passed along to Verify. I'm not fooling myself, presumably I am missing something.
- Instead of adjusting all existing tests, I added the new diagnostic to the default ignore set (`TestHelperOptions.Default`). Please let me know if that is ok for you. 
- I'm not happy with the naming of `WhereNotNullBy`. My motivation was to align it with the naming of other LINQ methods such as `GroupBy`

## Checklist

### DIFF
--- a/docs/docs/configuration/analyzer-diagnostics/RMG096.mdx
+++ b/docs/docs/configuration/analyzer-diagnostics/RMG096.mdx
@@ -1,0 +1,30 @@
+---
+sidebar_label: RMG096
+description: 'Mapperly analyzer diagnostic RMG096 — Ignored mapping is missing Justification'
+---
+
+# RMG096 — Ignored mapping is missing Justification
+
+If Mapperly detects a mapping ignore without a Justification, `RMG096` is emitted.
+
+The following attributes may raise this diagnostic:
+ 
+- `MapperIgnoreSource`
+- `MapperIgnoreTarget`
+- `MapperIgnoreSourceValue`
+- `MapperIgnoreTargetValue`
+
+To fix this, add a Justification to the attribute, for example:
+
+```csharp
+public partial class Mapper {
+	[MapperIgnoreSource(nameof(SourceClass.IgnoredProperty), Justification = "This source member is not relevant for the target mapping")]
+	public partial TargetClass Map(SourceClass source);
+}
+
+class SourceClass {
+	string IgnoredProperty { get; set; }
+}
+
+class TargetClass { }
+```


test/Riok.Mapperly.Tests/Mapping/DerivedTypeTest.cs
```diff
@@ -228,7 +228,7 @@ public Task WithBaseTypeMapPropertyAndSeparateMethodShouldWork()
             [MapProperty(nameof(A.BaseValueA), nameof(B.BaseValueB)]
             public partial B Map(A src);
 
-            [MapperIgnoreSource(nameof(A.BaseValueA)]
```
OlliMartin, line 231: I stumbled across this while going through the failing snapshot tests:
How/why does this work in the first place, I can't wrap my head around it - it is not valid C# syntax, right? 
There is a missing `)` here, it should be `[MapperIgnoreSource(nameof(A.BaseValueA))]`

docs/docs/configuration/analyzer-diagnostics/RMG096.mdx
```diff
@@ -0,0 +1,30 @@
+---
+sidebar_label: RMG096
+description: 'Mapperly analyzer diagnostic RMG096 — Ignored mapping is missing Justification'
```
latonz, line 3: Nitpick: lowercase `justification`

latonz: Same at several locations in this document.

docs/docs/configuration/analyzer-diagnostics/RMG096.mdx
```diff
@@ -0,0 +1,30 @@
...
+# RMG096 — Ignored mapping is missing Justification
+
+If Mapperly detects a mapping ignore without a Justification, `RMG096` is emitted.
+
+The following attributes may raise this diagnostic:
+ 
+- `MapperIgnoreSource`
+- `MapperIgnoreTarget`
+- `MapperIgnoreSourceValue`
+- `MapperIgnoreTargetValue`
```
latonz, line 15: We should add `MapperIgnoreAttribute`.

src/Riok.Mapperly/Helpers/EnumerableExtensions.cs
```diff
@@ -18,6 +18,14 @@ public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T?> enumerable)
...
     }
 
+    public static IEnumerable<TItem> WhereNotNullBy<TItem, TSelector>(this IEnumerable<TItem> enumerable, Func<TItem, TSelector?> selector)
+        where TSelector : notnull
+    {
+#nullable disable
+        return enumerable.Where(x => selector(x) != null);
+#nullable restore
+    }
+
```
latonz, line 28: What is the benefit of this new extension method? The `WhereNotNull` doesn't accept a parameter and ensures the source enumerable is notnull-annotated after calling the method. This does not work the same with a selector 🤔

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -259,6 +259,15 @@ private MembersMappingConfiguration BuildMembersConfig(MappingConfigurationRefer
             return MapperConfiguration.Members;
         }
 
+        foreach (var missingIgnoreSourceJustificationAttribute in ignoreSourceMemberAttributes.Where(attr => attr.Justification is null))
```
latonz, line 262: When adding support for the other attributes I'd move this into its own method and call it just next to resolving the attributes.

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -808,6 +808,15 @@ public static class DiagnosticDescriptors
         true
     );
 
+    public static readonly DiagnosticDescriptor MapperIgnoreAttributeMissingJustification = new(
+        "RMG096",
+        "Ignored mapping is missing Justification",
```
latonz, line 813: nitpick: I'd lowercase the `justification`.

src/Riok.Mapperly/Configuration/MapperConfigurationReader.cs
```diff
@@ -259,6 +259,15 @@ private MembersMappingConfiguration BuildMembersConfig(MappingConfigurationRefer
...
+        {
+            _diagnostics.ReportDiagnostic(
+                DiagnosticDescriptors.MapperIgnoreAttributeMissingJustification,
+                configRef.Method,
+                missingIgnoreSourceJustificationAttribute.Source
+            );
+        }
```
latonz, line 269: Add a help link for the added docs, see other diagnostics on how to do it.

test/Riok.Mapperly.Tests/Mapping/DerivedTypeTest.cs
latonz: I think the way to go would be disabling the new diagnostic for all tests except where we explicitly want to test it. I think you should be able to control editorsettings configs in the tests using `TestHelperOptions.AnalyzerConfigOptions`

latonz: Oops, my comment does not really answer your question but a question you raised in the PR description. 

It sometimes still works, even if the syntax tree is not parsed 100% correct as the parsed parts may still be enough fot mapperly to work🤷

src/Riok.Mapperly/Helpers/EnumerableExtensions.cs
OlliMartin: Thanks for your review @latonz!
Planning to continue on this tomorrow, but wanted to answer to this thread already.

Got to be honest, I don't know why the old one (`WhereNotNull`) is there, since the Target Property is **not** nullable when I inspect it, so my thought was that it was added because the contract might be violated during runtime (i.e. property is defined as `string Target { get; }` but the target may be null at runtime.

Now I'm guessing it has to do with different Roslyn versions or the like, in which case I don't know a good approach  other than iterating the list twice.. 

`WhereNotNullBy` (after removing the unnecessary `notnull`) to collect the attributes and then calling `WhereNotNull` on the target so that Target is notnull.

latonz: It may just be a leftover of a refactoring…

latonz:
Closing this for now due to inactivity. Feel free to reopen or comment if there are any updates.

Issue #1611, opened 2024-11-26
Invalid mapping for nullable arrays


TonEnfer:
**Please do the checklist before filing an issue:**
- [ ] I have read the [documentation](https://mapperly.riok.app), including the [FAQ](https://mapperly.riok.app/docs/getting-started/faq/)
- [ ] I can reproduce the bug using the latest prerelease version
- [ ] I have searched existing [discussion](https://github.com/riok/mapperly/discussions) and [issue](https://github.com/riok/mapperly/issues) to avoid duplicates

**Describe the bug**
Mapperly generates incorrect code for nullable arrays in some cases. See example code

**Declaration code**
```csharp
public record Dto(int[]? Array);

[Mapper]
public partial class Mapper
{
    public partial Dto Map(int[]? array);
}
```

**Actual relevant generated code**
```csharp
// <auto-generated />
#nullable enable
namespace ConsoleApp1
{
    public partial class Mapper
    {
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.1.1.0")]
        public partial global::ConsoleApp1.Dto Map(int[]? array)
        {
            return array == null ? throw new System.ArgumentNullException(nameof(array)) : new global::ConsoleApp1.Dto(array);
        }
    }
}
```

**Expected relevant generated code**
```csharp
// <auto-generated />
#nullable enable
namespace ConsoleApp1
{
    public partial class Mapper
    {
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.1.1.0")]
        public partial global::ConsoleApp1.Dto Map(int[]? array)
        {
            return new global::ConsoleApp1.Dto(array);
        }
    }
}
```

**Environment (please complete the following information):**
- Mapperly Version: [e.g. 4.1.1.0]
- Nullable reference types: enabled
- .NET Version: 9.0.100
- Target Framework: .net9.0
- Compiler Version: 4.12.0-3.24558.5 (21192bfc)
- C# Language Version: 13.0
- IDE: Visual Studio v17.5.2
- OS: Win 11

latonz:
Thank you for reporting this issue! It seems there are some bugs related to our null handling. I've created [issue #1614](https://github.com/riok/mapperly/issues/1614) to address this and explore ways to improve our approach to null handling in general.

PR #2180, opened 2026-03-08
fix: skip null guard for ctor mappings when parameter accepts nullable source

baracchande:
# Skip null guard for ctor mappings when parameter accepts nullable source

## Description

Fixes incorrect null guard generation when mapping a nullable source type to a non-nullable target whose single-argument constructor accepts the nullable type.

Fixes one of the bugs reported in [issue 1611](https://github.com/riok/mapperly/issues/1611):

e.g. `int[]? → record Dto(int[]? Array)`

Previously generated `model == null ? throw new ArgumentNullException(...) : new Dto(model)`

Now generates `return new Dto(model)` directly.

## Checklist

### DIFF
--- a/test/Riok.Mapperly.Tests/Mapping/CtorTest.cs
+++ b/test/Riok.Mapperly.Tests/Mapping/CtorTest.cs
@@ -172,7 +172,9 @@ public class CtorTest
                 "Mapping the nullable source of type int[]? to target of type A which is not nullable"
             )
             .HaveAssertedAllDiagnostics()
-            .HaveSingleMethodBody("return source == null ? throw new global::System.ArgumentNullException(nameof(source)) : new global::A(source);");
+            .HaveSingleMethodBody(
+                "return source == null ? throw new global::System.ArgumentNullException(nameof(source)) : new global::A(source);"
+            );
     }
 
     [Fact]


src/Riok.Mapperly/Descriptors/MappingBuilders/NullableMappingBuilder.cs
```diff
@@ -45,6 +58,23 @@ private static bool TryBuildNonNullableMappingKey(MappingBuilderContext ctx, out
...
+        INewInstanceMapping delegateMapping
+    )
+    {
+        if (
+            delegateMapping is CtorMapping ctorMapping
+            && ctx.Target is INamedTypeSymbol namedTarget
+            && CtorMappingBuilder.CtorAcceptsSourceType(namedTarget, ctx.Source, ctx.SymbolAccessor)
+        )
+        {
+            return new CtorMapping(ctx.Source, ctx.Target, ctorMapping.Constructor);
```
latonz: This doesn't really fit Mapperly's architecture. Usually only the specific mapping builder builds the mappings it is designated to.
The problem here is that the `CtorMappingBuilder` comes after the nullable mapping builder in `MappingBuilder._builders`. Another approach would be to add a new `CtorMappingBuilder.TryBuildNullable` and add it to the chain before the nullable builder. However we would need to check the consequences of this carefully.  I'm not sure whether we should move this into a bigger refactoring on how we handle nullability (https://github.com/riok/mapperly/issues/1614)…

latonz:
This is a breaking change, therefore the it needs to be documented in `docs/docs/breaking-changes/5-0.md`

latonz:
Thank you for this great contribution 🥳❤️

PR #2193, opened 2026-03-19
feat: forward additional mapping method parameters

### LINKED ISSUE #1472: Pass additional parameters to property mapper methods

### LINKED ISSUE #1765: Extra mapper parameters do not work with MapDerivedType
- [✅] I have read the [documentation](https://mapperly.riok.app), including the [FAQ](https://mapperly.riok.app/docs/getting-started/faq/)
- [✅] I can reproduce the bug using the latest prerelease version. 
- [✅] I have searched existing [discussion](https://github.com/riok/mapperly/discussions) and [issue](https://github.com/riok/mapperly/issues) to avoid duplicates. 

**Describe the bug**
I need to map a polymorphic item using an additional parameter. It works for concrete types (derivedDto->derived) but if I want to declare it as a polymorphic mapper for the base types and provide the additional parameter, Mapperly generates broken mapping code and does not issue any analyzer warnings

This could actually be #1719 but I am not sure. Though it does not seem like it's supposed to break like that

**Declaration code**
```csharp
[Mapper]
public partial class Mapper
{
    [MapDerivedType<DerivedDto, Derived>]
    public partial Base Map(BaseDto src, int number);
}


public abstract class Base(Guid id, int number)
{
    public Guid Id { get; } = id;
    public int Number { get; } = number;
}

public sealed class Derived(Guid id, int number) : Base(id, number);

public abstract class BaseDto(Guid id)
{
    public Guid Id { get; } = id;
}

public sealed class DerivedDto(Guid id) : BaseDto(id);
```

Note that I've stripped namespaces manually, it's not related to the issue.

See how it tries to use the `number` in `MapToDerived` but it does not actually pass it from the calling method and does not generate a relevant parameter in `MapToDerived`. If I remove `int number` from `Map()` I do get an analyzer error that it could not figure out which constructor to use
```csharp
    public partial class Mapper
    {
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.2.0.0")]
        public partial global::Base Map(global::BaseDto src, int number)
        {
            return src switch
            {
                global::DerivedDto x => MapToDerived(x),
                _ => throw new System.ArgumentException($"Cannot map {src.GetType()} to Base as there is no known derived type mapping", nameof(src)),
            };
        }

        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.2.0.0")]
        private global::Derived MapToDerived(global::DerivedDto source)
        {
            var target = new global::Derived(source.Id, number);
            return target;
        }
    }
```

```csharp
    public partial class Mapper
    {
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.2.0.0")]
        public partial global::Base Map(global::BaseDto src, int number)
        {
            return src switch
            {
                global::DerivedDto x => MapToDerived(x, number),
                _ => throw new System.ArgumentException($"Cannot map {src.GetType()} to Base as there is no known derived type mapping", nameof(src)),
            };
        }

        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.2.0.0")]
        private global::Derived MapToDerived(global::DerivedDto source, int number)
        {
            var target = new global::Derived(source.Id, number);
            return target;
        }
    }
```

- Mapperly Version: 4.2.0
- .NET Version: .NET 9.0.101
- Target Framework: .net8.0
- Compiler Version: 4.12.0-3.24570.6 (913fb4ec)
- C# Language Version: 12.0
- IDE: Rider  2024.3.3
- OS: Windows 22H2

**Additional context**
The real life context is as follows. I have an ordered array of polymorphic items that have some props. Since relational DBs can't preserve the order, I need to convert it to an array of "DB-ready" items which differ in that they have a number representing their position in the array so it can be later reconstructed in the same order.

Demivan:
## Description

Additional parameters on mapping methods (beyond source/target/referenceHandler) are now forwarded to user-defined nested mapping methods, `MapValue(Use=...)`, `MapProperty(Use=...)`, and `MapPropertyFromSource(Use=...)` methods.  
Works with queryable projections - additional parameters become captured variables in the generated expression.

Parameters are matched by name, but mappings to map them when types don't match are not generated. It is more complex to support that. Should I add type checks while that is not supported? Currently, parameter `string ctx` will match `int ctx` but will produce a non-compilable code.

Besides unit tests, I have tested it on real project and parameterized mapping methods and projections were generated correctly.

Note: I have used Claude, but I was steering is constantly - no vibecoding

Fixes #1472
Fixes #1765 (user needs to define mappings with parameters)

Supersedes #1946

## Checklist


src/Riok.Mapperly/Descriptors/ParameterScope.cs
```diff
@@ -0,0 +1,117 @@
+using Microsoft.CodeAnalysis;
+using Riok.Mapperly.Descriptors.Mappings;
+using Riok.Mapperly.Symbols;
+
+namespace Riok.Mapperly.Descriptors;
+
+public class ParameterScope
+{
+    private static readonly IReadOnlyDictionary<string, MethodParameter> EmptyParameters = new Dictionary<string, MethodParameter>();
```
latonz: This codebase names private static readonly fields as lower camel case with `_` prefix: `_emptyParameters`

src/Riok.Mapperly/Descriptors/ParameterScope.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    public ParameterScope(IReadOnlyCollection<MethodParameter> parameters)
+    {
+        if (parameters.Count == 0)
+        {
+            _parameters = EmptyParameters;
+            return;
+        }
+
+        _parameters = parameters.ToDictionary(p => p.NormalizedName, p => p, StringComparer.OrdinalIgnoreCase);
```
latonz, line 25: This could result in runtime exceptions, no (case sensitivity)?

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedInlinedExpressionMapping.cs
```diff
@@ -119,6 +119,19 @@ public override SyntaxNode VisitParenthesizedLambdaExpression(ParenthesizedLambd
...
+            // Note: ctx.AdditionalParameters uses case-insensitive keys (for descriptor-phase matching),
+            // but syntax replacement must be case-sensitive to avoid replacing unrelated identifiers
+            // (e.g., property name CurrentUserId vs parameter currentUserId).
+            if (ctx.AdditionalParameters != null)
+            {
+                foreach (var (key, value) in ctx.AdditionalParameters)
+                {
+                    if (string.Equals(key, node.Identifier.Text, StringComparison.Ordinal))
+                        return value.WithTriviaFrom(node);
+                }
```
latonz: Couldn't this be optimized significantly by first checking the dictionary?

src/Riok.Mapperly/Descriptors/Mappings/TypeMappingBuildContext.cs
```diff
@@ -75,9 +95,47 @@ SyntaxFactoryHelper syntaxFactory
...
+        {
+            foreach (var param in method.Parameters)
+            {
+                if (param.Ordinal == sourceParameter.Ordinal)
+                    yield return sourceParameter.WithArgument(source);
+                else if (targetArgument is not null && param.Ordinal == targetArgument.Value.Parameter.Ordinal)
+                    yield return targetArgument.Value;
+                else if (referenceHandlerParameter is not null && param.Ordinal == referenceHandlerParameter.Value.Ordinal)
+                    yield return referenceHandlerParameter.Value.WithArgument(refHandler);
+                else if (additionalParams?.TryGetValue(param.Name.TrimStart('@'), out var expr) == true)
```
latonz: The `TrimStart('@')` is duplicated across multiple code locations, should be centralized.

src/Riok.Mapperly/Descriptors/MappingCollection.cs
```diff
@@ -201,17 +203,27 @@ private class MappingCollectionInstance<T, TUserMapping>
...
+        /// for type-pairs which are actually in use and not referenced by name.
+        /// Within each group, the first mapping (which won the _defaultMappings race) is excluded.
+        /// </summary>
         public IEnumerable<TUserMapping> UsedDuplicatedNonDefaultNonReferencedUserMappings =>
-            _usedMappingKeys.SelectMany(_duplicatedNonDefaultUserMappings.GetOrEmpty).Where(x => !_referencedNamedMappings.Contains(x));
+            _userMappings
+                .AsDictionary()
+                .Where(g => _usedMappingKeys.Contains(g.Key) && !_explicitDefaultMappingKeys.Contains(g.Key))
+                .Select(g => g.Value.Where(m => !m.IsExternal && !m.Default.HasValue))
+                .Where(g => g.Count() > 1)
```
latonz: Is there a reason for counting all items? Why not simply using `g.Skip(1).Any()`

docs/docs/configuration/additional-mapping-parameters.mdx
```diff
@@ -61,10 +61,117 @@ but higher than a by-name matched regular member mapping.
...
+  // highlight-start
+  public static partial IQueryable<OrderDto> ProjectToDto(
+    this IQueryable<Order> q,
+    int currentUserId
+  );
+  // highlight-end
+}
+
+// Generated:
+// source.Select(x => new OrderDto()
```
latonz: Split generated and declaration into tabs as with other samples.

src/Riok.Mapperly/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -808,6 +808,24 @@ public static class DiagnosticDescriptors
...
+    );
+
+    public static readonly DiagnosticDescriptor NamedMappingParametersUnsatisfied = new(
+        "RMG097",
+        "Named mapping additional parameters cannot be satisfied",
+        "The named mapping {0} has additional parameters that cannot be matched from the caller's scope",
+        DiagnosticCategories.Mapper,
+        DiagnosticSeverity.Error,
+        true
+    );
```
latonz: Add the help links, see `RMG060` for example.

src/Riok.Mapperly/Descriptors/ParameterScope.cs
```diff
@@ -0,0 +1,117 @@
...
+    public bool IsRoot => _parent == null && _usedParameters != null;
+
+    public bool IsEmpty => _parameters.Count == 0;
+
+    public IReadOnlyDictionary<string, MethodParameter> Parameters => _parameters;
+
+    /// <summary>
+    /// Checks if all requested additional parameters can be satisfied by this scope (by normalized name).
+    /// </summary>
+    public bool CanMatchParameters(IReadOnlyCollection<MethodParameter> requested) =>
```
latonz: This is only used internally in this class, probably set its visibility to private?

src/Riok.Mapperly/Descriptors/MappingBuilders/InlineExpressionMappingBuilder.cs
```diff
@@ -61,7 +61,7 @@ ITypeSymbol targetType
 
         var methodSyntax = methodSyntaxRef.GetSyntax();
 
-        if (methodSyntax is not MethodDeclarationSyntax { ParameterList.Parameters: [var sourceParameter] } methodDeclaration)
+        if (methodSyntax is not MethodDeclarationSyntax { ParameterList.Parameters: [var sourceParameter, ..] } methodDeclaration)
```
latonz, line 64: I don't think this syntax is compatible with the source generator requirements, is it?

test/Riok.Mapperly.Tests/Mapping/UserMethodAdditionalParameterForwardingTest.cs
```diff
@@ -0,0 +1,879 @@
...
+                var target = new global::B();
+                target.StringValue = src.StringValue;
+                target.IntValue = Combine(first, second);
+                return target;
+                """
+            );
+    }
+
+    [Fact]
+    public void MapValueUseMethodWithZeroParamsStillWorks()
```
latonz: Nit: Remove the name fragment "Still"? Using 'Still' implies we're testing a change relative to the past, but the test should really just describe the expected state of the system.

latonz (changes requested):
Thanks for the pull request, and for your patience. I was a bit out during the Easter holidays, but I’ve now added my feedback.

src/Riok.Mapperly/Descriptors/MappingBuilders/InlineExpressionMappingBuilder.cs
Demivan: It is a part of the same C# 11 spec as `[var sourceParameter]` pattern, so it should be compatible

latonz: Good point, but I’m not sure about the runtime requirements. The spread operator might need different requirements compared to the a singleton list collection expression.

To be safe, we should add an integration test covering this line. The integration tests run on all supported target frameworks.

src/Riok.Mapperly/Descriptors/ParameterScope.cs
Demivan: What should `Map(Order source, int UserId, int userId)` do?
Separate diagnostic or silently ignore one of the parameters?

latonz: IMO best would be both: an error diagnostic and only the first one taken into account.

Demivan:
@latonz, I have fixed the review comments and rebased on the latest main

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
```diff
@@ -64,7 +57,20 @@ p.Type.TypeKind is TypeKind.TypeParameter or TypeKind.Error
         }
 
         var additionalParameters = additionalParameterSymbols.Select(p => ctx.SymbolAccessor.WrapMethodParameter(p)).ToList();
-        parameters = new MappingMethodParameters(sourceParameter.Value, targetParameter, refHandlerParameter, additionalParameters);
+
+        // detect and deduplicate case-insensitive duplicate additional parameter names (e.g., int UserId, int userId)
+        var parameterGroups = additionalParameters.ToLookup(p => p.NormalizedName, StringComparer.OrdinalIgnoreCase);
+        foreach (var group in parameterGroups.Where(g => g.Count() > 1))
```
latonz: No need to count all entries, is there? `g.Skip(1).Any()` should be sufficient, no?

latonz (changes requested):
Thank you for the updates! One nitpick left, then it looks good to merge 😊

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
Demivan: Fixed. Lookup is using Grouping that is ICollection, so Count is O(1). But let's be consistent

latonz: You are correct, I didn't know that!

latonz:
Thank you for this contribution to Mapperly 🥳

latonz (approved):

PR #2240, opened 2026-04-14
feat: add [MapperNoInlining] attribute and MapperAttribute.NoInlining property

trejjam:
# feat: add `[MapperNoExpressionInliningAttribute]` attribute and `MapperAttribute.NoExpressionInlining` property

## Description

When a mapper referenced via `[UseStaticMapper]` is used from a mapper that has IQueryable projection methods, Mapperly attempts to inline or rebuild the referenced mapping methods into expression trees. For enum mappings configured with `EnumMappingStrategy.ByName` where the source and target enums have different underlying types, this rebuild in expression context produces false `RMG037`/`RMG038` diagnostics — reporting all enum members as unmapped even though they match by name.

I understand the rationale behind the expression-context restrictions for the general case, and the 5.x alpha series has been catching up and tightening these rules further, which makes sense. However, in my project I was relying on `[UseStaticMapper]` with enum-by-name mappings in queryable projections, and the stricter inlining behavior broke that usage. We've discussed the projection + enum + inlining interaction before.

This PR adds a general opt-out mechanism for expression-tree inlining:

- **`[MapperNoInlining]`** — method-level attribute to prevent a specific mapping method from being inlined/rebuilt in expression context
- **`[Mapper(NoInlining = true)]`** — mapper-level property to prevent all methods of a mapper from being inlined when referenced via `[UseStaticMapper]`

Both are applied on the **callee** mapper (the one being referenced), not the consuming mapper.

### Example

```csharp
// Per-method opt-out
public static partial class EnumMapper
{
    [MapperNoInlining]
    public static partial TargetStatus MapToStatus(this SourceStatus source);
}

// Or per-mapper opt-out
[Mapper(EnumMappingStrategy = EnumMappingStrategy.ByName, NoExpressionInlining = true)]
public static partial class EnumMapper
{
    public static partial TargetStatus MapToStatus(this SourceStatus source);
}
```

When inlining is disabled, the method call is preserved as-is in the expression tree. The ORM evaluates it client-side instead of attempting to translate it to SQL.

### Changes

- New `MapperNoExpressionInliningAttribute` marker attribute (`AttributeTargets.Method`)
- New `NoExpressionInlining` property on `MapperAttribute` (default: `false`)
- Check in `InlineExpressionMappingBuilderContext.TryInlineMapping()` before dispatching to inline/rebuild
- Configuration pipeline wired through `MapperConfiguration` and `MapperConfigurationMerger`
- Tests covering: bug repro, method-level attribute, mapper-level property, combined, non-projection (no effect)
- Documentation in queryable projections guide with runtime behavior comparison

> **Disclosure:** This PR was produced with AI assistance. All code has been manually reviewed, verified, and iteratively refined by me to ensure correctness and adherence to project guidelines.

## Checklist


src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
```diff
@@ -216,6 +219,19 @@ private INewInstanceMapping TryInlineMapping(INewInstanceMapping mapping)
         };
     }
 
+    private bool ShouldSkipInlining(IMethodSymbol method)
+    {
+        if (SymbolAccessor.HasAttribute<MapperNoInliningAttribute>(method))
+            return true;
+
+        var containingType = method.ContainingType;
+        if (containingType == null)
```
latonz: Can the containing type of a method ever be null?

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
```diff
@@ -199,6 +199,9 @@ protected override MappingBuilderContext ContextForMapping(
 
     private INewInstanceMapping TryInlineMapping(INewInstanceMapping mapping)
     {
+        if (mapping is IUserMapping userMappingCheck && ShouldSkipInlining(userMappingCheck.Method))
```
latonz: Nit: why the `check` suffix in the name? Wouln'd `userMapping` be sufficient?

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
```diff
@@ -216,6 +219,19 @@ private INewInstanceMapping TryInlineMapping(INewInstanceMapping mapping)
         };
     }
 
+    private bool ShouldSkipInlining(IMethodSymbol method)
+    {
+        if (SymbolAccessor.HasAttribute<MapperNoInliningAttribute>(method))
```
latonz: Mapperly usually only reads configuration through `ctx.Configuration` / from the mapping itself. IMO the correct way here would be passing it through the mapping. `ctx.Configuration` is available in the user mapping extractor.

test/Riok.Mapperly.Tests/Mapping/UseStaticMapperTest.cs
```diff
@@ -773,4 +773,426 @@ public static partial class Mapper
...
+
+                public static partial Domain MapToDomain(this Entity entity);
+            }
+            """
+        );
+
+        TestHelper
+            .GenerateMapper(source, TestHelperOptions.AllowDiagnostics)
+            .Should()
+            .HaveDiagnostics(
```
latonz: This should result in `RMG032`, no? IMO this is somewhat the root cause. We should point to the new `NoInliningAttribute` in the help of `RMG032`.

latonz (changes requested):

test/Riok.Mapperly.Tests/Mapping/UseStaticMapperTest.cs
trejjam: That's possible. If I understand the flow correctly, since `RMG032` is only a warning. It should produce some code. But since the `EnumToEnumMappingBuilder` has difficulty producing a mapping as a fallback, we end up with Source/TargetNotMapped.

I can change `RMG032` to an error -> no fallback to the `EnumToEnumMappingBuilder`, and propose using NoInlining. Is this ok for you?

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
trejjam: It collided with the switch `IUserMapping userMapping`. I moved this as the first branch of the switch.

src/Riok.Mapperly/Descriptors/InlineExpressionMappingBuilderContext.cs
trejjam: You are right, it can never happen.

test/Riok.Mapperly.Tests/Mapping/UseStaticMapperTest.cs
latonz: Sgtm. Please note this as a breaking change in the 5.0 migration doc.

docs/docs/configuration/mapper.mdx
```diff
@@ -70,6 +70,20 @@ public partial class MyMapper
 }
 ```
 
+### Inlining
```
latonz: I'd move this into the queryable projection documentation page.

src/Riok.Mapperly.Abstractions/MapperAttribute.cs
```diff
@@ -142,4 +142,13 @@ public class MapperAttribute : Attribute
...
     public bool AutoUserMappings { get; set; } = true;
+
+    /// <summary>
+    /// Whether to prevent mapping methods of this mapper from being inlined
+    /// into expression trees for queryable projection mappings.
+    /// When <c>true</c>, methods from this mapper referenced via <see cref="UseStaticMapperAttribute"/>
+    /// will not be inlined or rebuilt in expression context.
+    /// Defaults to <c>false</c>.
+    /// </summary>
+    public bool NoInlining { get; set; }
```
latonz: IMO we should include the fact that this only applies to expressions / queryable projection mappings in the name. The same for the separate attribute.

latonz (changes requested):

test/Riok.Mapperly.Tests/Mapping/QueryableProjectionEnumTest.cs
```diff
@@ -41,7 +41,16 @@ TestSourceBuilderOptions.Default with
             .Should()
             .HaveDiagnostic(
                 DiagnosticDescriptors.EnumMappingNotSupportedInProjectionMappings,
-                "The enum mapping strategy ByName, ByValueCheckDefined, explicit enum mappings and ignored enum values cannot be used in projection mappings to map from C to D"
+                "The enum mapping strategy ByName, ByValueCheckDefined, explicit enum mappings and ignored enum values cannot be used in projection mappings to map from C to D, consider applying [MapperNoExpressionInlining] to the mapping method or Mapper(NoExpressionInlining = true) to the containing mapper"
+            )
+            .HaveDiagnostic(DiagnosticDescriptors.CouldNotMapMember, "Could not map member A.Value of type C to B.Value of type D")
```
trejjam, line 46: I am not sure if this is expected behavior or not.  
When I was looking for a similar tests, I found `MapPropertyFromSourceUseWithUnsatisfiableParametersShouldDiagnostic`. Is this cascade expected, or should we prevent it?

trejjam: If I am not mistaken, I saw Mapperly produce an exception for such occasions. That silences the `CouldNotMapMember`.

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
```diff
@@ -464,6 +488,20 @@ out bool hasAttribute
...
+    {
+        if (ctx.SymbolAccessor.HasAttribute<MapperNoExpressionInliningAttribute>(method))
+            return true;
+
+        if (!isExternal)
+            return ctx.Configuration.Mapper.NoExpressionInlining;
+
+        // the configuration of external mappers is not merged into ctx.Configuration,
+        // read the MapperAttribute of the containing type instead
+        var mapperAttribute = ctx.AttributeAccessor.AccessFirstOrDefault<MapperAttribute>(method.ContainingType);
```
trejjam: I hope the `ctx.SymbolAccessor.HasAttribute` is a sufficient solution for the `Mapperly usually only reads configuration through ctx.Configuration`. From what I found, it's cached, so it should not be a performance issue.

When constructing Mapper -> AnotherMapper -> Method. We need to get info about the `AnotherMapper`. I did not find better way than `ctx.AttributeAccessor.AccessFirstOrDefault`. If you want it differently, can you point me in the right way?

trejjam:
Rebased on main

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
latonz: This is not sufficient as it does not consider the `MapperDefaultsAttribute`, does it?
Probably just always use `ctx.Configuration.Mapper.NoExpressionInlining`?

docs/docs/configuration/queryable-projections.mdx
```diff
@@ -122,6 +122,112 @@ public static partial class CarMapper
...
+    // highlight-start
+    .AsEnumerable() // ← everything before runs in SQL, everything after is client-side
+    // highlight-end
+    .Select(x => new CarDto
+    {
+        // highlight-start
+        Status = EnumMapper.MapToStatus(x.Status), // ← runs in C#
+        // highlight-end
+    })
+    .ToListAsync();
```
latonz, line 218: This doesn't work does it? After `AsEnumerable` it is of type `IEnumerable<T>` which doesn't have a `ToListAsync` (or at least not as a built in method or extension method by ef core)…

test/Riok.Mapperly.Tests/Mapping/QueryableProjectionEnumTest.cs
latonz: would keep this unless we decide to change diagnostic suppression generally, not as part of this PR.

latonz (changes requested):

docs/docs/configuration/queryable-projections.mdx
trejjam: In older C# you need to use other implementation. But since net 10 it's there: https://learn.microsoft.com/en-us/dotnet/api/system.linq.asyncenumerable.tolistasync

trejjam: Ah, you are right. It should have been `.AsAsyncEnumerable()`

src/Riok.Mapperly/Descriptors/UserMethodMappingExtractor.cs
trejjam: I have added `MapperAttribute BuildMapperConfiguration(ISymbol mapperSymbol)` which builds configuration with defaults.
The `ctx.Configuration.Mapper.NoExpressionInlining` is not possible, as it's related to the current mapper and not to the targeted mapper.
If this is an issue, it can be possible to move the `NoExpressionInlining` from the `MapperAttribute` into the `UseStaticMapper`.

trejjam:
I have created a modified version with configuration on the `UseStaticMapper`:
https://github.com/trejjam/riok-mapperly/pull/1
if you would prefer it.

This also let you control not-inlining of not-owned code

trejjam:
Hi, can I do something to finalize this PR?
While working on it, I found this variant better:
```cs
[Mapper]
[UseStaticMapper(typeof(EnumMapper), NoExpressionInlining = true)]
// or:
[UseStaticMapper<EnumMapper>(NoExpressionInlining = true)]
partial class Mapper{}
```
It replaces the:
```cs
[Mapper]
[UseStaticMapper(typeof(EnumMapper))]
partial class Mapper {}

[Mapper(NoExpressionInlining = true)]
partial class EnumMapper {}
```
What is your opinion?

The code for it is here:
https://github.com/trejjam/riok-mapperly/pull/1 I can recreate this PR with the `[UseStaticMapper(typeof(EnumMapper), NoExpressionInlining = true)]` if you also prefer it.

latonz:
@trejjam, this would only allow the setting to be configured for imported mappers, correct? I think we need to support both: setting it on the mapper itself and overriding it for an imported mapper via UseStaticMapper.

IMO, the biggest issue with the current implementation is that the setting isn’t resolved correctly from imported mappers. Adding it to `UseStaticMapper`/`UseMapper` would fix that, but it shouldn’t replace the existing mapper-level setting, should it? This setting is also helpful within the same mapper…

trejjam:
Hello @latonz 
I hope I understand. I will try to create a little specification:

The new `[MapperNoExpressionInlining]` on methods. This prevents its mapper class from inlining its body:
```cs
record Source(ESource Value);
record Target(ETarget Value);

record Source2(ESource Value);
record Target2(ETarget Value);

[Mapper]
partial class MapperWithNoExprInlining
{
  [MapperNoExpressionInlining]
  partial ETarget MapToETarget(ESource source);

  partial Target MapToTarget(Source source); // no - inlining allowed for MapToETarget
}


[Mapper]
[UseStaticMapper(typeof(MapperWithNoExprInlining))]
partial class Mapper
{
  partial Target2 MapToTarget(Source2 source); // inlined - the inlining is blocked only in the owning MapperWithNoExprInlining
}
```

___

The `[Mapper(NoExpressionInlining = true)]` blocks inlining for all "owned" methods:
```cs
record Source(ESource Value);
record Target(ETarget Value);

record Source2(ESource Value);
record Target2(ETarget Value);

[Mapper(NoExpressionInlining = true)]
partial class MapperWithNoExprInlining
{
  partial ETarget MapToETarget(ESource source);

  partial Target MapToTarget(Source source); // no - inlining allowed for MapToETarget
}


[Mapper]
[UseStaticMapper(typeof(MapperWithNoExprInlining))]
partial class Mapper
{
  partial Target2 MapToTarget(Source2 source); // inlined - the inlining is blocked only in the owning MapperWithNoExprInlining
}
```

___

The `[UseStaticMapper(typeof(...), NoExpressionInlining = true)]` blocks inlining for all imported methods:
```cs
record Source(ESource Value);
record Target(ETarget Value);

record Source2(ESource Value);
record Target2(ETarget Value);

[Mapper]
partial class SourcedMapper
{
  partial ETarget MapToETarget(ESource source);

  partial Target MapToTarget(Source source); // inlining allowed for MapToETarget
}


[Mapper]
[UseStaticMapper(typeof(SourcedMapper), NoExpressionInlining = true)]
partial class Mapper
{
  partial Target2 MapToTarget(Source2 source); // MapToETarget is not inlined - the inlining is blocked for SourcedMapper imported methods
}
```

___

Do you like it to work this way? It should fix the not-scoped configuration read

latonz:
https://github.com/riok/mapperly/pull/2240#issuecomment-5584430974

Thanks for writing up the spec. I don’t particularly like this behavior from a user’s perspective, as it feels a bit unintuitive, but I think it’s the only way to support all three scopes and still resolving the configuration correctly…

trejjam:
I agree, I will modify the PR according to it when I have a spare moment.
Feel free to change the spec in the meantime if you find something better.

latonz (changes requested):
Implementation of latest spec proposal…

Issue #1311, opened 2024-05-28
Map to nested existing target instances


latonz:
Nested objects in existing target mappings should preserve the existing instance if possible. Currently always a new instance of the nested object is created.

Rel. #1310

ste4net:
Any update on this? It is really important for me. 
I am trying to implement mapping for objects like 
```
public class Company{
 ...
  public List<Address> Addresses{get;set;}
}
```
having the possibility to update list of entities when entity exist in source and target list, remove entities from target if not present in source and ad entities if added in source. 
Without this i am not able to find a way to map enumerable properties because i have this approach for all List<T> properties. 
I guess the only way should be using MapperIgnore for each enumerable properties and wrap mapper call in custom method to manually update the enumerable child properties but it's a lot of work . I was trying to move from Automapper but this part block mi migration.

Thankyou very much

p.s.
Maybe this issue is related to #1484

JenPullUp:
I'm running into the exact same problem, is there a workaround by using a usermapping? its kind of necessary when updating 'complex' objects in efcore without getting tracking errors

latonz:
I don't believe there's a workaround available at the moment.

kirides:
This issue is really bad for EF Core changetracking.

For us, this means any Relation in an update needs to be manually ignored, synchronized and mapped.
We have two helpers for that.

You can imagine how much clutter this adds just to make sure a user can't override a relation entity ID.
we have entities with many Owned-Properties and Relations, our mapping-files are 90% manually synchronization.

We really like using Mapperly, as it allows compile time checking, evaluation of generated code, is Hot-Reloadable, and is composable.

But some built-in way of doing
```cs
// use IComparable<T>, IEquatable<T> or similar to enable this feature.
[MapComparableProperty(nameof(source.Something), nameof(target.Something))]
// manually specifying an Equality method
[MapComparableProperty(nameof(source.Something), nameof(target.Something), Use = nameof(CompareSourceDestMethod))]
private static partial void MapInto(TheSource source, TheTarget target);

// maybe support generic overloads? idk.
private static bool CompareSourceDestMethod<A, B>(A a, B b)
    where A: IHaveId,
    where B: IHaveId => a.Id == b.Id;

```
would be exceptionally great.

<details>
<summary> Synchronize Property </summary>

```cs
private static void SynchronizeProperty<TSource, TDestination>(TSource? source,
	TDestination? destination,
	Action<TDestination?> setValueAction,
[...]

Issue #276, opened 2023-03-13
Add support for destination value in user implemented mapping methods


furier:
**Is your feature request related to a problem? Please describe.**

Currently, the **User implemented mapping methods** feature in Mapperly only allows users to provide a method body that maps a single source type to a single destination type. However, in some scenarios, it may be necessary to have more control over the mapping process, such as when mapping optional values or when the user wants to choose which value (source or destination) to use for the destination property.

**Describe the solution you'd like**

I propose adding support for destination values in the **User implemented mapping methods** feature. This would allow users to specify which value (source or destination) to use for the destination property when mapping between objects.

For example, the following code snippet demonstrates how this feature might be used to map an optional value:

```cs

[Mapper]
public partial class MyMapper
{
    public partial void MapPersonToPersonDto(Person person, ref PersonDto destination)
    {
        destination.Id = person.Id;
        destination.Name = person.Name;
        destination.Age = FromOptional(person.Age, destination.Age);
    }

    private static int? FromOptional<T>(Optional<int?> source, int? destination) => source.HasValue ? source.Value : destination;
}
```

With this change, users would have more control over how mappings are defined and could handle more complex mapping scenarios that are not possible with the current implementation.

latonz:
I think your use case could be implemented by using the after map feature with an ignored property:

```csharp
[Mapper]
public partial class MyMapper
{
    public void MapPersonToPersonDto(Person person, PersonDto destination)
    {
        MapPersonToPersonDtoInternal(person, destination);
        destination.Age = CustomAgeMapping(person, destination);
    }

    [MapperIgnoreTarget(nameof(PersonDto.Age))]
    private partial void MapPersonToPersonDtoInternal(Person person, PersonDto destination)
}

```

Would this work for your use case?

furier:
It would probably work but that would mean that the mapping was not generated for me and I would have to do that manually? Then the point of a mapper is kinda mute?

EniacMlezi:
This would not work in the case where PersonDto.Age is **`required`** and when using a non "void mapping method" (when not mapping to existing object instance)

```cs
[Mapper]
public partial class MyMapper
{
    public PersonDto MapPersonToPersonDto(Person person)
    {
        var personDto = MapPersonToPersonDtoInternal(person);  
        destination.Age = CustomAgeMapping(person, destination);
    }

    [MapperIgnoreTarget(nameof(PersonDto.Age))]
[...]

PR #2242, opened 2026-04-18
feat: Automatically use ref target property mappings

juliancarrivick:
# Automatically use ref target property mappings 

## Description

This was motivated by wanting to switch from Automapper, where I was able to conditionally update a value based on whether a custom `Optional` class had a value or not. With Mapperly this requires a ref target mapping, however they currently have to be explicitly opted into for every field on the object, which doesn't scale very well. e.g. When you have many DTOs with many properties that need such a mapping from a controller type to the DB entity type. It seemed timely to implement this functionality since 5.0 is in pre-release at the moment.

As a disclosure, I used Claude to generate the initially failing test cases and inspected that they were reasonable, adapting them to use the pre-existing test models. I also used it to generate the initial fix and also adapted that to the existing conventions.

I'm not entirely happy with the fix, it seems to duplicate checks that are performed elsewhere too (e.g. default checking) and having to downcast to see if it is a ref target or not. I investigated doing some refactoring in this space but it got very messy very quickly so I figured that would be better performed by someone more familiar with the code base.

Thank you for this great project, I hope to be able to use it with this change soon!

## Checklist

### DIFF
--- a/docs/docs/configuration/user-implemented-methods.mdx
+++ b/docs/docs/configuration/user-implemented-methods.mdx
@@ -99,17 +99,53 @@ which can be changed using the [`.editorconfig`](./analyzer-diagnostics/index.md
 ## Map properties using user-implemented mappings
 
 See [user-implemented property conversions](./mapper.mdx#user-implemented-property-mappings).
-The mapper will respect the `ref` keyword on the target parameter when using an existing target function, allowing the reference to be updated in the caller.
+
+A `void` method with a `ref` target parameter can be used to convert a property value in-place.
+Mapperly respects the `ref` keyword on the target parameter when using an existing target function,
+allowing the reference to be updated in the caller.
+
+When `AutoUserMappings` is enabled (the default), such methods are discovered and used automatically
+for property mapping — no `[MapProperty(Use = ...)]` attribute is required:
 
 ```csharp
-  [Mapper(AllowNullPropertyAssignment = false, UseDeepCloning = true, RequiredMappingStrategy = RequiredMappingStrategy.Source)]
-  public static partial class UseUserMethodWithRef
-  {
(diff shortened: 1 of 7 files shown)


latonz:
Thank you for this contribution! I'll look into this the coming week.

latonz:
I have some concerns about making this the default behavior… Wouldn't the use cases for this be more robustly handled by the proposal in #276 and #1311?
I'd like to understand the necessity of this change versus the more explicit alternatives.

kammerjaeger:
I agree that some solution would be good.
I basically have the same problem with an Optional class/struct.

I tried generic mappings from https://next.mapperly.riok.app/docs/configuration/generic-mapping/ and explicit method mappings. The current implementation does not support either (without being explicit with MapProperty). If needed I can give examples.
I don't see a reason why not. There is probably some other use case where ref support would also be useful.

But in this case https://github.com/riok/mapperly/issues/276 would probably be the better implementation.
Perfect would be something like this https://next.mapperly.riok.app/docs/configuration/mapper/#null-values for optional types, I'm just not sure how this could work.

juliancarrivick:
Hi @latonz, thanks for taking a look at this.

#276 would indeed cover my use case (and in-fact is how AutoMapper's API looks) - I wasn't aware of that one, thanks for that. I'm not particularly attached to the implementation here so happy for you to close this if you'd prefer something closer to the proposed solution in that issue.

Thanks @kammerjaeger for those additional links - I think the ideal solution would be like #276 with the generic mappings support so a single implementation of `Optional<T> -> T` could be provided.

The null support I did come across, however it doesn't work so well where `Optional.None` is distinct from `Optional.Some(null)` e.g. in the case of a HTTP PATCH where the presence of a property indicates it should be updated and value of the property itself could be null.

latonz:
Closing in favor of https://github.com/riok/mapperly/issues/276

PR #2256, opened 2026-05-01
fix: Add null checks when mapping nullable sources to non-nullable targets with MayBeNull and AllowNull Attribute

### LINKED ISSUE #2192: CS8601 while mapping properties annotated with `MaybeNullAttribute`
- [x] I have read the [documentation](https://mapperly.riok.app), including the [FAQ](https://mapperly.riok.app/docs/getting-started/faq/)
- [x] I can reproduce the bug using the latest prerelease version
- [x] I have searched existing [discussion](https://github.com/riok/mapperly/discussions) and [issue](https://github.com/riok/mapperly/issues) to avoid duplicates

**Describe the bug**

Generated code produces compilation error `CS8601 Possible null reference assignment` when:

- Using nullable reference types
- Using [MaybeNull] attribute

Any option to have the mapper consider the `MaybeNull` attribute in the source property?

**Declaration code**
```csharp
public partial class OriginalType
{
    [MaybeNull]
    public string Name { get; set; } = default!;
}

public partial class MappedType
{
    [MaybeNull]
    public string Name { get; set; }
}

[Mapper]
internal partial class OneMapper
{
    public partial MappedType Map(OriginalType source);
}
```

```csharp
internal partial class OneMapper
{
    [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.3.1.0")]
    public partial global::MapperlyBug.MappedType Map(global::MapperlyBug.OriginalType source)
    {
        var target = new global::MapperlyBug.MappedType();
        target.Name = source.Name;
        return target;
    }
}
```

```csharp
internal partial class OneMapper
{
    [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.3.1.0")]
    public partial global::MapperlyBug.MappedType Map(global::MapperlyBug.OriginalType source)
    {
        var target = new global::MapperlyBug.MappedType();
        target.Name = source.Name!;
        // or
        //if (source.Name != null)
        //{
        //    target.Name = source.Name;
        //}
        return target;
    }
}
```

- Mapperly Version: 4.3.1
- .NET Version: 10.0.201
- Target Framework: net10.0
- C# Language Version: 14.0
- IDE: VS 2026

**Additional context**

Changing the original code to use required properties is not an option because it would change the semantics of a large existing code base (forcing setting a bunch of properties at object initialization).
Additionally, most of the real classes are used as REST payloads (JSON serialization) and are also annotated with the [Required] attribute.
99,9% of the times using the null forgiving operator would be fine because the source of the mapping has already been validated.

Abhijithtv:
# Add null checks when mapping nullable sources to non-nullable targets with MayBeNull Attribute

## Description

For target members, [MaybeNull] should not be considered when determining nullability. 
Otherwise, it may lead to incorrect nullable analysis and result in CS8601 ("Possible null reference assignment") warnings.

Fixes #2192

## Checklist

### DIFF
--- a/docs/docs/breaking-changes/5-0.md
+++ b/docs/docs/breaking-changes/5-0.md
@@ -17,8 +17,8 @@ description: How to upgrade to Mapperly 5.0 and a list of all its breaking chang
 - `Stack<T>` deep cloning now preserves the order of elements by default.
 - Inaccessible members from other assemblies are now included in the mapping process if `IncludedMembers` or `IncludedConstructors` options are configured to include them.
 - Constructor mappings with nullable source types no longer generate null guards when the constructor parameter accepts null, see below.
-- `MaybeNull` attributes are now respected for fields, properties and constructor parameters in read context.
-- `AllowNull` attributes are now respected for fields, properties and constructor parameters in write context.
+- `MaybeNull` attributes are now respected for fields, properties and constructor parameters when reading values.
+- `AllowNull` attributes are now respected for fields, properties and constructor parameters when writing values.
 - Copy constructors are no longer selected over constructors annotated with `[MapperConstructor]` or when `[MapProperty]` configurations are present.
 
 ## Inaccessible members from other assemblies included
--- a/docs/docs/breaking-changes/5-0.md
+++ b/docs/docs/breaking-changes/5-0.md
@@ -17,7 +17,8 @@ description: How to upgrade to Mapperly 5.0 and a list of all its breaking chang
 - `Stack<T>` deep cloning now preserves the order of elements by default.
 - Inaccessible members from other assemblies are now included in the mapping process if `IncludedMembers` or `IncludedConstructors` options are configured to include them.
 - Constructor mappings with nullable source types no longer generate null guards when the constructor parameter accepts null, see below.
-- `MaybeNull` attributes are now respected for fields, properties and constructor parameters.
+- `MaybeNull` attributes are now respected for fields, properties and constructor parameters in read context.
+- `AllowNull` attributes are now respected for fields, properties and constructor parameters in write context.
 - Copy constructors are no longer selected over constructors annotated with `[MapperConstructor]` or when `[MapProperty]` configurations are present.
 
 ## Inaccessible members from other assemblies included
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
@@ -201,7 +201,7 @@ internal static class SourceValueBuilder
             ctx.BuilderContext.ReportDiagnostic(
                 DiagnosticDescriptors.MapValueMethodTypeMismatch,
                 methodReferenceConfiguration.Name,
-                FormattedReturnType(namedMethodCandidates[0]),
+                ctx.BuilderContext.SymbolAccessor.UpgradeReturnNullable(namedMethodCandidates[0]).ToDisplayString(),
                 memberMappingInfo.TargetMember.ToDisplayString()
             );
             sourceValue = null;
@@ -218,18 +218,5 @@ internal static class SourceValueBuilder
             additionalParameterNames
         );
         return true;
-    }
-
-    private static string FormattedReturnType(IMethodSymbol methodSymbol)
-    {
-        bool hasNullableAttribute = methodSymbol
-            .GetReturnTypeAttributes()
-            .Any(a => string.Equals(a.AttributeClass?.Name, nameof(MaybeNullAttribute)));
-
-        var effectiveType = hasNullableAttribute
-            ? methodSymbol.ReturnType.WithNullableAnnotation(NullableAnnotation.Annotated)
-            : methodSymbol.ReturnType;
-
-        return effectiveType.ToDisplayString();
     }
 }
--- a/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
+++ b/src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
@@ -77,7 +77,7 @@ internal static class SourceValueBuilder
         // but the provided value is null or default (for default IsNullable is also true)
         if (
             value.ConstantValue.IsNull
-            && memberMappingInfo.TargetMember.MemberType.IsReferenceType
+            && memberMappingInfo.TargetMember.MemberWriteType.IsReferenceType
             && !memberMappingInfo.TargetMember.Member.IsWriteNullable
         )
         {
@@ -92,8 +92,8 @@ internal static class SourceValueBuilder
         // target is value type but value is null
         if (
             value.ConstantValue.IsNull
-            && memberMappingInfo.TargetMember.MemberType.IsValueType
-            && !memberMappingInfo.TargetMember.MemberType.IsNullableValueType()
+            && memberMappingInfo.TargetMember.MemberWriteType.IsValueType
+            && !memberMappingInfo.TargetMember.MemberWriteType.IsNullableValueType()
             && value.Expression.IsKind(SyntaxKind.NullLiteralExpression)
         )
         {
@@ -116,7 +116,7 @@ internal static class SourceValueBuilder
 
         // use non-nullable target type to allow non-null value type assignments
         // to nullable value types
-        if (!SymbolEqualityComparer.Default.Equals(value.ConstantValue.Type, memberMappingInfo.TargetMember.MemberType.NonNullable()))
+        if (!SymbolEqualityComparer.Default.Equals(value.ConstantValue.Type, memberMappingInfo.TargetMember.MemberWriteType.NonNullable()))
         {
             ctx.BuilderContext.ReportDiagnostic(
                 DiagnosticDescriptors.MapValueTypeMismatch,
@@ -137,7 +137,7 @@ internal static class SourceValueBuilder
                 // expand enum member access to fully qualified identifier
                 // use simple member name approach instead of slower visitor pattern on the expression
                 var enumMemberName = ((MemberAccessExpressionSyntax)value.Expression).Name.Identifier.Text;
-                var enumTypeFullName = FullyQualifiedIdentifier(memberMappingInfo.TargetMember.MemberType.NonNullable());
+                var enumTypeFullName = FullyQualifiedIdentifier(memberMappingInfo.TargetMember.MemberWriteType.NonNullable());
                 sourceValue = new ConstantSourceValue(MemberAccess(enumTypeFullName, enumMemberName));
                 return true;
             case TypedConstantKind.Type:
@@ -186,7 +186,7 @@ internal static class SourceValueBuilder
         // to nullable value types
         // nullable is checked with nullable annotation
         var methodCandidates = namedMethodCandidates.Where(x =>
-            SymbolEqualityComparer.Default.Equals(x.ReturnType.NonNullable(), memberMappingInfo.TargetMember.MemberType.NonNullable())
+            SymbolEqualityComparer.Default.Equals(x.ReturnType.NonNullable(), memberMappingInfo.TargetMember.MemberWriteType.NonNullable())
         );
 
         if (!memberMappingInfo.TargetMember.Member.IsNullable)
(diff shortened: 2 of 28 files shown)


latonz:
Thanks for the contribution! The fix turns out to be a bit more involved than it might initially appear, IMO also my original `MayBeNull` support PR was short-sighted. To handle this correctly, I think we need to support both `MayBeNull` (when reading values) and `AllowNull` (when writing values).

Abhijithtv:
Thanks, that makes sense!

I checked and it looks like AllowNull isn't currently considered in target nullability checks (I also couldn’t find any tests covering it).

Would you prefer this to be handled in a separate PR focusing on AllowNull support?

Also, regarding the implementation: would you recommend modeling read and write nullability as separate properties (e.g., sourceMayBeNull vs targetAcceptsNull), or passing a context flag like isRead/isWrite?

latonz:
As these two strongly belong together and the changes shouldn't get too big, I'd do it in the same PR.
I'd prefer separate properties/methods.

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyValueMethodTest.cs
```diff
@@ -713,14 +711,12 @@ class B
...
-            .HaveSingleMethodBody(
-                """
-                var target = new global::B();
-                target.Value = BuildValue();
-                return target;
-                """
-            );
+            .HaveDiagnostic(
+                DiagnosticDescriptors.MapValueMethodTypeMismatch,
+                "Cannot assign method return type string of BuildValue() to B.Value of type string"
```
latonz: This error message is misleading (string mismatch vs string).

test/Riok.Mapperly.Tests/Mapping/ObjectPropertyValueMethodTest.cs
latonz: The test name no longer matches the assertion — it used to allow the nullable return, it now diagnoses it. Please rename to something like `MethodToMaybeNullTargetPropertyShouldDiagnoseNullableReturnType`.

latonz: Same as above, name says "should map", body now asserts a diagnostic. Please rename and consider whether this case (both sides annotated [MaybeNull]) is what users actually expect to break. Testing [AllowNull] as the way to opt into nullable writes would help here too.

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -92,6 +92,18 @@ private bool IsAccessible(ISymbol symbol, MemberVisibility visibility)
     }
 
     public bool IsNullable(ISymbol symbol)
```
latonz: Three near-identical helpers (IsNullable, IsReadNullable, IsWriteNullable) is a foot-gun, IsNullable now ignores both [MaybeNull] and [AllowNull], so any future call site that picks it without thinking will reintroduce the bug this PR fixes. Could the `IsNullable` methods be made private?

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
```diff
@@ -103,6 +115,18 @@ public bool IsNullable(ISymbol symbol)
         };
     }
 
+    public bool IsWriteNullable(ISymbol symbol)
```
latonz, line 106: The new `[AllowNull]` write-side branch doesn't seem to be covered by any test in this PR. Could you add tests covering:                                                                 
- `[AllowNull]` target property with a nullable source (expect direct assign, no null guard, no diagnostic)                                                                   
- `[AllowNull]` target with a [MaybeNull] source                                                                                                                             
Otherwise half of the semantic change is untested.

src/Riok.Mapperly/Symbols/Members/ParameterSourceMember.cs
```diff
@@ -20,6 +20,8 @@ public class ParameterSourceMember(MethodParameter parameter) : IMappableMember,
     public ITypeSymbol Type => parameter.Type;
     public INamedTypeSymbol? ContainingType => null;
     public bool IsNullable => parameter.Type.IsNullable();
+    public bool IsReadNullable => parameter.Type.IsNullable();
+    public bool IsWriteNullable => parameter.Type.IsNullable();
```
latonz: These ignore `[MaybeNull]` / `[AllowNull]` on the parameter symbol, while `ConstructorParameterMember` correctly delegates to `SymbolAccessor.IsReadNullable`/`IsWriteNullable`…

latonz (changes requested):

Please also update the `5-0.md` breaking changes docs.

src/Riok.Mapperly/Symbols/Members/ParameterSourceMember.cs
```diff
@@ -14,12 +14,15 @@ namespace Riok.Mapperly.Symbols.Members;
...
 /// </summary>
 [DebuggerDisplay("{Name}")]
-public class ParameterSourceMember(MethodParameter parameter) : IMappableMember, IMemberGetter
+public class ParameterSourceMember(MethodParameter parameter, SymbolAccessor symbolAccessor) : IMappableMember, IMemberGetter
 {
     public string Name => parameter.Name;
     public ITypeSymbol Type => parameter.Type;
     public INamedTypeSymbol? ContainingType => null;
-    public bool IsNullable => parameter.Type.IsNullable();
+    public bool IsReadNullable =>
```
Abhijithtv, line 22: For MethodMapping, is it sufficient to rely on the parameter type to determine nullability?

Currently, we do not store the parameter symbol within MethodMapping (I’m not sure whether this is feasible or intended). Could you provide some guidance on how nullability should be handled in this case?

Abhijithtv:
@latonz  Thanks for the review.

 I realized I forgot to mark the PR as a draft earlier. Your feedback was really helpful.
I believe the PR is ready for review now.

src/Riok.Mapperly/Descriptors/SymbolAccessor.cs
Abhijithtv: I have removed the **IsNullable** and replaced all its reverences with **IsReadNullable** and **IsWriteNullable**

docs/docs/breaking-changes/5-0.md
```diff
@@ -17,7 +17,8 @@ description: How to upgrade to Mapperly 5.0 and a list of all its breaking chang
 - `Stack<T>` deep cloning now preserves the order of elements by default.
 - Inaccessible members from other assemblies are now included in the mapping process if `IncludedMembers` or `IncludedConstructors` options are configured to include them.
 - Constructor mappings with nullable source types no longer generate null guards when the constructor parameter accepts null, see below.
-- `MaybeNull` attributes are now respected for fields, properties and constructor parameters.
+- `MaybeNull` attributes are now respected for fields, properties and constructor parameters in read context.
+- `AllowNull` attributes are now respected for fields, properties and constructor parameters in write context.
```
latonz: instead of `in read/write context` IMO `when reading/writing values` may be clearer?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
```diff
@@ -219,4 +219,17 @@ private static bool TryBuildMethodProvidedSourceValue(
...
         return true;
     }
+
+    private static string FormattedReturnType(IMethodSymbol methodSymbol)
+    {
+        bool hasNullableAttribute = methodSymbol
+            .GetReturnTypeAttributes()
+            .Any(a => string.Equals(a.AttributeClass?.Name, nameof(MaybeNullAttribute)));
+
+        var effectiveType = hasNullableAttribute
```
latonz: IMO `SymbolAccessor.TryHasAttribute<MaybeNullAttribute>(methodSymbol.GetReturnTypeAttributes())` should be used.

latonz: this logic is duplicated with the `UserMethodMappingExtractor`… extract into a new helper, e.g. `SymbolAccessor.UpgradeReturnNullable(IMethodSymbol)`

src/Riok.Mapperly/Symbols/MethodParameter.cs
```diff
@@ -3,15 +3,21 @@
 
 namespace Riok.Mapperly.Symbols;
 
-public readonly record struct MethodParameter(int Ordinal, string Name, ITypeSymbol Type, RefKind RefKind = RefKind.None)
+public readonly record struct MethodParameter(
+    int Ordinal,
+    string Name,
+    ITypeSymbol Type,
+    IParameterSymbol? symbol = null,
```
latonz: Why is this lower cased?

src/Riok.Mapperly/Symbols/Members/ParameterSourceMember.cs
latonz: We can't only rely on the type as that would ignore these attributes.

I think storing the symbol within the method parameter should be okay. I'd add helpers though (`IsWriteNullable` and `IsReadNullable`).

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
```diff
@@ -219,4 +219,17 @@ private static bool TryBuildMethodProvidedSourceValue(
         );
         return true;
     }
+
+    private static string FormattedReturnType(IMethodSymbol methodSymbol)
+    {
+        bool hasNullableAttribute = methodSymbol
```
latonz: We use var across this codebase unless the type really adds some value or is necessary for the compilation. IMO it doesn't add any value here…

src/Riok.Mapperly/Symbols/Members/MemberPath.cs
```diff
@@ -44,29 +50,31 @@ public abstract class MemberPath(ITypeSymbol rootType, IReadOnlyList<IMappableMe
...
                 continue;
 
             yield return pathParts.ToArray();
         }
     }
 
-    public bool IsAnyNullable() => Path.Any(p => p.IsNullable);
+    public bool IsAnyReadNullable() => Path.Any(x => x.IsReadNullable);
+
+    public bool IsWriteNullable() => Path[^1].IsWriteNullable;
```
latonz: Shouldn't this be moved to `NonEmptyMemberPath`?

latonz (changes requested):

src/Riok.Mapperly/Symbols/Members/ParameterSourceMember.cs
Abhijithtv: I was initially concerned about `parameter.Symbol `being null, but after tracing the flow it appears all method parameters used by `ParameterSourceMember` are created through `WrapMethodParameter(IParameterSymbol symbol)` , which always provides a symbol.
 So the current implementation should be safe.

src/Riok.Mapperly/Symbols/Members/MemberPath.cs
Abhijithtv: Oops, Nice catch.

Abhijithtv:
@latonz 
I have updated the PR based on the feedback.

src/Riok.Mapperly/Symbols/Members/NonEmptyMemberPath.cs
```diff
@@ -23,13 +23,23 @@ public NonEmptyMemberPath(ITypeSymbol rootType, IReadOnlyList<IMappableMember> p
...
-    public override ITypeSymbol MemberType =>
-        IsAnyNullable() ? Member.Type.WithNullableAnnotation(NullableAnnotation.Annotated) : Member.Type;
+    public override ITypeSymbol MemberReadType =>
+        IsAnyReadNullable() ? Member.Type.WithNullableAnnotation(NullableAnnotation.Annotated) : Member.Type;
+
+    /// <summary>
+    /// Gets the type of the <see cref="Member"/> in the context of write. If last part of the path is nullable, this type will be nullable too.
+    /// </summary>
+    public override ITypeSymbol MemberWriteType =>
+        IsWriteNullable() ? Member.Type.WithNullableAnnotation(NullableAnnotation.Annotated) : Member.Type;
```
latonz: Wouldn't it make sense to assign these once as `Member` is immutable? Otherwise this creates a new object for each access, no?

latonz (changes requested):
Thanks for the updates. One small feedback point left.
Could you please also rebase? Afterwards I'll merge it 😊

Abhijithtv:
> Thanks for the updates. One small feedback point left. Could you please also rebase? Afterwards I'll merge it 😊

I have updated the PR and rebased the branch.

latonz (approved):

latonz:
@Abhijithtv thank you for this contribution

PR #2306, opened 2026-06-06
feat: support [MappingTargetOriginalValue] parameter in user-defined mapping methods

juliancarrivick:
# Support [MappingTargetOriginalValue] parameter in user-defined mapping methods

## Description

Following feedback in #2242, fixing #276 instead to cover the `Optional` use-case.

Here we add `MappingTargetOriginalValueAttribute`, which marks a parameter in a user-implemented mapping method to receive the existing target member value. Useful for merge-style mappings that need to combine source and destination data rather than fully overwrite.

Support for `[MappingTarget]` is not included as the use-case there is to work with entire object references instead of members. In that case both the "target" and the "original" value will end up being the same reference lead to potential confusion.

Thank you for considering this PR.

## Checklist

  - _Used to implement tests and generate first pass at implementation. Iterated heavily after that point to avoid leaking state everywhere (particularly around setting `MemberAssignmentMapping.TargetValueGetter` in `MemberMappingBuilder`)_


src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
```diff
@@ -47,22 +48,22 @@ public static bool TryBuildAssignment(
...
         return true;
     }
 
     public static bool TryBuild(
         IMembersBuilderContext<IMapping> ctx,
         MemberMappingInfo memberMappingInfo,
         CodeStyle codeStyle,
-        [NotNullWhen(true)] out ISourceValue? sourceValue
+        [NotNullWhen(true)] out SourceValueBuilder.MappedSourceValue? result
     )
```
juliancarrivick, line 67: _note_: This function really belongs in `SourceValueBuilder` - that is the only consumer and it leads to the confusing `MemberMappingBuilder.TryBuildContainerAssignment/TryBuildAssignment() -> SourceValueBuilder.TryBuildMappedSourceValue() -> MemberMappingBuilder.TryBuild()`. I haven't refactored that here to avoid an even larger diff, but can do so if you like.

docs/docs/configuration/additional-mapping-parameters.mdx
```diff
@@ -205,3 +205,56 @@ Mappings with additional parameters do have some limitations:
 - Generic and runtime target type mappings do not support additional type parameters.
 - Derived type mappings do not support additional type parameters.
   :::
+
+## Target original value parameter
```
latonz: Isn't this more a topic for the `User-implemented mapping methods
` page?

docs/docs/configuration/additional-mapping-parameters.mdx
```diff
@@ -205,3 +205,56 @@ Mappings with additional parameters do have some limitations:
...
+        // highlight-start
+        target.Age = FromOptional(source.Age, target.Age);
+        // highlight-end
+        return target;
+    }
+    ```
+  </TabItem>
+</Tabs>
+
+:::info
```
latonz: I'd remove this info block as it doesn't really add value.

src/Riok.Mapperly.Abstractions/MappingTargetOriginalValueAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+
+/// <summary>
+/// Marks a parameter of a user-implemented mapping method as receiving the current value
+/// of the destination member being mapped to.
+/// This allows the method to use the existing destination value (e.g. as a fallback).
+/// </summary>
+/// <example>
+/// <code>
+/// private static int? FromOptional(Optional&lt;int?&gt; source, [MappingTargetOriginalValue] int? original)
+///     =&gt; source.HasValue ? source.Value : destination;
```
latonz: This doesn't compile, does it? `destination` is not defined…

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberAssignmentMapping.cs
```diff
@@ -30,6 +35,9 @@ public IEnumerable<StatementSyntax> Build(TypeMappingBuildContext ctx, Expressio
 
     public ExpressionSyntax BuildExpression(TypeMappingBuildContext ctx, ExpressionSyntax? targetAccess)
     {
+        if (_targetValueGetter is not null && targetAccess is not null)
+            ctx = ctx.WithTargetValue(_targetValueGetter.BuildAccess(targetAccess));
```
latonz, line 40: Nit: This code base uses block-styled if bodies except for return and throw expressions.

src/Riok.Mapperly/AnalyzerReleases.Unshipped.md
```diff
@@ -1,2 +1,9 @@
 ; Unshipped analyzer release
 ; https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+
+### New Rules
+
+Rule ID  | Category | Severity | Notes
+---------|----------|----------|--------------------
+RMG100   | Mapper   | Error    | Multiple [MappingTargetOriginalValue] parameters on a mapping method.
+RMG101   | Mapper   | Error    | [MappingTargetOriginalValue] is not supported on generated (partial) mapping methods.
```
latonz: Move this to the shipped file as we don't really use the unshipped file.

src/Riok.Mapperly/Symbols/MappingMethodParameters.cs
```diff
@@ -7,5 +7,6 @@ public record MappingMethodParameters(
     MethodParameter Source,
     MethodParameter? Target,
     MethodParameter? ReferenceHandler,
+    MethodParameter? OriginalValueParameter,
```
latonz: Include `Target` in the name?

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
```diff
@@ -15,10 +15,13 @@ public static bool BuildParameters(
         [NotNullWhen(true)] out MappingMethodParameters? parameters
     )
     {
+        var matchedParameters = new List<MethodParameter?>();
```
latonz: Instead of a list, wouldn't it be much more efficient keeping a set of consumed parameter ordinals?

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
```diff
@@ -143,25 +155,44 @@ public static bool BuildRuntimeTargetTypeMappingParameters(
...
+        out MethodParameter? originalValueParameter
+    )
+    {
+        var candidates = method
+            .Parameters.Where(p =>
+                !MatchesExistingOrdinal(matchedParameters, p) && ctx.SymbolAccessor.HasAttribute<MappingTargetOriginalValueAttribute>(p)
+            )
+            .ToList();
+
+        if (candidates.Count > 1)
```
latonz: For the other attributes this is handled on L54 (`var hasInvalidAdditionalParameter = …`), IMO this should also be checked there. Not sure if the separate diagnostic is really required as we don't do this for the others. We should keep this consistent.

test/Riok.Mapperly.Tests/Mapping/UserMethodMappingTargetOriginalValueTest.cs
```diff
@@ -0,0 +1,139 @@
...
+            partial B Map(A src);
+            """,
+            "class A { public Optional Age { get; set; } }",
+            "class B { public int? Age { get; init; } }",
+            "struct Optional { public bool HasValue { get; } public int? Value { get; } }"
+        );
+
+        return TestHelper.VerifyGenerator(source);
+    }
+}
```
latonz, line 188: Add a test involving a source value / target value that needs an unsafe accessor.

docs/docs/configuration/additional-mapping-parameters.mdx
```diff
@@ -205,3 +205,56 @@ Mappings with additional parameters do have some limitations:
...
+`[MappingTargetOriginalValue]` has the following limitations:
+
+- Only one `[MappingTargetOriginalValue]` parameter is allowed per method.
+- Only supported on user-implemented (non-`partial`) mapping helper methods. It cannot be used on generated `partial` mapper method declarations.
+- When mapping to a new instance (no `[MappingTarget]`), the destination value is `default(T)` — the type's default — since there is no existing value to read.
```
latonz: This is not entirely true is it? For settable members, a value could already be set.

test/Riok.Mapperly.Tests/Mapping/UserMethodMappingTargetOriginalValueTest.cs
```diff
@@ -0,0 +1,139 @@
+using Riok.Mapperly.Diagnostics;
+
+namespace Riok.Mapperly.Tests.Mapping;
+
+public class UserMethodMappingTargetOriginalValueTest
+{
+    [Fact]
+    public Task BasicDestinationValueParameter()
```
latonz: Adjust the naming to match the naming of the source code. `DestinationValue` vs. `TargetOriginalValue` / `TargetValue`.

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedMethodMapping.cs
```diff
@@ -19,6 +19,7 @@ public class UserImplementedMethodMapping(
     ITypeSymbol sourceType,
     ITypeSymbol targetType,
     MethodParameter? referenceHandlerParameter,
+    MethodParameter? originalValueParameter,
```
latonz: Include the term `Target` in the name (same at several other locations in the code). While it makes the name slightly longer, I think it adds meaningful clarity.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
```diff
@@ -9,12 +9,15 @@
...
 using Riok.Mapperly.Diagnostics;
 using Riok.Mapperly.Helpers;
+using Riok.Mapperly.Symbols.Members;
 using static Riok.Mapperly.Emit.Syntax.SyntaxFactoryHelper;
 
 namespace Riok.Mapperly.Descriptors.MappingBodyBuilders;
 
 internal static class SourceValueBuilder
 {
+    internal sealed record MappedSourceValue(ISourceValue SourceValue, MemberPathGetter? TargetValueGetter);
```
latonz: Maybe adjust the name to `MappedValue` as it isn't a source value only anymore?

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/MemberMappingBuilder.cs
latonz: Thanks for the note. I'd be happy to see that refactoring, but I'd prefer to keep it separate PR from this change. If you're willing to, feel free to submit a follow-up PR 😊

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
```diff
@@ -23,26 +26,44 @@ public static bool TryBuildMappedSourceValue(
...
         [NotNullWhen(true)] out ISourceValue? sourceValue
-    ) => TryBuildMappedSourceValue(ctx, memberMappingInfo, MemberMappingBuilder.CodeStyle.Expression, out sourceValue);
+    )
+    {
+        if (!TryBuildMappedSourceValue(ctx, memberMappingInfo, MemberMappingBuilder.CodeStyle.Expression, out var result))
+        {
+            sourceValue = null;
+            return false;
+        }
+        sourceValue = result.SourceValue;
```
latonz, line 37: Nit: this codebase usually adds an empty line after a block. Same at other locations, ex. L58.

src/Riok.Mapperly/Descriptors/Mappings/TypeMappingBuildContext.cs
```diff
@@ -133,6 +143,10 @@ public TypeMappingBuildContext WithRefHandler(ExpressionSyntax refHandler) =>
                     yield return targetArgument.Value;
                 else if (referenceHandlerParameter is not null && param.Ordinal == referenceHandlerParameter.Value.Ordinal)
                     yield return referenceHandlerParameter.Value.WithArgument(refHandler);
+                else if (originalValueParameter is not null && param.Ordinal == originalValueParameter.Value.Ordinal)
+                    yield return originalValueParameter.Value.WithArgument(
+                        targetValue ?? DefaultExpression(SyntaxFactoryHelper.FullyQualifiedIdentifier(originalValueParameter.Value.Type))
```
latonz: The naming needs to be unified, `originalValue` / `targetOriginalValue` / `targetValue`, but all is meaning the same, isn't it?

latonz (changes requested):
Thank you for this contribution! I added my feedback.

latonz:
The vulnerability scanning which fails the test GH Action job should be gone once you rebase onto main.

juliancarrivick:
Thanks for the review @latonz ! I'm currently on holiday but will address your comments after I'm back next week :)

latonz:
enjoy your time off!

docs/docs/configuration/additional-mapping-parameters.mdx
juliancarrivick: ✔️

docs/docs/configuration/additional-mapping-parameters.mdx
juliancarrivick: Moved!

docs/docs/configuration/additional-mapping-parameters.mdx
juliancarrivick: This is for the case where the mapper creates an instance of the target itself e.g. in `test/Riok.Mapperly.IntegrationTests/MappingTargetOriginalValueMapperTest.cs` and there is no opportunity to put a setter.
```
        [Fact]
        public void MapToDtoShouldUseDefaultWhenOptionalEmpty()
        {
            var mapper = new MappingTargetOriginalValueMapper();
            var source = new OptionalObject();

            var dto = mapper.MapToDto(source);

            dto.Name.ShouldBeNull();
        }
```

I guess though this is slightly misleading because it won't receive `default(T)` it will receive the default constructed value on the DTO instead which might be something else.

In any case, I have removed the block as requested :)

src/Riok.Mapperly.Abstractions/MappingTargetOriginalValueAttribute.cs
juliancarrivick: Good catch! I went through a few different naming conventions as I iterated and missed one update. `destination` -> `original` which is now valid.

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
juliancarrivick: ✔️

src/Riok.Mapperly/AnalyzerReleases.Unshipped.md
juliancarrivick: ✔️

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
juliancarrivick: This was my attempt at avoiding repeating the `p?.Ordinal ?? -1` logic but you're right it's not great performance wise. I've introduced `MatchedParameterOrdinalList`, a thin private wrapper around `List<int>` (Set seemed inappropriate when the largest dataset will fit in a cache line) to get the best of both worlds. Let me know if you think that's overkill.

src/Riok.Mapperly/Descriptors/Mappings/MemberMappings/MemberAssignmentMapping.cs
juliancarrivick: ✔️

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
juliancarrivick: I was trying to keep the validation logic for the parameter close to it's searching location, but yep it's better to be consistent. Given this I've removed the diagnostic for it too.

test/Riok.Mapperly.Tests/Mapping/UserMethodMappingTargetOriginalValueTest.cs
juliancarrivick: Oh nice, I hadn't come across unsafe accessors before - added! I've only added a single unsafe source/target property test since `UnsafeAccessorsTest.cs` seems to cover all the possible permutations and I would think here we want to check the two features interact correctly.

src/Riok.Mapperly/Symbols/MappingMethodParameters.cs
juliancarrivick: ✔️

src/Riok.Mapperly/Descriptors/MappingBodyBuilders/SourceValueBuilder.cs
juliancarrivick: Yeah I found this one hard to name - I've gone with your suggestion given you're looking at it with a fresh set of eyes.

src/Riok.Mapperly/Descriptors/Mappings/UserMappings/UserImplementedMethodMapping.cs
juliancarrivick: ✔️

src/Riok.Mapperly/Descriptors/Mappings/TypeMappingBuildContext.cs
juliancarrivick: I think I was trying to be clever and reduce the verbosity of the names but coming back to this 2 weeks later it's just more confusing isn't it 😅 I've done a pass to name everything consistently - if you notice any more let me know.

test/Riok.Mapperly.Tests/Mapping/UserMethodMappingTargetOriginalValueTest.cs
juliancarrivick: Another missed update 😅  Did this and updated the generated files ✔️

juliancarrivick (commented):
Alright, I think I've covered everything - let me know if you require any further changes (and build should pass this time).
Thanks!

src/Riok.Mapperly/Descriptors/Mappings/TypeMappingBuildContext.cs
```diff
@@ -133,6 +143,11 @@ public TypeMappingBuildContext WithRefHandler(ExpressionSyntax refHandler) =>
                     yield return targetArgument.Value;
                 else if (referenceHandlerParameter is not null && param.Ordinal == referenceHandlerParameter.Value.Ordinal)
                     yield return referenceHandlerParameter.Value.WithArgument(refHandler);
+                else if (targetOriginalValueParameter is not null && param.Ordinal == targetOriginalValueParameter.Value.Ordinal)
+                    yield return targetOriginalValueParameter.Value.WithArgument(
+                        targetValue
+                            ?? DefaultExpression(SyntaxFactoryHelper.FullyQualifiedIdentifier(targetOriginalValueParameter.Value.Type))
```
latonz, line 149: When `targetValue` is absent, this fallback uses the original parameter symbol type. For generic user-implemented methods used in constructor/init-only mappings, that can emit an unsubstituted type parameter at the call site, e.g. `FromOptional<int>(src.Age, default(T?))`.  Use the concrete target/original-value type for the fallback and add a test for generic `[MappingTargetOriginalValue]` with an init-only or constructor-mapped member.

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
```diff
@@ -143,25 +152,35 @@ public static bool BuildRuntimeTargetTypeMappingParameters(
...
         var targetParameterSymbol = method.Parameters.FirstOrDefault(p =>
-            p.Ordinal != sourceParameter.Ordinal
-            && p.Ordinal != refHandlerParameterOrdinal
-            && !ctx.SymbolAccessor.HasAttribute<ReferenceHandlerAttribute>(p)
+            !existingMatchedParameters.ContainsByOrdinal(p) && !ctx.SymbolAccessor.HasAttribute<ReferenceHandlerAttribute>(p)
         );
         return ctx.SymbolAccessor.WrapOptionalMethodParameter(targetParameterSymbol);
     }
 
+    private static MethodParameter? TryFindFirstTargetOriginalValueParameter(
```
latonz: This method name could be simplified to `FindTargetOriginalValueParameter` as all of these methods only use the first param and also don't include it in the name.

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
```diff
@@ -15,10 +15,13 @@ public static bool BuildParameters(
         [NotNullWhen(true)] out MappingMethodParameters? parameters
     )
     {
+        var matchedParameters = new MatchedParameterOrdinalList();
+
         var refHandlerParameter = FindReferenceHandlerParameter(ctx, method);
-        var refHandlerParameterOrdinal = refHandlerParameter?.Ordinal ?? -1;
+        matchedParameters.AddParameter(refHandlerParameter);
 
         var sourceParameter = FindSourceParameter(ctx, method, refHandlerParameter);
```
latonz: Shouldn't the source parameter be matched after the `targetOriginalValueParameter`? Otherwise signatures like `FromOptional([MappingTargetOriginalValue] int? original, Optional source)` won't work, do they?

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
```diff
@@ -201,4 +220,32 @@ public static bool BuildRuntimeTargetTypeMappingParameters(
 
         return refHandlerParameter;
     }
+
+    private readonly struct MatchedParameterOrdinalList
```
latonz: Does this really add any value? Wouldn't a simple `HashSet<int> matchedParamOrdinals` be easier to read and have better performance?

latonz (changes requested):

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
juliancarrivick: I wouldn't expect it to have better performance actually since we're only dealing with O(3) and HashMap will have the hashing overhead. I thought I'd quantitatively check though:

```
➜ dotnet new console
➜ dotnet add package BenchmarkDotNet
➜ cat > Program.cs << 'EOF'
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

BenchmarkRunner.Run<HashSetVsListBenchmark>();

[MemoryDiagnoser]
public class HashSetVsListBenchmark
{
    private readonly List<int> _list = new() { 1, 2, 3 };
    private readonly HashSet<int> _set = new() { 1, 2, 3 };

    [Params(1, 2, 3, 99)]
    public int Target;

    [Benchmark(Baseline = true)]
    public bool ListContains() => _list.Contains(Target);

    [Benchmark]
    public bool HashSetContains() => _set.Contains(Target);
}
EOF
➜ dotnet run -c Release
...
| Method          | Target | Mean     | Error     | StdDev    | Ratio | Allocated | Alloc Ratio |
|---------------- |------- |---------:|----------:|----------:|------:|----------:|------------:|
| ListContains    | 1      | 1.304 ns | 0.0016 ns | 0.0013 ns |  1.00 |         - |          NA |
| HashSetContains | 1      | 2.197 ns | 0.0124 ns | 0.0097 ns |  1.68 |         - |          NA |
|                 |        |          |           |           |       |           |             |
| ListContains    | 2      | 1.816 ns | 0.0035 ns | 0.0031 ns |  1.00 |         - |          NA |
| HashSetContains | 2      | 2.206 ns | 0.0156 ns | 0.0130 ns |  1.21 |         - |          NA |
|                 |        |          |           |           |       |           |             |
| ListContains    | 3      | 1.985 ns | 0.0043 ns | 0.0033 ns |  1.00 |         - |          NA |
| HashSetContains | 3      | 2.209 ns | 0.0121 ns | 0.0113 ns |  1.11 |         - |          NA |
|                 |        |          |           |           |       |           |             |
| ListContains    | 99     | 2.272 ns | 0.0095 ns | 0.0084 ns |  1.00 |         - |          NA |
| HashSetContains | 99     | 2.216 ns | 0.0135 ns | 0.0120 ns |  0.98 |         - |          NA |
```

However, this isn't performance critical code, so HashMap will probably still be fast enough, given you'd prefer it for readability. I wanted to avoid repeating the `p?.Ordinal ?? -1` logic but if you feel that's more readable than a separate helper I'm happy to defer to you here.

src/Riok.Mapperly/Descriptors/Mappings/TypeMappingBuildContext.cs
juliancarrivick: Oooh excellent catch. Updated so that `GenericUserImplementedNewInstanceMethodMapping` will inject the concrete type so the type at this call-site is correct :white_check_mark:

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
juliancarrivick: ✔️

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
juliancarrivick: Noticed on closer inspection that the `!Contains()` HashSet case is actually marginally faster so maybe this evens out in the wash, regardless, I've updated 👍

src/Riok.Mapperly/Descriptors/UserMappingMethodParameterExtractor.cs
juliancarrivick: Ah yep, another good catch. I've fixed this and added a test case.

latonz (approved):

latonz:

PR #2369, opened 2026-07-14
feat: add QueryableProjectionNullHandling option to skip projection null handling

fdipuma:
## Description

Adds a `QueryableProjectionNullHandling` mapper option that lets `IQueryable<T>` projection mappings skip null handling, for relational providers (e.g. EF Core) where the projection lambda is translated to SQL and never executed in memory.

This implements part 3 of #2317 and should close:
- Fixes #2350: nullable navigation -> non-nullable DTO no longer emits unnecessary SQL `CASE` blocks nor reports `RMG002` ("no accessible parameterless constructor").
- Fixes #1293: projection output is now identical under `#nullable enable`/`disable`.

### Option

```csharp
public enum QueryableProjectionNullHandling
{
    NullSafe = 0, // default, current null-safe behaviour
    Ignore,       // skip null handling in queryable projections
}
```

Set at mapper or assembly (`MapperDefaults`) level:

```csharp
[Mapper(QueryableProjectionNullHandling = QueryableProjectionNullHandling.Ignore)]
public static partial class CarMapper
{
    public static partial IQueryable<CarDto> ProjectToDto(this IQueryable<Car> q);
}
```

Under `Ignore`, projection member access is emitted directly (`x.Owner.Address.City`, `x.IntValue.Value` for nullable value types) with no `?.`, ternary `!= null ? … : default`, coalesce, or null fallback; the `RMG089` nullability diagnostic is suppressed. Nullable value-type members are unwrapped so the generated expression compiles. The default `NullSafe` keeps today's behaviour unchanged (non-breaking).

### note

The related idea from #2317 of honoring a conversion operator's own nullable annotations in **non-projection** mappings is intentionally **not** included here: it turned out to require changes to the shared member-mapping null-guard path and is best handled as a separate, focused PR. This PR is limited to the queryable-projection option.

## Checklist

- [x] I did not use AI tools to generate this PR, or I have manually verified that the code is correct, optimal, and follows the project guidelines and architecture
- [x] I understand that low-quality, AI-generated PRs will be closed immediately without further explanation
- [x] The existing code style is followed
- [x] The commit message follows our guidelines
- [x] Performed a self-review of my code
- [x] Hard-to-understand areas of my code are commented
- [x] The documentation is updated (as applicable)
- [x] Unit tests are added/updated
- [x] Integration tests are added/updated (as applicable, especially if feature/bug depends on roslyn or framework version in use)

Issue #2317, opened 2026-06-12
Annotation-driven null-guard strategy for user-defined conversions (with a configurable override)


fdipuma:
**Is your feature request related to a problem? Please describe.**

When a member mapping goes through a user-defined conversion on a nullable source (`ValueObject? -> primitive?`), Mapperly always wraps the conversion in a null guard:

```csharp
// object mapping and queryable projection alike
PromoCode = source.PromoCode != null ? (string)source.PromoCode : default,
```

The guard is the right default for conversions that do not accept null, but it is emitted unconditionally, even when the conversion itself is null-propagating, and even in contexts where the conversion is never executed at all. Two concrete pain points:

1. **Null-safe conversions get a useless guard.** A reference-type value object can declare a null-propagating cast (`explicit operator string?(Code? value)`); value-object generators are moving in this direction for class-based VOs, and struct VOs already behave this way via C# lifted conversions. The guard adds noise to the generated code for no semantic benefit.
2. **In queryable projections the guard becomes redundant, and potentially harmful, SQL.** EF Core translates the guard to `CASE WHEN [o].[PromoCode] IS NULL THEN NULL ELSE [o].[PromoCode] END`, although reading the column already propagates `NULL`. When the query composes further over the projected member (OData-style `$filter`/`$orderby` on the DTO), predicates over a `CASE WHEN` are not sargable, so indexes on the underlying column cannot be used. EF Core's own projections of converter-backed members emit the bare column read.

**Describe the solution you'd like**

A three-part, annotation-driven strategy:

**1. `Auto` (proposed default): derive the guard from the conversion operator's NRT contract.**

- `explicit operator string(Code value)` (non-nullable parameter) → the type declares it does not accept null → emit the guard (current behavior).
- `explicit operator string?(Code? value)` (nullable-annotated parameter) → the type declares the conversion is null-propagating → emit the bare conversion, no guard, in object mappings and projections alike.
- Struct value objects → nothing to do, C# lifted conversions already propagate null.

This reuses the same Roslyn nullable-annotation information Mapperly already relies on for member nullability, so the trust model is consistent: annotations are the contract.

**2. Mapper-level override for types whose signatures cannot be changed:**

```csharp
// Relevant new / adjusted API surface
[Mapper(ConversionNullGuard = ConversionNullGuard.Auto)]   // proposed default
// ConversionNullGuard.Always, current behavior, guard every nullable conversion
[...]

PR #2371, opened 2026-07-17
feat: honor null-accepting user-defined conversion operators

fdipuma:
## Description

Implements parts 1–2 of #2317: Mapperly now honors a user-defined conversion operator's nullable reference type annotations, consistent with how NRT annotations are trusted elsewhere.

When a nullable source is converted through a user-defined operator whose parameter is nullable (e.g. `explicit operator string?(Code? c)`), the operator can accept `null` directly, so no null guard is generated:

```csharp
// class Code { public static explicit operator string?(Code? c) => c?.Value; }
// A { Code? PromoCode } -> B { string? PromoCode }
target.PromoCode = (string?)source.PromoCode;          // before: source.PromoCode != null ? (string)source.PromoCode : default
```

An operator whose parameter is non-nullable keeps the guard, so it is never invoked with `null`:

```csharp
// public static explicit operator string(Code c) => c.Value;
target.PromoCode = source.PromoCode != null ? (string)source.PromoCode : default;   // unchanged
```

This applies to object member/property mappings, nested members, and top-level mapper-method signatures.

### How it works

A detector (`MappingBuilderContext.HasNullAcceptingUserDefinedConversion`) checks the operator's parameter annotation via `Compilation.ClassifyConversion`. At the two points where source nullability is normally stripped (`FindOrBuildLooseNullableMapping` for members, `NullableMappingBuilder` for top-level), the source nullability is preserved for such conversions so the resulting `CastMapping` keeps a nullable source type, and the existing null-handling logic then skips the guard. No guard-decision code was changed. Builder priority is preserved: the guard is only dropped when a cast is the conversion Mapperly would have selected anyway (a higher-priority `Parse`/constructor/etc. still wins for a nullable source).

### Notes

- Independent of #2369 (the `QueryableProjectionNullHandling` projection change, part 3 of #2317), no shared files; can be reviewed/merged in any order.
- For a nullable source mapped to a **non-nullable** target member through a null-accepting operator, the existing `RMG089` diagnostic is not reported (the rebuilt cast satisfies the same exemption `RMG090` uses); behavior is pinned by a test. Happy to adjust if you'd prefer a diagnostic there.

## Checklist


latonz:
I rebased this PR and made a few final adjustments before merging:
* added support for parameter nullability attributes
* remove conversions docs section as this is the expected behavior and is not documented for non-cast conversions too

Thanks @fdipuma for your patience and for the contribution!

